PR #7, opened 2015-08-12
Small bits

vext01:
A handful of small fixes to the warmup experiment.

Looks OK?


ltratt:
I guess the two CPython commits should be squashed into one? I can't see any advantage to having them separate in this case.

vext01:
I don't mind either way. Will squash them.

vext01:
Ready for re-review.

ltratt:
Looks good.

PR #13, opened 2015-09-28
Graal tweaks

vext01:
Tweaks to make the Graal branch build on bencher2.

For some reason we are unable to build the jvmci using the jdk8 built from source. Something in maven barfs over SSL. Probably a configuration issue.

Fix suggested by Chris: use the system jdk8 to build jvmci.

Fixes the issue for me.

OK?


vext01:
Use a more descriptive name for the helper function. OK?

PR #15, opened 2015-10-02
make fasta deterministic

cfbolz:


benchmarks/fasta/javascript/bench.js
```diff
@@ -19,18 +20,22 @@ var ALU =
...
-  g: 0.1975473066391,
-  t: 0.3015094502008
+function HomoSap() {
+  return {
+    a: 0.3029549426680,
+    c: 0.1979883004921,
+    g: 0.1975473066391,
+    t: 0.3015094502008
+  }
 }
```
vext01, line 39: Well spotted.

benchmarks/fasta/php/bench.php
```diff
@@ -10,7 +10,8 @@
 ob_implicit_flush(1);
 ob_start(NULL, 4096);
 
-$last = 42.0;
+$INITIAL_STATE = 42.0;
```
vext01, line 13: Nitpicking here, but you can use `define` to define constants in PHP. I don't mind if you leave it either.

ltratt, line 13: From memory, using `define` is faster in Zend (and probably Hippy, a bit). I might be wrong.

benchmarks/fasta/php/bench.php
```diff
@@ -128,15 +129,16 @@ function makeRepeatFasta($s, $n) {
    'AGGCGGAGGTTGCAGTGAGCCGAGATCGCGCCACTGCACTCC' .
    'AGCCTGGGCGACAGAGCGAGACTCCGTCTCAAAAA';
 
+makeCumulative($iub);
+makeCumulative($homosapiens);
+
```
vext01, line 134: Don't we give PHP an unfair advantage here? In the javascript version `makeCumulative()` is called once per call to `fastaRandom()`. Here it happens once per execution.

benchmarks/fasta/ruby/bench.rb
```diff
@@ -62,6 +63,7 @@ def make_random_fasta(table, n)
     prob = 0.0
     rwidth = (1..width)
     table.each{|v| v[1]= (prob += v[1])}
+    table = table.collect{|v| prob += v[1]; [v[0], prob] }
```
vext01, line 65: Hrm, I don't understand this.

cfbolz, line 65: this is making a copy instead of mutating the table.

benchmarks/fasta/php/bench.php
cfbolz, line 134: note that that was completely inconsistent across the languages before and after this change. we could consistencify it, but probably in an extra checkin, because it requires touching all of them.

vext01, line 134: That sounds like a plan.

benchmarks/fasta/ruby/bench.rb
vext01, line 65: Sorry if I am being daft, but does this line not also mutate prob?

cfbolz, line 65: yes, but `prob` is a local variable, so that's fine. the previous code was mutating `table`, which is a global.

vext01, line 65: I don't know if I'm missing something here:
- line 65 mutates table (from function scope) in-place and mutates prob (also in function scope).
- line 66 mutates the same prob, and creates a new table.

So prob is mutated twice where before it was mutated once. Did you mean to to remove line 65?

vext01:
OK, looks good.

Bug for `makeCumulative`: https://github.com/softdevteam/warmup_experiment/issues/16

Thanks

PR #24, opened 2015-10-06
Improve graph styles.

vext01:
A fair amount of duplication here at the moment. We can eliminate that later in another PR if we so wish.


vext01:
This ended up being a large overhaul.
- Fix styles, as per email convo.
- Move all duplicated code into a common place.
- PEP8 the whole lot.

Looks OK?

ltratt:
Looks good!

PR #27, opened 2015-10-09
Pre load classes

cfbolz:
to make sure that the benchmarking Java class is loaded already during the first iteration of each benchmark, I call an empty static init method in the `KrunEntry` classes.


ltratt:
Well spotted!

PR #29, opened 2015-10-12
C richards

vext01:
This PR imports richards and tidies the build system a bit.

I've kept the import and modification of the benchmark itself separate.

A couple of issues with the original:
- It only accepts 10 or 100 reps.
- I think it is wrong!

The benchmark makes five task chains. In the original C version, the first chain is created thusly:

```
createtask(I_IDLE, 0, wkq, S_RUN, idlefn, 1, Count * reps);
```

Notice `Count * reps` where as in the other language versions this section is in a loop. I could believe that this is equivalent (and the checksums were passing) if it were the case that the other four task chains had also been scaled up, which they have not!

```
    createtask(I_WORK, 1000, wkq, S_WAITPKT, workfn, I_HANDLERA, 0);
    ...
    createtask(I_HANDLERA, 2000, wkq, S_WAITPKT, handlerfn, 0, 0);
    ...
    createtask(I_HANDLERB, 3000, wkq, S_WAITPKT, handlerfn, 0, 0);
    ...
    createtask(I_DEVA, 4000, wkq, S_WAIT, devfn, 0, 0);
    createtask(I_DEVB, 5000, wkq, S_WAIT, devfn, 0, 0);
```

So it is likely that this does less work than the other languages. Am I right?

I've amended this to accept any number of reps, and in doing so fixed the bug. Checksums passing.

I've hacked it so that only existing C benchmarks and built and run. In each subsequent PR, I will import one more benchmark and eventually those hacks can go for good.

OK?

### DIFF
--- a/benchmarks/richards/c/bench.c
+++ b/benchmarks/richards/c/bench.c
@@ -1,0 +1,395 @@
+
+/*  C version of the systems programming language benchmark 
+**  Author:  M. J. Jordan  Cambridge Computer Laboratory. 
+**  
+**  Modified by:  M. Richards, Nov 1996
+**    to be ANSI C and runnable on 64 bit machines + other minor changes
+**  Modified by:  M. Richards, 20 Oct 1998
+**    made minor corrections to improve ANSI compliance (suggested
+**    by David Levine)
+**
+**  Compile with, say
+**
+**  gcc -o bench bench.c
+**
+**  or
+**
+**  gcc -o bench100 -Dbench100 bench.c  (for a version that obeys
+**                                       the main loop 100x more often)
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#define                Count           10000
+#define                Qpktcountval    23246
+#define                Holdcountval     9297
+
+#define                TRUE            1
+#define                FALSE           0
+#define                MAXINT          32767
+
+#define                BUFSIZE         3
+#define                I_IDLE          1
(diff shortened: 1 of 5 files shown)


ltratt:
It may be that during the deterministic check, we can also work out definitively whether this is doing the same work as other languages?

benchmarks/richards/c/bench.c
```diff
@@ -0,0 +1,389 @@
...
+void createtask(int id,
+                int pri,
+                struct packet *wkq,
+                int state,
+                struct task *(*fn)(struct packet *),
+                long v1,
+                long v2)
+{
+    struct task *t = (struct task *)malloc(sizeof(struct task));
+
```
cfbolz, line 139: Who ever frees these `reps * 6` tasks? It seems nobody?

ltratt, line 139: Looks like `free(t)` before all but the first `createtask` call might be an easy solution?

cfbolz:
It looks pretty good to me now, but I'd appreciate it it @ltratt would take another brief look, since my C isn't the best.

vext01:
can do

ltratt:
Nothing jumps out at me as obviously wrong, but I must admit that I haven't followed through all the places that pointers get thrown around.

vext01:
Try this.

vext01:
splat

cfbolz:
happy, merging

PR #30, opened 2015-10-13
C fasta

vext01:
Here is the C fasta with checksums.

I have another branch that begins adding the checksums to other languages:
https://github.com/softdevteam/warmup_experiment/tree/fasta_checksums

I will come back to that later.

As discussed with CF, due to time constraints we are going to do all the C benchmarks first.

OK?

### DIFF
--- a/benchmarks/fasta/c/bench.c
+++ b/benchmarks/fasta/c/bench.c
@@ -1,0 +1,138 @@
+/*  The Computer Language Benchmarks Game
+ *
+ *  http://shootout.alioth.debian.org/
+ *
+ *  contributed by Mr Ledrug
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+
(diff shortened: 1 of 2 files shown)


benchmarks/fasta/c/bench.c
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/*
+ * Used to generate a checksum to verify benchmark correctness
+ */
+void wrap_write(int fd, char *buf, size_t len) {
+   int i;
+
+   for (i = 0; i < len; i++) {
+      /* In languages like Python, we would modulo 2^{32} */
```
cfbolz: I don't think we need this comment, do we?

vext01: Not really. It was to demonstrate the idea. Will remove.

vext01:
OK to squash?

cfbolz:
looks good, merging

PR #34, opened 2015-10-13
C fannkuch2

vext01:
Please check this one carefully.

I've lifted the restriction that `max_n <= 16`


ltratt:
Is it just me or does this use an indent of 3?

vext01:
Yes, a l lot of the C benchmarks do. Although I think this is silly, I've kept with this convention to minimise diff distance. In case we diff any of these against the VMs experiment in the future...

cfbolz:
looks good to me

PR #35, opened 2015-10-13
Export Tikz charts where possible

snim2:
This PR uses this library:

https://pypi.python.org/pypi/matplotlib2tikz

to export charts in `.tikz` format, which should look better when included in a LaTeX document. A PDF should be exported whether the .tikz can be produced or not.

To test this, create a LaTeX document similar to this:

``` tex
\documentclass[a4paper]{article}

\usepackage{tikz}
\usepackage{pgfplots}
\usepgfplotslibrary{groupplots}
\pgfplotsset{compat=newest}

\newlength\figureheight{}
\newlength\figurewidth{}
\setlength\figureheight{4cm}
\setlength\figurewidth{6cm}
\begin{document}
\input{FILENAME.tikz}
\end{document}
```

where `FILENAME` is one of the charts created by Make.


snim2:
NB, this PR is waiting on an upstream issue: https://github.com/nschloe/matplotlib2tikz/issues/64

snim2:
OK. this works now, so the next question is how to use it within a LaTeX document. The attached PDF is the output of this document:

``` latex
\documentclass[a4paper]{article}

\usepackage{graphicx}

\usepackage{tikz}
\usepackage{pgf}
\usepackage{pgfplots}
\usepgfplotslibrary{groupplots}
\pgfplotsset{compat=newest}

\newlength\figureheight{}
\newlength\figurewidth{}
\setlength\figureheight{6cm}
\setlength\figurewidth{8cm}

\title{Chart comparison}
\author{}
\date{}
\begin{document}
\maketitle

\section{Original PDF}
\includegraphics[width=\figurewidth,height=\figureheight]{chart}

\section{Tikz}
\input{chart.tikz}
\end{document}
[test.pdf](https://github.com/softdevteam/warmup_experiment/files/14163/test.pdf)
```

vext01:
That's great Sarah!

There's a few formatting bits that we need to apply (if possible):
- Overlapping subtitles and axes labels (make font smaller?)
- Can we turn off ticks? I think they are unneccesary when the grid mirrors them.
- Can we turn off the solid black border? This makes it easier to see data against the left axis.
- Can we get our punctuation back? Just a matter of escaping? Or you could even use a library (too overkill?) Maybe: https://pypi.python.org/pypi/pylatexenc/

As for using the data -- I think the idea is to import the data into the softdev repo (right @ltratt?). Krun now emits the result json files bzip2'd so that the files are not horrendously large.

For now, we could import our scratch data to get us off the ground (don't foget to bzip the json files). We would then need a make target that generates the tikz graphs in a directory. We then make this make target a dependency of the paper and start `input`ing the graphs as we need them.  To be clear, we would not import the generated tikz files.

Please don't import the data just yet. Let's see if what I have proposed is OK with everyone.

Also, I want to make it so `mk_graphs*.py` don't depend on a krun config file (this would also decouple our graphing scripts from krun). This should happen first. See https://github.com/softdevteam/krun/issues/48

I'll try to prioritise that for you so you don't need to block.

Makes sense?

ltratt:
The working plan is to import the _final_ data into the softdev repo. We don't want to import a zillion intermediate versions, otherwise we'll just end up with whacking great `.git` directories and make future `clone`s a miserable experience. One possibility is for people to have a local copy of the data in the meantime; another is that the Makefile automagically downloads it from somewhere. I know this a bit icky, but large binary files will remain a problem for a while to come...

vext01:
Fetching from someplace is fine with me.

snim2:
@vext01 For the "formatting bits", I didn't want to do anything here without some discussion. In my view a sensible way forward would be to have separate "look and feel" configs for the `.tikz` file and the `.pdf`/ interactive. Does that sounds like a good way foward?

@ltratt a third option would be to have a separate Git repo and add it to this repo as a submodule. Then, whenever someone recursively pulled or cloned the repo they would get the "current" data. Or fourthly, GitHub has a large file storage system: https://git-lfs.github.com/

vext01:
Does `git submodule --init` fetch all submodules in the current repo? If so, it doesn't help. As people add more submodules, we will be forced to download them all as if the data were in the repo proper.

Regarding styles, whatever is easier. Laurie and I spent some time making the graphs look "nice". If you could try to mimic the look that would be cool. Of course improvements are also welcomed. :)

snim2:
OK, watch this space...

ltratt:
I don't feel strongly about submodules or whatever. In a sense, it's a temporary thing (we want the final data to be in the main repo so that we can't lose it), so we maybe don't want to over think it.

vext01:
@snim2: Would the seaborn work deem this PR unnecessary?

snim2:
I don't know. These are probably different concerns. 
I suggest we come back to this after the pre-print though - the sticking point is that font sizes and so on are slightly different with tikz, so all the careful adjustments would need to be re-worked.

snim2:
@vext01 and I discussed this and came to the conclusion that it wouldn't be worth bringing this up to date.

PR #36, opened 2015-10-15
C nbody

vext01:
Add C version of nbody and add checksum.

Renaming of `advance` is to avoid a clash with a function of the same name in glibc.

OK?

### DIFF
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -16,10 +16,12 @@ BENCHMARKS = 		binarytrees \
 C_BENCHMARKS =	richards \
 		fasta \
 		binarytrees \
-		fannkuch_redux
+		fannkuch_redux \
+		nbody
 
 # Extra linker flags for C benchmarks
 C_EXTRA_LDFLAGS_binarytrees = -lm
+C_EXTRA_LDFLAGS_nbody = -lm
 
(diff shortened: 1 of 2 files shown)


ltratt:
I don't understand why `setup_state` is a separate function. It would seem to make sense _iff_ the C runner in Krun separately called "setup" vs "run iter"?

vext01:
I don't understand what you mean.

`setup_state()`  needs to be called once  per innermost iteration, not once per krun iteration.

ltratt:
I guess my point is: a two line function seems pointless -- it makes the code harder to read.

I'm also curious as to whether the `malloc` should be inside the `for` loop or not (I kind of think it should, but I suppose it depends on how the other language versions do this).

vext01:
Inlining the function is trivial. The malloc can go in either position, but to put it in side the loop is wasteful (in the other benchmarks, I have tried to be as efficient as possible too).

vext01:
OK?

ltratt:
OK!

ltratt:
Do you want to squash or should I merge as-is?

vext01:
I'll splat that last commit in.

vext01:
Squelch. Go ahead.

PR #39, opened 2015-10-16
Fasta checksums

vext01:
Bring all fasta benchmarks in line with the C one.

We can do a ton of squashing here, but I'm showing you the separate commits as their messages may help comprehension.

Please check very carefully.

OK?

### DIFF
--- a/benchmarks/fasta/c/bench.c
+++ b/benchmarks/fasta/c/bench.c
@@ -56,11 +56,6 @@ void wrap_write(int fd, char *buf, size_t len) {
    for (i = 0; i < len; i++) {
       checksum += buf[i];
    }
-
-#ifdef DEBUG
-   /* In real benchmarking we can't emit to stdout */
-   write(fd, buf, len);
-#endif
 }
 
 inline void str_write(char *s) {
--- a/benchmarks/fasta/c/bench.c
+++ b/benchmarks/fasta/c/bench.c
@@ -47,11 +47,6 @@ static u_int32_t checksum = 0;
 
 static unsigned rseed = START_RAND_SEED;
 
-void reset_state(void) {
-   checksum = 0;
-   rseed = START_RAND_SEED;
-}
-
 /*
  * Used to generate a checksum to verify benchmark correctness
  */
@@ -151,13 +146,8 @@ void run_iter(int n) {
    int i = 0;
 
    for (i = 0; i < n; i++) {
-      /* str_write(">ONE Homo sapiens alu\n"); */
       str_repeat(alu, SCALE * 2);
-
-      /* str_write(">TWO IUB ambiguity codes\n"); */
       rand_fasta(iub, SCALE * 3);
-
-      /* str_write(">THREE Homo sapiens frequency\n"); */
       rand_fasta(homosapiens, SCALE * 5);
 
       if (checksum != EXPECT_CKSUM) {
@@ -165,6 +155,8 @@ void run_iter(int n) {
 	     checksum, EXPECT_CKSUM);
       }
 
-      reset_state();
+      /* reset state */
+      checksum = 0;
+      rseed = START_RAND_SEED;
    }
 }
--- a/benchmarks/fasta/java/fasta.java
+++ b/benchmarks/fasta/java/fasta.java
@@ -16,7 +16,6 @@ class ChecksumOutputStream extends OutputStream {
 
     private long checksum = 0;
     private static long MOD;
-    private static final boolean DEBUG = false;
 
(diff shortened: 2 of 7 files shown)


ltratt:
Nothing jumps out as being wrong (but there is a lot of code!), which is good.

However, I'm pretty unconvinced by the `DEBUG` flag. I don't think that should be in any of the benchmarks. At best, it won't do anything; at worst, it might penalise a VM in ways that we can't really predict.

vext01:
I'm happy to remove them. They are only there for my benefit when DEBUGging :)

Let me get to a good break point and I'll come back to this.

snim2:
I've had a read through this and I can't see any obvious issues.

vext01:
Maybe I'm striving for an impossible goal, but I've just realised that the modulo stuff I am doing here might not be consistent. Let me run this past you guys and gals.
- Checksums in C are modulo 2^32 by construction of unsigned machine integers.
- Other languages have different number representations and may need to do extra work to achieve the same checksum (i.e. a modulo operation).

The extra modulo operations wlil give the languages that use them an unfair disadvantage, yet obviously the modulo is required for languages like python where the integers will never roll over.

I wonder if I should be using parameters that ensures checksums never go > 2**32. This would mean a modulo is never required, thus making this all more fair.

(If we go that route I would even consider ensuring that the checksum never goes higher than 2^31-1, so that the sum fits in a signed 32 (e.g.. Java ints)).

The problem is, how will I know the sum never exceeded the desired upper bound?

Am I reading into this too much?

ltratt:
You're reading too much into it.

vext01:
To be clear, we are OK if some of the languages execute modulo operations whereas others don't?

ltratt:
How do we even know?

snim2:
ISTM that the important thing is that the algorithms are really identical. You are thinking about whether the internals of each VM work in the same way, which is something I don't think we need to consider. Or maybe I have the wrong end of the stick?

vext01:
There are explicit modulo operations in some of the implementations to ensure that the checksums overflow as they would in C.

snim2:
Hmm. So you mean the benchmarks are implemented slightly differently in the different languages?

ltratt:
I don't see how you can work around the fact that different languages might represent numbers differently.

vext01:
Yes, but only for checksum related code. If we didn't modulo, we would have no guarantee that the checksums would match those of the C version, thus meaning I'm not sure the benchmarks are doing the same work.

@ltratt I think you might be right, but I wanted to raise this in case.

ltratt:
There is no problem here. If language X implements modulo differently, that's merely something to factor into the benchmarks.

In the paper, however, we need to be _very_ clear that we have created a slight variant of the traditional fasta benchmark. Our fasta isn't _quite_ comparable to other fastas any more.

vext01:
Just to be clear:

In C we might have:

```
u_int32_t checksum;
...
loop {
    checksum += something(); // no modulo, as this is implied.
}
```

And in Python:

```
checksum = 0
MOD = 2**32
...
some_loop:
    checksum += something() % MOD
```

ltratt:
Since we're mostly running on 64 bit machines, it's a reasonable assumption that most languages will fit a 32 bit int happily in a machine word. Let's move on.

vext01:
Sorry to persist, but statement is neither here nor there.

The problem is that to achieve the same checksum values some languages will need modulo operations whereas others will not. If we are OK with this discrepancy, then that's fine, as it stands I think I have failed to explain the problem.

Do you follow?

snim2:
Well, the only possible fix for this, to make the benchmarks more "fair" is to add the unnecessary modulo to the C code. The question is whether that extra bit of fairness is useful, given that a C compiler may remove the modulo anyway.

I'm not sure what the answer is TBH, I can see arguments either way, and the impact is probably not great.

vext01:
Right, I'm not certain the C compiler wont throw it away, which is why I have not mentioned this as a solution. In the above cases, the type alone is grounds enough to throw away the modulo. I.e. `x + 1 % 2^32 <=> x + 1` if x is of type `uint32_t`.

I think we can conclude that we can't win here, and now I _will_ move on. Sorry for the noise.

vext01:
Here's the debug stuff gone

vext01:
Looks ok, will squash?

ltratt:
OK, although I don't think this needs to be squashed to a single commit necessarily.

vext01:
I'll see what i can do

vext01:
This is probably reasonable? Merge if you like.

ltratt:
Looks good.

PR #40, opened 2015-10-19
Assume fasta checksums will not overflow or saturate.

vext01:
This checksum business was intruding my thoughts over the weekend. I have come to peace with the following idea: Let's assume the checksums do not overflow or saturate.

Facts that lead me to think this is a better idea:
- All of the languages we are using support numbers in the range of a signed64. Some as integers, some as doubles.
- Signed64 allows a huge numeric range. Maxing this out would be unlikely.
- If a checksum does exceed this range, it will either over/underflow (in the case of integer types) or saturate (for doubles). In this case, we will know about it (checksum will fail), and thus we can change the benchmark parameters to ensure this does not happen.
- With this assumption, no language has to do modulo's ever. Thus making it fairer.

FWIW here are the "integer" types in our languages:
- C: whatever int type you choose.
- Java: int=signed32, long=signed64.
- JS: int is really a double.
- Lua: int is really a double.
- PHP: Platform dependent, but on our architectures: signed64 and double above this range.
- Python: Bignum
- Ruby Bignum.

I'm pretty sure this is the best we can do. So let's put it to bed and charge on.

CC @snim2 

OK?

### DIFF
--- a/benchmarks/fasta/c/bench.c
+++ b/benchmarks/fasta/c/bench.c
@@ -43,7 +43,7 @@ const amino homosapiens[] = {
 /* Expected checksum tied to this specific scale factor */
 #define SCALE 10000
 #define EXPECT_CKSUM 9611973
-static u_int32_t checksum = 0;
+static u_int64_t checksum = 0;
 
 static unsigned rseed = START_RAND_SEED;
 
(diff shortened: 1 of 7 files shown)


ltratt:
I think we're actually using machine ints on Python, right? [Probably Ruby too.] Not that it matters in some ways.

I think this is ready for merging?

vext01:
Python and ruby use machine ints if possible, otherwise fall onto bignum.

Let's actually leave this open for a moment, as I've noticed binarytrees really does generate humongous numbers...

vext01:
I'm OK with the merge on this. Please go ahead

PR #43, opened 2015-10-20
More OpenBSD fixes

ltratt:
This also updates us to a much newer v8, which appears (appears!) to fix the intermittent build error we were previously encountering.

In essence, everything builds on OpenBSD except Graal/Truffle and HHVM. [PyPy currently doesn't build, but that's because of LibreSSL fallout. There should be a new PyPy release in the next week or so which fixes this, so it's not worth patching ourselves.]

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -20,6 +20,17 @@ check_for python
 check_for svn
 check_for unzip
 check_for xml2-config
+check_for bash
+# java versions need to be jdk7
+# we will build a jdk 8 later.
+# graal build needs both of these!
+check_for java
+check_for javac
+
+case `uname` in
+    OpenBSD) check_for eg++ ;;
+esac
+
 which pypy > /dev/null 2> /dev/null
 if [ $? -eq 0 ]; then
     PYTHON=`which pypy`
@@ -27,19 +38,12 @@ else
     PYTHON=`which python2.7`
 fi
 
-# java versions need to be jdk7
-# we will build a jdk 8 later.
-# graal build needs both of these!
-check_for java
-check_for javac
-
 which gmake > /dev/null 2> /dev/null
 if [ $? -eq 0 ]; then
     MYMAKE=gmake
 else
     MYMAKE=make
 fi
-check_for bash
 
 if [ $missing -eq 1 ]; then
     exit 1
@@ -94,7 +98,6 @@ build_cpython() {
 	cd cffi-${CFFI_V} && ${CPYTHON} setup.py install
 }
 
-
 LUAJITV=2.0.4
 build_luajit() {
 	cd ${wrkdir} || exit $?
@@ -121,8 +124,7 @@ build_pypy() {
 	rm -rf $usession
 }
 
-
-V8_V=4.5.38
+V8_V=4.8.90
 DEPOT_V=015cdc34ba4be808c47267123b0a97b93f5a0407
 DEPOT_REPO="https://chromium.googlesource.com/chromium/tools/depot_tools.git"
 build_v8() {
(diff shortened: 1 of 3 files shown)


ltratt:
I should probably also add that the v8 update changed the API used, so I had to update Edd's patch. These compile, but I haven't tried _running_ them, so it's possible they might break in use.

Tested on bencher5 and bencher6.

ltratt:
Edd, once this is merged, can you do a quick run with Krun on the results of a build to check that nothing's broken / changed hugely from what we saw before? Once per VM/benchmark combo is probably enough.

build.sh
```diff
@@ -20,6 +20,18 @@ check_for python
...
+check_for java
+check_for javac
+check_for xzdec
+
+case `uname` in
+    OpenBSD)
+        check_for egcc
+        check_for eg++
+	;;
+esac
```
vext01, line 34: I still think we need to build our own gcc. Will raise a bug for discussion.

build.sh
```diff
@@ -138,12 +141,19 @@ build_v8() {
 	# The build actually requires that you clone using this git wrapper tool
 	cd ${wrkdir}
 	OLDPATH=${PATH}
-	PATH=${wrkdir}/depot_tools:${PATH}
-	fetch v8 || exit $?
+	# v8's build needs python 2.7.10; as we've already built that, we might
+	# as well use it rather than forcing the user to install their own.
+	PATH=${wrkdir}/cpython-inst/bin:${wrkdir}/depot_tools:${PATH}
+	# XXX we should check for errors when fetching, but currently that
+	# causes problems because fetch runs a script which aborts on OpenBSD
```
vext01, line 149: Did you look into why this is?

build.sh
ltratt, line 34: I now agree, as Debian has GCC 4.9.2 and OpenBSD 4.9.3. But that's not related to this PR as such.

build.sh
```diff
@@ -20,6 +20,18 @@ check_for python
 check_for svn
 check_for unzip
 check_for xml2-config
+check_for bash
+check_for java
+check_for javac
+check_for xzdec
```
vext01, line 26: Add a check for wget for OpenBSD?

build.sh
ltratt, line 149: Yes https://code.google.com/p/v8/issues/detail?id=4500

build.sh
ltratt, line 26: Agreed (and pushed).

build.sh
vext01, line 34: That's not as bad a situation as I thought actually. Regardless, the C benchmarks at the very least need to be built with the same GCC. And if we go that far, we may aswell use this GCC to build the VMs. Will raise a bug.

build.sh
vext01, line 149: Ah, so v8 doesn't actually build? Or you need to run the build twice to get it to build? Or?

vext01:
Looks good bar minor issues I commented on. As you suggest, we need to test this with a few benchmark runs. I would also like to run the build on linux to check we didn't break anything.

build.sh
ltratt, line 149: The main VM builds fine; it's just a test that fails. So if you run "make" twice on OpenBSD, everything is fine. I'd prefer not to suppress this, so that we're reminded to check to see if future v8 versions fix the bug in question. [I must admit, it's unclear to me from the google code issue as to whether the v8 people might look into this themselves or not.]

ltratt:
It's already been tested on bencher5 and bencher6.

vext01:
Excellent! Shall we merge now and fix any benchmark fallout later, or test ahead of time?

ltratt:
I'd suggest merging now and fixing problems as they come up. [In particular, I'm curious whether the JDK change on bencher3 might have jiggled things around a bit, but that's probably not specific to this PR.]

vext01:
Would you like to squash anything?

ltratt:
No.

PR #44, opened 2015-10-20
Nbody checksums

vext01:
Checksums for nbody.

Uses a epsilon tolerance, as the double operations vary between languages. Not ideal, but the best we have.

Please check carefully. Some benchmarks were still mutating global state, and thus some of the benchmarks were changed a fair amount to cater with this.

OK?

### DIFF
--- a/benchmarks/nbody/c/bench.c
+++ b/benchmarks/nbody/c/bench.c
@@ -17,7 +17,10 @@
 #define days_per_year 365.24
 
 static double checksum = 0;
-#define EXPECT_CHECKSUM -0.33814926871151740339627167486469261348247528076171875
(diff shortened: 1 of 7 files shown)


snim2:
This is looking good to me, but I notice that in the C and Java benchmarks you define PI, but in the Ruby version you use the value from the standard library. Since the solar mass depends on PI I would guess this could effect all the subsequent calculations.

vext01:
Nicely spotted Sarah! I think we should use the a consistent PI for all benchmarks. This will be the first thing I do tomorrow.

vext01:
OK?

snim2:
Yup, is this ready to merge or do you want to squash?

vext01:
Good as is.

PR #46, opened 2015-10-21
Btrees checksums

vext01:
Checksums for btrees benchmarks.

You may recall I said that this benchmark makes huge checksums. It does not, my unsigned integer was underflowing. I've instead used a signed integer and I've checked that there are no over/underflow situations by (temporarily) replacing `+` and `-` on checksums with `Math.{add,subtract}Exact()` in Java. No exceptions were raised, therefore we have no over/underflows. These checks have now been removed.

Note that the original benchmark created three (unchecked) checksums. I make a single checksum by summing all three.

OK?

### DIFF
--- a/benchmarks/binarytrees/c/bench.c
+++ b/benchmarks/binarytrees/c/bench.c
@@ -15,9 +15,7 @@
 
 #define MIN_DEPTH 4
 #define MAX_DEPTH 12
-#define EXPECT_CKSUM 4294956382
-
-static u_int32_t checksum = 0;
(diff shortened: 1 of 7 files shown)


ltratt:
Just to be clear: the expectation is that the checksum continues to fit in 32/64 bits?

vext01:
Yes, it fits in a signed 32. I have confirmed that there is no overflow/underflow for the parameters under which we use the benchmark (MIN_DEPTH=4, MAX_DEPTH=12).

snim2:
Looks good to me.

PR #47, opened 2015-10-21
Richards checksums

vext01:
Richards already had predictable checksums and an outer loop like the one I have been adding to other benchmarks.

Just tidyings. No functional change.


snim2:
lgtm

PR #50, opened 2015-10-22
Remove conditional tracing from richards benchmarks.

vext01:
And fix an undefined variable use in PHP version.

I've commented them, as I imagine these will be useful to bring back without having to dig in version control.

OK?


snim2:
OK, lgtm

ltratt:
Unless there's a really good reason -- and there isn't here -- we mustn't commit commented out code. I've reopened the branch so that you can push a new patch that deletes those lines.

ltratt:
OK, so reopening pull requests doesn't do what I expected. So I'm deleting the branch again, but let's open a new PR that deletes the commented out code.

PR #52, opened 2015-10-30
Various updates (including PyPy 4.0 and OpenBSD fixes)

ltratt:
This is ready for comments, but not for merging. I am doing some final testing on bencher5 and bencher6 (should be done early this evening if I'm lucky).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ all: build-vms build-benchs bench
 build-vms:
 	./build.sh
 
-build-benchs:
+build-benchs: build-krun
 	cd benchmarks && \
 		${MAKE} CC=${CC} JAVAC=${JAVAC}
 
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,16 @@
 PYTHON ?= python2.7
 PWD != pwd
-JAVA_HOME = ${PWD}/work/openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image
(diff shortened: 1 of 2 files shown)


vext01:
Looks OK so far.

ltratt:
The build eventually dies on OpenBSD with http://pastebin.com/fyM4MY2A -- but I think this is partly to be expected (the krun config file is trying to do stuff with Graal, and that doesn't exist). I'll leave that for you to deal with in a separate PR :)

ltratt:
OK, this builds fine on bencher5 (until it complains about the lack of a tickless kernel) and bencher6 (modulo the error noted earlier). I think it can be merged.

vext01:
WRT the errors you are seeing, it's indeed because you don't have graal jvmci in the work dir.

I would expect a better error message though. Will look into that.

In the meantime, in the warmup experiment config file, you should put guards around the VMs you don't expect to work on openbsd.

ltratt:
Can I leave the guards to you (possibly in a separate PR)? I don't know how to do that.

vext01:
I can do it if you like, but it shouldn't be hard and it is better tested on the real openbsd environment.

Remember, the config file is just a Python script, so you can add `if not sys.platform.startswith(...)` guards to prevent definitions of VMs.

ltratt:
You can do this about 10x faster than me -- I've never really looked at the config file! You can tack it onto this PR if you want and I'll test it. [Sarah would then need to do the merging.]

PR #57, opened 2015-11-09
measuring startup

cfbolz:
the goal is to retain all the krun features but without having to change
krun itself in major ways. startup.krun documents the patching that it
does to achieve that.

### DIFF
--- a/startup.krun
+++ b/startup.krun
@@ -1,0 +1,173 @@
+import os
+from krun.vm_defs import (PythonVMDef, LuaVMDef, JavaVMDef, GraalVMDef,
+    PHPVMDef, JRubyTruffleVMDef, V8VMDef, NativeCodeVMDef,
+    find_internal_jvmci_java_bin, GenericScriptingVMDef)
+from krun import EntryPoint
+
+# this file is a minimally hacky way to do startup measurements with all the
+# krun goodies, but without changing krun.
+#
+#
+# the way it works is as follows:
+#
+# before the invocation of the VM, another executable is inserted:
+#
+# sudo -u krun nice -20 taskset 0x2 env <some variables> \
+# outer_startup_runner_c jruby startup_runner.rb
+#
+# outer_startup_runner_c is a small C program that prints the current time,
+# then execs its arguments. startup_runner.rb will then also print the current
+# time and exit. The warmup time is therefore the difference of the two (both
+# of these times are stored in the JSON file, the substraction happens in
+# post-processing). That way, all the extra commands are not included in the
+# measurement, but we still get the krun goodies.
+#
+# this works by patching two things:
+# - the runner file for each VM is patched to point to the startup runner
+#   instead (function patch_runner). The startup runners are in
+#   startup_runners/
+# - the bench_cmdline_adjust method is patched to insert the call to
+#   outer_startup_runner_c
+#
+
+# Who to mail
+MAIL_TO = []
+
+# Maximum number of error emails to send per-run
+#MAX_MAILS = 2
+
+DIR = os.getcwd()
+JKRUNTIME_DIR = os.path.join(DIR, "krun", "libkruntime", "")
+JDK8_HOME = os.path.join(DIR, "work/openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image/")
+STARTUP_RUNNER_DIR = os.path.join(DIR, "startup_runners")
+
+HEAP_LIMIT = 2097152  # K == 2Gb
+
+# Variant name -> EntryPoint
+VARIANTS = {
+    "default-c": EntryPoint("bench.so", subdir="c"),
+    "default-java": EntryPoint("KrunEntry", subdir="java"),
+    "default-lua": EntryPoint("bench.lua", subdir="lua"),
+    "default-python": EntryPoint("bench.py", subdir="python"),
+    "default-php": EntryPoint("bench.php", subdir="php"),
+    "default-ruby": EntryPoint("bench.rb", subdir="ruby"),
+    "default-javascript": EntryPoint("bench.js", subdir="javascript"),
+}
+
+ITERATIONS_ALL_VMS = 1
+
+def patch_runner(vm_def):
+    from krun.env import EnvChangeAppend
+    if vm_def.iterations_runner == "IterationsRunner":
+        vm_def.iterations_runner = "startup_runner"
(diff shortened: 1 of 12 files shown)


startup.krun
```diff
@@ -0,0 +1,162 @@
...
+    "default-ruby": EntryPoint("bench.rb", subdir="ruby"),
+    "default-javascript": EntryPoint("bench.js", subdir="javascript"),
+}
+
+ITERATIONS_ALL_VMS = 1
+
+def patch_runner(vm_def):
+    from krun.env import EnvChangeAppend
+    if vm_def.iterations_runner == "IterationsRunner":
+        #import pdb; pdb.set_trace()
```
vext01: debugging code

startup.krun
```diff
@@ -0,0 +1,162 @@
...
+    return vm_def
+
+from krun.platform import DebianLinuxPlatform, BasePlatform
+
+def bench_cmdline_adjust(self, args, env_dct):
+    prepend_args = BasePlatform.bench_cmdline_adjust(self, [], env_dct)
+    if "check_user_change" in args[-3]:
+        return prepend_args + args
+    return prepend_args + ["krun/iterations_runners/startup_runner_runner_c"] + args
+DebianLinuxPlatform.bench_cmdline_adjust = bench_cmdline_adjust
```
vext01: Does this only work on Debian, or is the approach (in principal) platform specific? I imagine Laurie will want to measure startup on OpenBSD.

startup.krun
```diff
@@ -0,0 +1,162 @@
...
+        assert "iterations_runner." in vm_def.iterations_runner
+        _, suffix = vm_def.iterations_runner.rsplit(".", 1)
+        vm_def.iterations_runner = os.path.join(STARTUP_RUNNER_DIR, "startup_runner." + suffix)
+    return vm_def
+
+from krun.platform import DebianLinuxPlatform, BasePlatform
+
+def bench_cmdline_adjust(self, args, env_dct):
+    prepend_args = BasePlatform.bench_cmdline_adjust(self, [], env_dct)
+    if "check_user_change" in args[-3]:
```
vext01: Why do we need this? Can we add a comment here?

startup.krun
```diff
@@ -0,0 +1,162 @@
...
+        return prepend_args + args
+    return prepend_args + ["krun/iterations_runners/startup_runner_runner_c"] + args
+DebianLinuxPlatform.bench_cmdline_adjust = bench_cmdline_adjust
+
+VMS = {
+        'C': {
+                'vm_def': patch_runner(NativeCodeVMDef()),
+                'variants': ['default-c'],
+                'n_iterations': ITERATIONS_ALL_VMS,
+                'warm_upon_iter': 0,
```
vext01, line 99: AFAIK, `warm_upon_iter` is not used. I have been meaning to remove all traces of this from the krun repo.

startup.krun
```diff
@@ -0,0 +1,162 @@
...
+    #"*:Hotspot:*",
+    #"*:Graal:*",
+    #"*:LuaJIT:*",
+    #"*:HHVM:*",
+    #"*:JRubyTruffle:*",
+    #"*:V8:*",
+]
+
+N_EXECUTIONS = 200  # Number of fresh processes.
+N_GRAPHS_PER_BENCH = 2
```
vext01, line 173: This is also an unused config variable.

startup.krun
```diff
@@ -0,0 +1,162 @@
...
+# this file is a minimally hacky way to do startup measurements with all the
+# krun goodies, but without changing krun.
+#
+#
+# the way it works is as follows:
+#
+# before the invocation of the VM, another executable is inserted:
+#
+# sudo -u krun nice -20 taskset 0x2 env <some variables> \
+# startup_runner_runner_c jruby startup_runner.rb
```
vext01: It took me a while to realise that there is a `startup_runner_c` and a `startup_runner_runner_c`. I wonder if we can find a better naming convention.

startup_runners/startup_runner.rb
```diff
@@ -0,0 +1,15 @@
+if /linux/ =~ RUBY_PLATFORM
```
vext01: This needs the same hack as we just applied to the regular `iterations_runner.rb`. I.e. on JRuby we should call to `clock_gettime_monotonic()` in `libkruntime`.

startup_runners/startup_runner_runner.c
```diff
@@ -0,0 +1,38 @@
...
+
+/* To correctly expose asprintf() on Linux */
+#define _GNU_SOURCE
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <limits.h>
+#include <errno.h>
+#include <string.h>
+#include <dlfcn.h>
```
vext01: Some of these includes are redundant. Certainly `dlfcn.h`, maybe `limits.h` and `string.h`.

startup.krun
cfbolz: `outer_startup_runner_c`?

startup.krun
cfbolz: yeah, should work, though the C file might need adaptations.

startup.krun
cfbolz: will add one.

cfbolz: The basic ideas is that all the platform checks are run as benchmarks krun. That means their output needs to be valid JSON. This does not work with the startup_runner_runner_c for the check_user_change platform check.

startup.krun
cfbolz, line 99: ok, it's copy-pasted from warmup.krun, will kill.

startup.krun
cfbolz, line 173: ditto

startup_runners/startup_runner.rb
cfbolz: Good catch (missed a push from bencher3).

startup.krun
vext01: OK, so I wonder if we can patch the right platform class automatically.  One idea: we use `detect_platform()` in `platform.py` then ask the return value for it's type. This is then the class we patch. What do you think?

startup.krun
vext01, line 99: Yes, I've been slacking on that front. I will make a PR to eliminate them now.

startup.krun
cfbolz: works for me, will do that

vext01:
This all looks good to me. Do you want to do any squashing?

cfbolz:
squashed

PR #58, opened 2015-11-10
Remove old config vars

vext01:
Also revise the C SKIP section. All of the C benchmarks are now implemented, so there is no need for any default SKIPs to be enabled.


cfbolz:
lgtm

PR #63, opened 2015-12-14
Fix the mx version for graal.

vext01:
This fixes the mx version we use to that of the graal version we use. The build now passes on bencher5.

I'm not sure why bencher3 built graal without this, whereas bencher5 failed. I'm going to re-run the build again to check that the failure isn't non-deterministic.

Either way, we certainly should fix the mx version we use.

[As you can see we have already gone through similar problems to our v8 problen with graal and mx (expand the lines just under the changes)]

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -251,6 +251,7 @@ build_jdk() {
 }
 
 MX_REPO=https://bitbucket.org/allr/mx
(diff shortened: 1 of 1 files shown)


ltratt:
Yes, `mx` should have a fixed version -- good spot!

PR #64, opened 2015-12-15
Attempt to fix v8 build.

ltratt:
Don't merge quite yet (a full build is underway). But if you spot anything dumb, please let me know!

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -150,8 +150,7 @@ build_pypy() {
 	fi
 }
 
-V8_V=4.8.90
-DEPOT_V=015cdc34ba4be808c47267123b0a97b93f5a0407
+V8_V=4.8.271.9
 DEPOT_REPO="https://chromium.googlesource.com/chromium/tools/depot_tools.git"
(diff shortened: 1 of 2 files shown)


vext01:
Looks OK. Can we check the result of `git checkout` and `gclient sync` or do they also fail like `fetch` does?

ltratt:
They can also fail :( Though it's not ideal, I think we're OK, because the later build can't possibly succeed if the earlier things fail -- and the later build is checked properly. So, at worst, we pick up the error a bit later than we'd like.

vext01:
Fair enough.

vext01:
Should we wait until we hear back from Jakob? I have a suspicion that the 4.8 branch is not stable code. But then again, nor is graal or jruby/truffle.

ltratt:
If he comes back and tells us we should be using another tag, we can merge that later. [I'm not sure what timezone he's operating in.]

ltratt:
It works! So it's ready for merging.

PR #65, opened 2015-12-15
Use egcc to translate PyPy on OpenBSD. Tidy up.

vext01:
- Use egcc on OpenBSD.
- Tidy up such that function can be re-run without re-fetching etc.

I've used tabs to fall in line with the rest of the file. Should we do a tabs to spaces sweep after this is in?

OK?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -121,14 +121,33 @@ PYPYV=4.0.0
 build_pypy() {
 	cd ${wrkdir} || exit $?
 	echo "\n===> Download and build PyPy\n"
-	if [ -f ${wrkdir}/pypy/pypy/goal/pypy-c ]; then return; fi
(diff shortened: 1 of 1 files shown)


ltratt:
I didn't know that `!` could appear outside of the `[...]` tests, but assuming that's legitimate, all looks good.

You're right, we should do a tabs -> spaces sweep, but only after the v8 PR is in, pretty please :)

vext01:
I've never thought about the placement of the `!` actually. I just did what I've always done. Quick search suggests placement doesn't matter in this case: http://stackoverflow.com/questions/26475358/negate-if-condition-in-bash-script

Today we learned :)

Ack for tab sweep. No-one like merge conflicts.

PR #69, opened 2015-12-18
Tidy

vext01:
Small tidy ups and improvements to the warmup experiment. See individual commit messages for details.

OK?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,9 @@ work
 **/*.swp
 **/*.pyc
 **/*.class
+**/*.so
 warmup_results.json
 benchmarks/richards/java/Benchmark.java
 benchmarks/richards/java/Program.java
 benchmarks/richards/java/richards.java
+benchmarks/richards/java/richards.java.orig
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -4,7 +4,9 @@ work
 **/*.swp
 **/*.pyc
 **/*.class
+**/*.so
 warmup_results.json
 benchmarks/richards/java/Benchmark.java
 benchmarks/richards/java/Program.java
 benchmarks/richards/java/richards.java
+benchmarks/richards/java/richards.java.orig
```
ltratt, line 12: I don't think we should be excluding `.orig` files, because we don't ever generate them, right?

ltratt:
Just one minor comment, then we can merge.

.gitignore
vext01, line 12: `build.sh` creates this file when applying `java_richards.patch`, so we do create this file.

ltratt, line 12: Ah, this is due to a BSD `patch` oddity :( OK...

PR #70, opened 2015-12-18
Tiny cleanups

vext01:
A few more small items.

OK?

### DIFF
--- a/startup_runners/startup_runner.c
+++ b/startup_runners/startup_runner.c
@@ -21,7 +21,7 @@
 double clock_gettime_monotonic();
 
 int
-main(int argc, char **argv)
+main(void)
 {
     double    start_time = -1;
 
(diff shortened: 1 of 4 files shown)


startup_runners/startup_runner.c
```diff
@@ -25,6 +25,10 @@ main(int argc, char **argv)
 {
     double    start_time = -1;
 
+    /* silence warnings */
+    argc = argc;
+    argv = argv;
```
ltratt: I assume that this program doesn't use `argc` or `argv` in which case it would be better to use `int main(void)` as the declaration.

ltratt:
One minor comment, then we can merge.

startup_runners/startup_runner.c
vext01: Didn't know you could do that. Fixed.

vext01:
Can I squash?

ltratt:
Yep!

vext01:
Kersplat!

PR #71, opened 2015-12-18
Fix .PHONY

vext01:
OK?


ltratt:
I think I see the problem: shouldn't it be `.PHONY`?

ltratt:
See https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html#Phony-Targets

vext01:
oh, oh, oh! :P Woops.

vext01:
Fixed. Can I squish?

ltratt:
Just to confirm: this definitely solves the problem that running make ran the wrong rules? Assuming so, squash, and then I'll merge.

vext01:
Correct -- squashed -- merge at will.

PR #75, opened 2015-12-21
improve performance

cfbolz:
I switched from a few explicit loops to doing stuff within numpy code, which greatly speeds stuff up.


snim2:
OK, LGTM

PR #79, opened 2016-01-12
Add a switch that allows us to zoom in on a sample range.

vext01:
Example usage:

```
python2.7 plot_krun_results.py -f ~/warmup_xmas_results/bencher3/warmup_results.json.bz2 -l -1 -z 10,20
```

Gives graphs like this:
zoom

OK?


plot_krun_results.py
```diff
@@ -448,9 +464,9 @@ def create_cli_parser():
     plt.close()  # avoid extra blank window
     if not options.interactive:
         print 'Saving results to: %s' % options.outfile
-    print (('Charting with sliding window size: %d, confidence level: %d ' +
+    print (('Charting with sliding window size: %d, zoom: %s, confidence level: %d ' +
             'bootstrap samples: %d and percentiles: %s') %
```
snim2: This is a bit pedantic, but just to be clear can we say something like "xlimits: %s" rather than "zoom"?

plot_krun_results.py
```diff
@@ -435,6 +447,10 @@ def create_cli_parser():
                         help=('Only draw charts for a specific benchmark/' +
                               'vm/variant triplet. ' +
                               'e.g. binarytrees:Hotspot:default-java'))
+    parser.add_argument('--zoom', '-z',
+                        help=("Zoom in on sample range "
+                              "(comma separated pair 'start,end'. Samples "
+                              "start from 0, upper bound is exclusive)"))
     return parser
```
snim2: I think it's worth giving an example here. e.g. `Zoom in on a small sample of the measurements. e.g. '--zoom 100,250' displays measurements in the range x in [100, 250).` or some such.

snim2:
OK, apart from those very minor comments this looks good to me. Do you want to keep this as-is or add the "zoomed in" charts as insets?

vext01:
The last commit reworks the terminology and adds an example in the help text.

re: insets, let's leave it like this for now. I would like to have the draft finished this week and matplotlib is a time sink.

snim2:
That looks a lot nicer. Are you happy to squash?

ltratt:
Agreed -- we don't want to fiddle with inserts at this point.

vext01:
Splat. Go ahead and merge.

snim2:
Ta

PR #80, opened 2016-01-12
single exec support

vext01:
Adds the ability to plot a single execution.

Also a couple of other tweaks.

There's a bug (not introduced by this): if you pass a mis-spelled key to `-b` then you get a bogus PDF, as the key never matches. I would expect an error. I started hacking, but the structure of the outer loops made it awkward and it was taking too long. I think those loops could do with a refactor some time.

OK for the changes I did make?


snim2:
When you say "a bogus PDF", does it give you a PDF with some random benchmark or a PDF with no plots?

vext01:
A corrupted PDF. Both mupdf and xpdf choke.

snim2:
Can we add something like this, just under the line `from matplotlib.backends.backend_pdf import PdfPages` in `main()`:

``` python
        all_benchmark_keys = set()
        for machine in data_dcts:
            all_benchmark_keys.add(data_dcts[machine]['data'].keys())
            if benchmark is not None and not (benchmark in all_benchmark_keys):
                print 'Benchmark %s does not appear in any of the data files that have been loaded.'
                sys.exit(1)
```

plot_krun_results.py
```diff
@@ -452,6 +460,10 @@ def create_cli_parser():
                               "'start,end'. Samples start from 0. e.g. "
                               "'-x 100,130' will show samples in the range "
                               "100 to 130."))
+    parser.add_argument('--single-exec', '-e',
+                        type=int,
+                        help=("Emit only a single execution. "
+                              "e.g. '-e 0' emits only the first."))
```
snim2, line 466: This is fine, but are we still calling these "executions"?

vext01:
A similar hack would need to be implemented in interactive mode too. I think ideally these two loops need to be merged, and instead of looping over all available keys, we should iterate over a singleton list (in the case of -b), or the union of all of the keys found in `-f` files.

It's a little awkward, so I suggest we hold off for now. after all, it will be obvious if the key was mis-typed, as you just wont get what you wanted.

plot_krun_results.py
vext01, line 466: If we want to call them something else we should update all the variable names and the graph titles also.

snim2:
hmm. OK, do you want to add an issue for that and we can get to it when we have more time?

plot_krun_results.py
snim2, line 466: ugh. That's fiddly. Another one for an issue I think.

snim2:
@vext01 OK, since we are just merging this into a branch I'm happy to merge if you are?

vext01:
Im OK to merge this.

PR #84, opened 2016-01-18
New charts with outliercounts

snim2:
- New script `mark_outliers_in_json` to add outliers to Krun results files
- New script `outlier_summaries` to generate LaTeX documents with a summary of outliers per benchmark, VM and language variant
- Changes to `plot_krun_data` to allow outliers to be annotated in plots, with the new `--with_outliers` switch


mark_outliers_in_json.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def find_outliers(data, window_size):
+    outliers = list()
+    for index, datum in enumerate(data):
+        l_slice, r_slice = _clamp_window_size(index, len(data), window_size)
+        window = data[l_slice:r_slice]
+        mean = numpy.mean(window)
+        five_sigma = 5 * numpy.std(window)
+        if datum > (mean + five_sigma):
```
cfbolz: should we also consider outliers that are exceptionally slow?

vext01: I think we should consider both fast and slow spike iterations as outliers.

cfbolz:
thanks

snim2:
Thanks!

PR #85, opened 2016-01-22
Remove suptitle.

vext01:
Remove the suptitle. It overlaps the graph headers and says nothing that isn't already said in the individual graph headers.

OK?


snim2:
I agree that removing the suptitle is a good idea.

At some point today I need to push the changes which add percentage outliers to the charts, separately label the unique and common outliers, and draw in the mean and 5-sigma. 

Do you want to merge this now, then have me pull that into the new changes and merge it in, or should we close this PR now and I'll add this to the upcoming PR?

vext01:
Whatever is easiest for you really. Please make sure that all the statistical stuff can be switch off at the command line, as the draft won't include any statistics.

Thanks

snim2:
It already can, don't worry.

vext01:
So up to you if you want to merge this now.

snim2:
I've added this to my other branch and will push in a minute...

vext01:
ok, thanks

Issue #81, opened 2016-01-12
[new_charts] refactor outer loop


vext01:
As mentioned previously, if you pass a non-existing key to the plotting script's `-b`, then no graphs are processed.

Before we fix this, I recommend refactoring the outer loops, otherwise we have to apply a hack twice.

snim2:
I'm not totally convinced that merging the loops makes sense. In order to produce the large PDF the `with PdfPages(outfile) as pdf:` must be around the calls to `savefig()`...

vext01:
We don't have to use the context manager if we don't want to. You can probably do something like:

```
if not interactive:
    pdf = PdfPages(outfile)

... common logic

if not interactive:
    pdf.close()
```

The script should certainly bomb out if the key passed to `-b` is a typo -- i.e. that isn't a key that appears in the json file at all.

vext01:
Fixed

PR #86, opened 2016-01-22
New charts differentiated outliers

snim2:
This PR:
- Removes the `suptitle` from charts
- Only uses the `TkAgg` backend to matplotlib in interactive mode (this means we can draw charts on headless machines, such as the benchers)
- Separately labels "unique" and "common" outliers on charts, and adds percentages to the legend
- Removes the old `outlier_counter` script which we are no longer using
- Removes percentiles and CIs from plots, since we are no longer using these
- Allows the user to plot the rolling mean and 5-sigma

The new CLI is as follows:

``` bash
$ ./plot_krun_results.py --help
Example usage:

$ python plot_krun_results.py -f results1.json.bz2
$ python plot_krun_results.py -i -f results1.json.bz2 -f results2.json.bz2 --window 250 --mean --sigma --with-outliers
$ python plot_krun_results.py -b  binarytrees:Hotspot:default-java
       [-h] [--interactive] [--filename JSON_FILES] [--outfile OUTFILE]
       [--window WINDOW_SIZE] [--benchmark BENCHMARK] [--xlimits XLIMITS]
       [--single-exec SINGLE_EXEC] [--mean] [--sigma] [--with-outliers]
       [--with-unique-outliers]

optional arguments:
  -h, --help            show this help message and exit
  --interactive, -i     Display graphs (rather than writing to a file)
  --filename JSON_FILES, -f JSON_FILES
                        Krun results file. This switch can be used repeatedly
                        to chart data from a number of results files.
  --outfile OUTFILE, -o OUTFILE
                        Name of the PDF file to write to.
  --window WINDOW_SIZE, -w WINDOW_SIZE
                        Size of the sliding window used to draw percentiles.
  --benchmark BENCHMARK, -b BENCHMARK
                        Only draw charts for a specific benchmark/vm/variant
                        triplet. e.g. binarytrees:Hotspot:default-java
  --xlimits XLIMITS, -x XLIMITS
                        Specify X-axis limits as a comma separated pair
                        'start,end'. Samples start from 0. e.g. '-x 100,130'
                        will show samples in the range 100 to 130.
  --single-exec SINGLE_EXEC, -e SINGLE_EXEC
                        Emit a graph for a single process execution. e.g. '-e
                        0' emits only the first.
  --mean, -m            Draw a rolling mean.
  --sigma, -s           Draw 5-sigma interval around the rolling mean.
  --with-outliers       Annotate outliers. Only use this if your Krun results
                        file contains outlier information.
  --with-unique-outliers
                        Annotate outliers common to multiple executions and
                        those unique to single executions differently.
```


mark_outliers_in_json.py
```diff
@@ -35,27 +35,52 @@ def main(in_files, window_size):
...
-def find_outliers(data, window_size):
-    outliers = list()
+def get_outliers(all_outliers, window_size, threshold=1):
+    common, unique = list(), list()
+    for index, outliers in enumerate(all_outliers):
+        common_exec = list()
+        unique_exec = list()
+        for outlier in outliers:
+            other_execs = all_outliers[:index] + all_outliers[(index + 1):]
+            if sum(map(lambda lst: outlier in lst, other_execs)) > threshold:
```
cfbolz: I have two issues with this line: a) I think it should be `... >= threshold` otherwise 2 other executions are needed, not 1 for threshold = 1.

b) as usual I find this kind of code pretty hard to wrap my head around. Why not simply

```
sum = 0
for execution in other_execs:
    sum += outlier in execution
if sum > threshold:
    ...
```

snim2:
OK, I think I have responded to the comments so far. Let me know when we are ready to squash this.

cfbolz:
Yes, looks good to me. Squash then I'll merge?

snim2:
OK, let's keep it like this. There will need to be a bigger merge on the other branch.

plot_krun_results.py
```diff
@@ -77,9 +79,9 @@
 DPI = 300
 
 
-def main(is_interactive, data_dcts, percentiles, window_size,
-         level, num_samples, outfile, xlimits, single_exec,
-         with_outliers, benchmark=None):
+def main(is_interactive, data_dcts, window_size, outfile, xlimits, single_exec,
+         with_outliers, unique_outliers, mean=False, sigma=False,
+         benchmark=None):
```
vext01, line 84: On a related note, #81 describes a refactoring and an improvement which could be made to main() here.

snim2, line 84: OK, maybe we can apply that change to the new_charts branch before merging into master?

snim2:
Thanks

PR #87, opened 2016-01-22
Larger fonts

vext01:
Larger fonts, to make the graphs in the paper readable.

fonts

OK?


snim2:
Did you test this with interactive mode and with a whole page of charts yet?

vext01:
I tested a whole page of charts, hence the spacing adjustment. I'm not too fussed about interactive mode right now, just trying to hammer the paper draft into shape.

snim2:
OK, testing this now...

snim2:
OK, I guess you tested this with measurements only, which is what will go in the paper. On both my machines, if I turn on the mean and sigma plotting the text overlays the charts in all sorts of places, if I only plot data I get something like this:

larger_fonts

vext01:
Why are the plots so narrow?

snim2:
Which axis?

vext01:
Y. The graph heights are surprisingly short. Not sure why...

snim2:
I think the first iteration is covered by the axis itself?

vext01:
OK, I just realised you are in interactive mode.

In this mode, it tries to fit all execs in the window causing overlap. The change I just pushed reverts the old font size only for interactive mode, but I still see overlap even on my large monitor and maximised window.

This is a separate bug. We have never used interactive mode with so many executions.

snim2:
OK, testing this again.

snim2:
OK, so the interactive mode is better, still a bit of overlap but readable. The non-interactive mode is OK, but the legend goes outside the chart, and even if you only plot data the legend text is very large. Here is an example: 
larger_fonts_legend

vext01:
OK, I think the legend text needs to:
- Be shorter
- Be in one column

Do you mind if we fix that separately so I can push on with the draft, which will not include a legend?

snim2:
OK, that's fine. BTW one column will mean that the legend text is always written over the data, which is why I increased the column count. So, I think we will need to play around quite a bit before we get something that we are happy with. What I would have preferred was to have just one legend, outside of all the plots, but I couldn't figure out how to do it.

Do you want to squash this PR before I merge?

vext01:
splat.

vext01:
Thanks!

PR #88, opened 2016-01-25
Refactor main loops. 

### LINKED ISSUE #81: [new_charts] refactor outer loop

snim2:
Replace assertions with custom error function. Add error message if argument to -b does not appear in Krun results files. [plot_krun_results]

Fixes #81


plot_krun_results.py
```diff
@@ -85,74 +86,64 @@ def main(is_interactive, data_dcts, window_size, outfile, xlimits, single_exec,
...
-                                                     mean,
-                                                     sigma)
-                        fig.set_size_inches(*export_size)
-                        pdf.savefig(fig,
-                                    dpi=fig.dpi, orientation='landscape',
-                                    bbox_inches='tight')
-                        plt.close()
-        print('Saved: %s' % outfile)
+                else:
+                    fig, export_size = draw_page(is_interactive,
```
vext01: Do we need the call to drawpage() twice? Perhaps we can always assign the return values, but only use them in one mode?

vext01:
Looks good apart from one minor nit I commented upon.

vext01:
(tested on bencher2 headless+PDF and with interactive/X-forwarding)

snim2:
OK, just testing a fix for that comment.

snim2:
Ready to squash?

vext01:
Looks good, go ahead.

snim2:
Splat.

vext01:
Many thanks!

PR #96, opened 2016-01-29
Reorder charts

snim2:
Format pages in benchmark order.

Also formats y-ticks to 3dp.


vext01:
Looks good and works for me. Any squashing?

snim2:
I think it probably makes sense to keep those commits separate.

vext01:
Thanks so much

PR #100, opened 2016-02-05
Remove unused benchmarks and other miscellaneous bits.

vext01:
This whole repo was based on the VM experiment. There's a lot of unused stuff in the `benchmarks/` directory. Mostly benchmarks which were never ported to Krun. There's an unused `php.ini` and some random ruby file. This PR removes them.

If we plan to port more benchmarks later, we can re-import.

OK?


ltratt:
I approve!

PR #101, opened 2016-02-05
Put "Print-tracing" Versions of Benchmarks into Master Branch

vext01:
This PR moves all of the tracing versions of benchmarks into the master branch and updates the documentation making users aware of their existence. While I was in `README.md` I also linked to the paper and said roughly how to run the experiment.

@cfbolz Can you let me know if the file (in your branch) `benchmarks/binarytrees/ruby/trace_debug_bench.rb` should be included. I think it should not be included as we already have `benchmarks/binarytrees/ruby/trace_bench.rb`, right?

OK?

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,45 @@
-# Warmup Experiment
+# SoftDev Warmup Experiment
 
-This is a work in progress. Watch this space.
+This is the main repository for the Software Development Team Warmup Experiment
+as detailed in the paper "Virtual Machine Warmup Blows Hot and Cold", by Edd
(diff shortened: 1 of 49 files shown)


cfbolz:
I agree that the file can go. if we add more benchmarks it might be useful again (maybe), but in that case we can fish it from history.

vext01:
OK, so if all looks in check, go ahead and merge.

ltratt:
Is there a reason why this wasn't merged?

ltratt:
[I just quickly did the rebase -- though I haven't pushed! -- and there's a single, rather trivial, conflict to fix. So I think we should rebase then merge.]

PR #105, opened 2016-02-18
Use our own GCC.

vext01:
_Alert the media! We have managed to bootstrap GCC-4.9.3 and use it throughout!_

Our gcc is installed as `zgcc` and `zg++`. I was able to get most VMs to use these using environment or make variables. However, it seems a couple of VMs (HHVM and Graal) cannot be swayed. HHVM fails to pass down the correct GCC to Ocaml (for the offline type checker) and Graal has gcc hard-coded (confirmed with Chris Seaton). For these I make temporary symlinks to our `zgcc` and force our bin path first.

The build succeeds on OpenBSD and Linux. During testing, after bootstrapping GCC, I crippled the system compilers (`chmod -x`) before continuing. This confirms that the system GCC isn't being invoked.

Getting GCC bootstrapped on OpenBSD was a challenge. For future reference, here are bug reports relating to this:
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69757
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69744

I've also fixed a some bogus stuff that seems to have crept into `warmup.krun`. The "dummy" paths were not supposed to have been pushed into master. Similarly for the `MAIL_TO` lines.

OK?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,9 @@
 #! /bin/sh
 
-# This assumes a linux x86_64 machine. We used Debian 8.
+unknown_platform() {
+    echo "Unknown platform: `uname`"
+    exit 1
+}
 
 missing=0
 check_for () {
@@ -27,14 +30,6 @@ check_for xzdec
 check_for wget
 
 case `uname` in
-    OpenBSD)
-        check_for egcc
-        check_for eg++
-    ;;
-esac
-
-
-case `uname` in
     Linux*) PATCH_ARGS=--backup
 esac
 
@@ -45,6 +40,7 @@ else
     PYTHON=`which python2.7`
 fi
 
+# This all requires GNU make and assumes you have it installed.
 which gmake > /dev/null 2> /dev/null
 if [ $? -eq 0 ]; then
     MYMAKE=gmake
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -74,6 +65,59 @@ fetch_krun() {
...
+        --enable-languages=c,c++ \
+        --with-system-zlib \
+        --disable-tls \
+        --enable-threads=posix \
+        --enable-wchar_t \
+        --disable-libstdcxx-pch \
+        --enable-cpp \
+        --enable-shared \
+      || exit $?
+    $MYMAKE -j4 || exit $?
```
ltratt, line 122: I wonder if we should be less aggressive with `-j`? If machines have low-ish RAM, we might accidentally force machines into swap?

build.sh
```diff
@@ -175,19 +210,21 @@ build_v8() {
...
+
+    # The build fails for silly reasons near the very end, even though the main
+    # v8 binary has been built. So we simply check that the binary exists and
+    # suppress unrelated build errors.
+    # Bug report https://code.google.com/p/v8/issues/detail?id=4500
+    #
+    # This used to be only the case on OpenBSD, but since we started building
+    # our own gcc, the tests also fail in strange ways on linux.
+    #
+    # V8 also mistakes our compiler for clang for some reason.
```
ltratt: This last comment should make clear that's why we do `GYP_DEFINES`.

build.sh
```diff
@@ -225,10 +262,10 @@ build_jdk() {
     xzdec ${JDK_DIST} | tar xf - || exit $?
     mv ${JDK_INNER_DIR} openjdk
     cd openjdk || exit $?
-    JDK_BUILD_PATH=${wrkdir}/make-${GMAKE_V}:${PATH}
+    JDK_BUILD_PATH=${wrkdir}/make-${GMAKE_V}:`dirname ${OUR_CC}`:${PATH}
     case `uname` in
         Linux)
-	    PATH=${JDK_BUILD_PATH} bash configure \
+	    env CC=zgcc CXX=zg++ PATH=${JDK_BUILD_PATH}  bash configure \
```
ltratt: Why an extra space before bash?

build.sh
```diff
@@ -410,7 +464,18 @@ fetch_libkalibera() {
     fi
 }
 
+
 fetch_external_benchmarks
+build_gcc
+
+# Put GCC libs into linker path
+# Needed for (e.g.) V8 to find libstdc++
+if [ `uname` = "Linux" ]; then
```
ltratt: This should be a `case ... esac` so that we can easily add other OSs later (the default branch should probably error?).

warmup.krun
```diff
@@ -6,7 +6,7 @@ from krun.vm_defs import (PythonVMDef, LuaVMDef, JavaVMDef, GraalVMDef,
 from krun import EntryPoint
 
 # Who to mail
-MAIL_TO = ["vext01@gmail.com", "mount.sarah@gmail.com"]
+MAIL_TO = []
 
 # Maximum number of error emails to send per-run
 #MAX_MAILS = 2
```
ltratt: Why is this line commented? [I know it isn't changed, but I only just spotted it!]

warmup.krun
```diff
@@ -21,8 +21,13 @@ else:
 
 JDK8_BIN = os.path.join(JDK8_HOME, "bin", "java")
 
-HEAP_LIMIT = 2097152  # K == 2Gb
-STACK_LIMIT = 8192  # K
+if sys.platform.startswith("linux"):
+    GCC_ENV = {"LD_LIBRARY_PATH": os.path.join(DIR, "work/gcc-inst/lib64")}
+else:
```
ltratt: This should probably be `elif sys.platform.startswith("openbsd")` and the actual `else` branch should raise a `NotImplementedException` or similar.

ltratt:
A few minor comments only to look at.

build.sh
vext01, line 122: Since we are building PyPy, I think we can assume we have plenty of RAM?

build.sh
vext01: Typo

build.sh
ltratt, line 122: Who knows how much RAM is required to compile this? At least with PyPy, it's a single process so we can easily monitor it. Personally, I wish `make` just chose a sensible number of jobs itself :/

warmup.krun
vext01: Not sure what you mean.

ltratt: The `#MAX_MAILS` line should probably be deleted.

vext01: Ah, it's just there to show you can change it if you like. We could just remove it tbh.

warmup.krun
vext01: I was working under the assumption that we only support OpenBSD and linux, but I'm OK with the change.

warmup.krun
ltratt: It looks pretty random to me, so killing it won't hurt?

build.sh
vext01, line 122: Gmake chooses a suitable -j, BSD make does not. PyPy needs about 8GB of RAM.

ltratt, line 122: Maybe we should use `gmake` here and leave it to do whatever's sensible?

vext01, line 122: Come to think of it ${MYMAKE} should be gmake on OpenBSD. I will improve the check though, it's not very robust.

ltratt, line 122: Why don't we do a separate commit / PR that changes `MYMAKE` to `gmake` on OpenBSD and maybe also calls the variable something sensible like `GMAKE`?

vext01, line 122: Can do.

vext01:
I think that's everything. OK?

ltratt:
Let's squash, then I can merge.

vext01:
Splat.

PR #108, opened 2016-02-20
Missing OpenBSD gcc patches.

vext01:
These are the missing patches for OpenBSD/gcc.

Cheers

PR #107, opened 2016-02-18
Clean up usage of gmake.

vext01:
Don't merge yet. Needs a rebuild on Linux and OpenBSD to check for problems.

Looks OK?

PR #109, opened 2016-02-20
Update VMs to the latest versions

ltratt:
Tested on Linux but not yet on OpenBSD.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,21 +32,21 @@ build-vms:
 build-benchs: build-krun
 	cd benchmarks && \
 		env LD_LIBRARY_PATH=${GCC_LIB_DIR} \
-		${MAKE} CC=${CC} JAVAC=${JAVAC}
+		CC=${CC} JAVAC=${JAVAC} ${MAKE}
 
 build-krun:
-	cd krun && env LD_LIBRARY_PATH=${GCC_LIB_DIR} ${MAKE} CC=${CC} \
+	cd krun && env LD_LIBRARY_PATH=${GCC_LIB_DIR} CC=${CC} \
 		JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_INC}"' \
 		JAVA_LDFLAGS=-L${JAVA_HOME}/lib \
-		JAVAC=${JAVAC} ENABLE_JAVA=1
+		JAVAC=${JAVAC} ENABLE_JAVA=1 ${MAKE}
 
 build-startup: build-krun
 	cd startup_runners && \
-		env LD_LIBRARY_PATH=${GCC_LIB_DIR} ${MAKE} CC=${CC} \
-		${MAKE} JAVA_CPPFLAGS='"-I${JAVA_HOME}/include \
+		env LD_LIBRARY_PATH=${GCC_LIB_DIR} CC=${CC} \
+		JAVA_CPPFLAGS='"-I${JAVA_HOME}/include \
 		-I${JAVA_HOME}/include/linux"' \
 		JAVA_LDFLAGS=-L${JAVA_HOME}/lib \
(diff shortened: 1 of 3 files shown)


ltratt:
OK, now tested to build successfully fully on Linux and OpenBSD (with PRs #108 and #107 merged in, for testing purposes).

Makefile
```diff
@@ -32,21 +32,21 @@ build-vms:
 build-benchs: build-krun
 	cd benchmarks && \
 		env LD_LIBRARY_PATH=${GCC_LIB_DIR} \
-		${MAKE} CC=${CC} JAVAC=${JAVAC}
```
vext01, line 35: Did you mean to move make variables to environment variables?

build.sh
```diff
@@ -318,9 +318,9 @@ build_jdk() {
 }
 
 MX_REPO=https://bitbucket.org/allr/mx
-MX_VERSION=d94328febb37
+MX_VERSION=072da2b
 GRAAL_REPO=http://hg.openjdk.java.net/graal/graal-compiler
-GRAAL_VERSION=9dafd1dc5ff9
+GRAAL_VERSION=graal-0.13
```
vext01, line 321: Yes, this! Versions are good.

vext01:
Apart from the variable query this looks OK.

Makefile
ltratt, line 35: Why are they `make` variables and not environment variables?

[The Makefile didn't even work, at least, on Linux before. I assume -- but don't know for sure -- that if you specify make variables you need to specify a target. And at least one of the commands specified `${MAKE}` twice, which was obviously wrong, on top of that.]

vext01:
So this doesn't have the gmake changes in. We need to merge that first.

vext01:
So you need a rebase

ltratt:
What about the variable query? Are you OK with the new version or ...?

vext01:
I'm not sure why we moved some variables from make to the environment.

ltratt:
I'm not sure why they were variables. What's the difference / point?

ltratt:
[As I said, the Makefile was broken in _several_ ways, and simply didn't run at all on Linux.]

ltratt:
[However, I may have identified the wrong fix!]

vext01:
`VAR=x make` is not the same as `make VAR=x`. The former sets the environment the latter sets make variables.

ltratt:
Yes, I get that bit ;) But does this make any difference in our case?

vext01:
I think it works because we don't define any of the variables that were moved in the Makefiles we are invoking. If later someone added `CC=cc` to (e.g.) the krun makefile, then our override would be ignored. I would rather make the override king.

Example of issue:

```
$ cat Makefile 
VAR=default

all:
        @echo "VAR=${VAR}"

$ make
VAR=default
$ VAR=override make
VAR=default  <------ environment var ignored
$ make VAR=override
VAR=override
```

ltratt:
I guess my counter-argument is that it's _really_ unusual to specific make variables (I've done so precisely zero so far in this lifetime), and that if people do things like set `CC` manually in the Makefile, they either know what they're doing or they get what they deserve...

Anyway, I'm not sure I feel strongly either way. If you want to add another commit moving things back, I'm fine, _but_ it will need testing, as the previous Makefile was broken (and I'm not sure I want to tease out which of my fixes is the minimum necessary).

vext01:
Right let's merge this, then I will apply the change separately. I know I'm sounding anal, but it would be a pain if we ran a 2 month experiment to find out that parts were built using the system C compiler accidentally.

Do you want to splat anything?

ltratt:
OK, rebased against master. Ready for merge.

vext01:
I take your point about hard-coding variables. Ideally people should not use `VAR=default` but instead `VAR?=default`.

PR #110, opened 2016-02-22
Since we only want 64 bit binaries, there's no point in supporting 32 bit arches

ltratt:
This solves the "I suspect your system does not have 32-bit developement
libraries" message on some Linux boxes.

Warning: currently under testing. Do not merge yet.


ltratt:
OK, this builds on bencher[356], so is ready for merge IMHO.

PR #112, opened 2016-02-24
Rename a target to be more consistent with clean target.

vext01:
We had a target `build-benchs` but the corresponding clean target `clean-benchmarks`.

This just renames `build-benchs` to `build-benchmarks` for consistency.

OK?


ltratt:
Sounds sensible to me.

PR #113, opened 2016-02-24
Print u_int64_t consistently across platforms.

vext01:
One of the C benchmarks (fasta) emits a warning when printing a unsigned 64. If I (naively) fix this on one platform (e.g. Linux), the other platform (BSD) will start emitting the warning. Seems there's a discrepancy in integer types between the platforms.

This is the correct way to do this.

OK?


ltratt:
I just learnt something new about integer types!

PR #114, opened 2016-02-25
Update OpenJDK version used.

ltratt:
Build tested on bencher[56].

This also fixes a problem whereby OpenBSD-only GCC patches were applied on Linux (this might have caused the HHVM problems on bencher3?).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -249,8 +249,7 @@ build_gmake() {
     cp make gmake
 }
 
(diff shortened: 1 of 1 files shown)


vext01:
Looks OK to me. Did you check it compiles and runs Java code OK?

ltratt:
It compiles some stuff during the build. Whether it runs our entire benchmark suite is another thing. But we have to do a "test run" soon, so I'd rather merge and then do the test run after.

PR #115, opened 2016-02-26
Don't build tracing versions of Java benchmarks.

vext01:
Let's not build the tracing versions of benchmarks. We don't use them in the experiment proper.

Currently only the Java tracing versions are built. This PR disables building them.

OK?


ltratt:
OK, I guess this makes sense.

PR #116, opened 2016-03-03
Update config file.

vext01:
Tidy up and I'd like to include our pre/post execution commands in our in-tree config. Would anyone object?

CC @snim2

### DIFF
--- a/warmup.krun
+++ b/warmup.krun
@@ -1,5 +1,6 @@
 import os
 import sys
+import platform
 from krun.vm_defs import (PythonVMDef, LuaVMDef, JavaVMDef, GraalVMDef,
 		          PHPVMDef, JRubyTruffleVMDef, V8VMDef, NativeCodeVMDef,
 			  find_internal_jvmci_java_bin, PyPyVMDef)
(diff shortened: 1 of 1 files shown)


snim2:
lgtm

warmup.krun
```diff
@@ -126,16 +126,49 @@ SKIP=[
...
+        "sudo /etc/rc.d/smtpd start",
+    ]
+else:
+    assert False
+
+
+# Copy off results after each execution -- soft-dev specific!
+POST_EXECUTION_CMDS.append(
+    "tar czf - ${KRUN_RESULTS_FILE} ${KRUN_LOG_FILE} | ssh -i id_rsa "
+    "vext01@bencher2.soft-dev.org 'cat > research/krun_results/bencher3.tgz'")
```
ltratt: Why don't we make the `bencher3` bit be the output from the `hostname` command? That way we can use the same config file for all machines (I think...).

warmup.krun
```diff
@@ -126,16 +126,49 @@ SKIP=[
...
+
+assert len(PRE_EXECUTION_CMDS) == 2
```
ltratt: I'm not sure I see the point of this assertion (or the one following). Remove?

vext01: I want to be _really_ sure we have the commands in place, as if they are not, we wasted time running as our results will be invalid.

warmup.krun
vext01: Oops, I was supposed to have done that! Good spot.

vext01:
Try this.

ltratt:
Works for me! Squash and then I'll merge?

vext01:
splat.

PR #117, opened 2016-03-15
Pre/post command tweaks

vext01:
I just noticed that Linux has another daemon that could run big tasks:  atd. In OpenBSD this is part of cron. One change ensures this is off -- just in case.

Also allow scp to fail as discussed on email.

OK?

CC @cfbolz @snim2 

Related note:

Just found rouge instances of gkrelmd on benchers. For some reason the line for gkrellmd (`service --status-all`) came out of stderr (not stdout) so I missed it when I was inspecting  the running services. I've removed gkrellmd on both benchers.

Looking at top, gkrellmd sometimes zips to the top of the process list, so there's a chance it has had a negative impact. I think we need to re-start the experiment.

### DIFF
--- a/warmup.krun
+++ b/warmup.krun
@@ -145,12 +145,14 @@ N_EXECUTIONS = 10  # Number of fresh processes.
 if sys.platform.startswith("linux"):
     PRE_EXECUTION_CMDS = [
         "sudo service cron stop",
+        "sudo service atd stop",
         "sudo service postfix stop",
     ]
 
     POST_EXECUTION_CMDS = [
-        "sudo service cron start",
-        "sudo service postfix start",
+        "sudo service cron start || true",
+        "sudo service atd start || true",
+        "sudo service postfix start || true",
     ]
 elif sys.platform.startswith("openbsd"):
     PRE_EXECUTION_CMDS = [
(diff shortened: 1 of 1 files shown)


ltratt:
I wonder if we should also make the `rc.d start` (not the stop!) commands `||true` too? After all, we don't _deeply_ care if they start up successfully or not do we? I don't feel strongly about this, though, and there is an argument that says "if they don't start up successfully, something's really gone wrong".

vext01:
So I think the plan here is to do something like:

```
scp blah blah || { sleep 3 ; scp blah blah } || true
```

Am i right?

As for the other commands, I am torn. On one hand, if daemons fail to start, most of the time we are about to reboot anyway. On the other hand, after the final execution, it would be nice to be sure that everything came up ok.

Thoughts?

ltratt:
Agreed on the `scp` command.

I don't really have a strong opinion either way on the daemon restarting. I'd _probably_ be inclined to do the `|| true` thing on them, simply because I probably care more about the experiments finishing than the machine being in a perfect state after the experiments have finished. But I don't feel super strongly about it.

vext01:
This seems to work. I had problems grouping with `{}` (syntax error), but found grouping with `()` to work. A bit of a mystery...

OK to squash?

ltratt:
I just tried on dash, and spotted the same issue. `()` works fine for me, so let's squash 'n' merge.

vext01:
Ah, its a dash thing.

ltratt:
[I also checked it works on zsh, which means `()` seems the more portable option, full stop.]

vext01:
OK squashed

warmup.krun
```diff
@@ -159,16 +161,21 @@ elif sys.platform.startswith("openbsd"):
...
 else:
     assert False
 
 
 # Copy off results after each execution -- soft-dev specific!
+#
+# We allow failure, otherwise Krun will halt the experiment if (e.g.) the
+# SSH server hostname is temporarily unavailable.
 HOSTNAME = platform.node().split(".")[0]
+SCP_CMD = ("tar czf - ${KRUN_RESULTS_FILE} ${KRUN_LOG_FILE} | "
```
ltratt, line 176: Just noticed... does this need to be in brackets?

vext01, line 176: just means we don't need backslashes for line continuations.

ltratt:
Ah, OK. Makes sense.

PR #118, opened 2016-03-30
Fix a crash.

vext01:
Triggered when one of the result files is a machine which runs a subset
of the experiments run in another results file.

I think this is the correct fix. OK?


snim2:
LGTM.

PR #119, opened 2016-04-14
Stop pflogd.

vext01:
`pflogd` is a daemon that runs by default on OpenBSD. Its purpose is to log firewall packets. I don't think this should run during benchmarking. 

OK?

### DIFF
--- a/warmup.krun
+++ b/warmup.krun
@@ -158,11 +158,13 @@ elif sys.platform.startswith("openbsd"):
     PRE_EXECUTION_CMDS = [
         "sudo /etc/rc.d/cron stop",
         "sudo /etc/rc.d/smtpd stop",
+        "sudo /etc/rc.d/pflogd stop",
     ]
 
     POST_EXECUTION_CMDS = [
         "sudo /etc/rc.d/cron start || true",
         "sudo /etc/rc.d/smtpd start || true",
+        "sudo /etc/rc.d/pflogd start || true",
     ]
 else:
     assert False


snim2:
Looks OK to me. Is there a chance that any of those commands can fail? Are all the daemons installed by default?

vext01:
All of those daemons are installed by default and a double stop is not an error, so we should be good.

vext01:
Ill just sort the conflict

vext01:
Sorry it should look like this. Starting is allowed to fail with `| true`

snim2:
OK, and you have tested this?

vext01:
i've run it on an OpenBSD host (without reboots) yes:

```
...
[2016-04-14 17:23:43 DEBUG] start temperatures: {'hw.sensors.cpu0.temp0': 34.0, 'hw.sensors.acpitz1.temp0': 29.8, 'hw.sensors.acpitz0.temp
0': 27.8}
[2016-04-14 17:23:43 DEBUG] temp reading: {'hw.sensors.cpu0.temp0': 34.0, 'hw.sensors.acpitz1.temp0': 29.8, 'hw.sensors.acpitz0.temp0': 27
.8}
[2016-04-14 17:23:43 DEBUG] execute shell cmd: sudo /etc/rc.d/cron stop
[2016-04-14 17:23:43 DEBUG] execute shell cmd: sudo /etc/rc.d/smtpd stop
[2016-04-14 17:23:43 DEBUG] execute shell cmd: sudo /etc/rc.d/pflogd stop
[2016-04-14 17:23:43 DEBUG] Writing out wrapper script to /tmp/krun_wrapper.dash
```

PR #121, opened 2016-06-13
Tweak inset axis font size and positioning.

vext01:
Just realised that the font size on the inset axis is unreadable. This fixes this and bumps the inset up a bit so the now larger x axis labels don't overlap one of the graphs.

OK?


snim2:
LGTM.

PR #126, opened 2016-06-20
Update JRuby and Truffle. Tidy use our JDK to build. Tidying.

vext01:
We can start reviewing this. Just doing a build on OpenBSD in order to know the JDK path (this will fix the XXX).
- Update JRuby
- For JRuby, use a maven buildpack so as to not fetch moving targets.
- Use the recommended graal-0.12 from a new graal repo.
- Use our JVM to build graal and jruby
- Various tidying.

Note: JRuby patch disappears as we can now access the monotonic clock from stock JRuby. Krun patch will follow.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -62,11 +62,10 @@ echo "===> Working in $wrkdir"
 PATCH_DIR=`pwd`/patches/
 
 # XXX when we stabilise, fix the krun revision.
-KRUN_REPO=https://github.com/softdevteam/krun.git
 fetch_krun() {
     echo "\n===> Download and build krun\n"
     if ! [ -d "${HERE}/krun" ]; then
-        cd ${HERE} && git clone ${KRUN_REPO} || exit $?
+        cd ${HERE} && git clone https://github.com/softdevteam/krun.git || exit $?
     fi
 }
 
@@ -199,14 +198,13 @@ build_pypy() {
 }
 
 V8_V=5.1.281.65
-DEPOT_REPO="https://chromium.googlesource.com/chromium/tools/depot_tools.git"
 build_v8() {
     cd ${wrkdir} || exit $?
     echo "\n===> Download and build V8\n"
 
     if [ -f ${wrkdir}/v8/out/native/d8 ]; then return; fi
 
-    git clone ${DEPOT_REPO} || exit $?
+    git clone "https://chromium.googlesource.com/chromium/tools/depot_tools.git" || exit $?
     cd depot_tools || exit $?
 
     # The build actually requires that you clone using this git wrapper tool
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -321,37 +323,29 @@ build_jdk() {
...
-# Instead we use the system JDK8.
-case `uname` in
-    Linux)   SYSTEM_JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64;;
-    OpenBSD) SYSTEM_JAVA_HOME=/usr/local/jdk-1.8.0;;
-    *) unknown_platform;;
-esac
-MX="env DEFAULT_VM=jvmci JAVA_HOME=${SYSTEM_JAVA_HOME} python2.7 ${wrkdir}/mx/mx.py"
+# The latest Graal and MX at the time of writing.
+# MX doesn't have releases.
+MX_REPO=https://github.com/graalvm/mx
```
ltratt: Elsewhere in the file, we have the "static" things (like websites) in the body of the functions, and only the "changeable" things (version numbers) as variables.

build.sh
```diff
@@ -255,23 +255,25 @@ build_gmake() {
...
 
-JDK_DIR=openjdk-8u72b15-bsd-port-20160220
+# We use the JDK we just built for all consequent Java compilation (and as a
+# basis for the Graal compiler).
 case `uname` in
-    Linux)
-      JDK_JAVAC=${wrkdir}/openjdk/build/linux-x86_64-normal-server-release/jdk/bin/javac;;
-    OpenBSD)
-      JDK_JAVAC=${wrkdir}/openjdk/build/bsd-x86_64-normal-server-release/jdk/bin/javac;;
+    Linux)   OUR_JAVA_HOME=${wrkdir}/openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image/;;
```
ltratt, line 259: Can we find a nicer name than `OUR_`? I always find it really ambiguous. Maybe just call it `JAVAC`? [`JDK_` seems a bit redundant.]

build.sh
vext01: Searching around a bit, we seem to have a mix. Which style do you prefer, and I will update them all?

build.sh
vext01, line 259: `OUR` is indicating it's the one we built. Not the system one. I'm open to suggestions to another word, but I don't think we should remove the implication that this is the one we built. For the warmup experiment It's super important that we don't compile using any system compilers (other than for the initial gcc bootstrap).

vext01:
Fixed a couple of bits. Let me know about what to change `OUR` to and I'll fix that.

build.sh
snim2, line 259: Is there a particular reason not to use `JAVA_HOME`?

vext01, line 259: Just to be explicit. We did the same with GCC.

ltratt, line 259: OK, let's call it `OUR_JDK` then.

build.sh
ltratt: I don't mind hugely. Let's pick whichever is most common, and rewrite the less common ones into that idiom (but perhaps as a separate PR).

build.sh
vext01, line 259: Ah wait. I see some confusion. `JAVA_HOME` is a standard Java variable used to specify which JDK to use for various Java gubbins. The particular variable `OUR_JAVA_HOME` is a variable which I am going to pass to a part of the build system which expects me to set a `JAVA_HOME`. Hence the naming I used. E.g.:

```
MX="env JAVA_HOME=${OUR_JAVA_HOME} python2.7 ${wrkdir}/mx/mx.py"
```

vext01, line 259: Maybe I should bite the bullet and just call it `JAVA_HOME`, although I'm a little worried it might interfere with system packaged Java apps that we are using, like ant. Thoughts?

ltratt, line 259: `JAVA_HOME` is overloaded (and not, I think, correct, since we're pointing at javac, not the top-level JDK/JRE dir?).

vext01, line 259: I don't think so. I replaced the variable that points to a javac with one which points to a JAVA_HOME, since part of the build wants me to specify a JAVA_HOME and javac is easily expressed in terms of a JAVA_HOME.

ltratt, line 259: What I mean is: we might accidentally corrupt the environment was pass to other programs (since, I think, `JAVA_HOME` might be a semi-standard name for Java thingies that they might check for).

vext01, line 259: Do you mean we shouldn't shotgun blast it with `JAVA_HOME`?

FWIW: A JAVA_HOME should should traditionally point to a jdk or jre "image", but you are right that software in the wild may have a different idea.

ltratt, line 259: Not that we should blast it, but that we shouldn't fiddle with it either way IMHO. Let's just use another variable name. `OUR_JAVAC` is fine with me.

vext01, line 259: Eh? Now I am confused.

We are not pointing to a javac,

ltratt, line 259: Oops, yes, me too. Sorry! `OUR_JAVA_HOME` is OK. Doh.

vext01, line 259: Uses of `OUR_JAVA_HOME`:
- `if [ -f ${OUR_JAVA_HOME}/bin/javac ]; then return; fi` references a javac
- `MX="env JAVA_HOME=${OUR_JAVA_HOME} python2.7 ${wrkdir}/mx/mx.py"`references a java jdk  image
- `JAVACMD=${OUR_JAVA_HOME}/bin/java` references a java vm binary

All are in standard locations relative to a JAVA_HOME. So if anything, we want `OUR_JAVA_HOME`, no?

ltratt, line 259: Agreed.

vext01, line 259: Ah phew.Thanks

vext01:
So, OK to squash at this point?

ltratt:
Yep, squash away!

vext01:
ready

PR #129, opened 2016-07-01
Use Maven3.

vext01:
This makes our build use maven3, which seems to fix our offline build issues. I also check there is no maven cache files lying around which could change the way the build happens.

Also improve some code locality.
## Tabs

Looks like there are some naughty tabs in the build script. Following this PR I will do a tab sweep.


snim2:
This looks fine to me, as-is. I'm happy to review the tab sweep, also.

Issue #125, opened 2016-06-17
Startup experiment does not build.


vext01:
```
cd startup_runners && \
        env LD_LIBRARY_PATH=/home/ltratt/warmup_experiment/work/gcc-inst/lib CC=/home/ltratt/warmup_experiment/work/gcc-inst/bin/zgcc \
        JAVA_CPPFLAGS='"-I/home/ltratt/warmup_experiment/work/openjdk/build/bsd-x86_64-normal-server-release/images/j2sdk-image/include \
        -I/home/ltratt/warmup_experiment/work/openjdk/build/bsd-x86_64-normal-server-release/images/j2sdk-image/include/linux"' \
        JAVA_LDFLAGS=-L/home/ltratt/warmup_experiment/work/openjdk/build/bsd-x86_64-normal-server-release/images/j2sdk-image/lib \
        JAVAC=/home/ltratt/warmup_experiment/work/openjdk/build/bsd-x86_64-normal-server-release/jdk/bin/javac ENABLE_JAVA=1 gmake
gmake[1]: Entering directory '/home/ltratt/warmup_experiment/startup_runners'
/home/ltratt/warmup_experiment/work/gcc-inst/bin/zgcc -fPIC -Wall -Wextra -pedantic  -L/home/ltratt/warmup_experiment/startup_runners/../krun/libkruntime \
         outer_startup_runner.c \
        -o outer_startup_runner_c -lkruntime  \

/usr/bin/ld: cannot find -lkruntime
```

vext01:
Fixed by @cfbolz

PR #131, opened 2016-07-02
Disable building the startup experiment for now.

vext01:
It currently does not build. See #125.


ltratt:
So long as we don't forget to reenable it, I'm fine with this.

PR #132, opened 2016-07-14
Reenable a no-longer-crashing benchmark (and tidy another)

vext01:
This re-enables a benchmark (spectralnorm:JRubyTruffle) which no longer crashes, and also performs OK now (about a second per in-process iteration).

There were two other previously misbehaving benchmarks I investigated too:
- fasta:JRubyTruffle no longer crashes but has a checksum error only under Jrubytruffle. Reported to Chris. Details: https://gist.github.com/vext01/4f5547c6199b2542ba8f5fd9e7b6abf0
- richards:HHVM is still amazingly slow. About 35 seconds per in-process iteration. We could enable this and add 8 days to the experiment. What do you think?

Also tidied up some bogus and somewhat redundant comments.

### DIFF
--- a/benchmarks/fasta/ruby/bench.rb
+++ b/benchmarks/fasta/ruby/bench.rb
@@ -93,9 +93,8 @@ def make_random_fasta(table, n)
 end
 
 
-# work around ruby scoping using lambda
 def run_iter(n)
-    for i in 0..n-1  # inclusive upper bound
+    for i in 0..n-1
         make_repeat_fasta(@alu, @SCALE*2)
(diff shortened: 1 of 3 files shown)


ltratt:
Looks good to me.

vext01:
Did you want to re-enable hhvm/richards though?

ltratt:
Ah, oops, sorry, didn't realise you wanted that in this PR.

I think we should enable hhvm/richards (we can live with 8 days).

vext01:
OK, another PR coming.

PR #136, opened 2016-08-04
Fix startup

cfbolz:
This pull request syncs `startup.krun` to the changes that were done in krun. The core approach of how startup is measured stays the same. The pull request also makes `startup.krun` contain considerably less copy-pasted code, which means it should be more robust to future changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ endif
 
 CC=${PWD}/work/gcc-inst/bin/zgcc
 
-all: build-vms build-benchmarks build-krun
+all: build-vms build-benchmarks build-krun build-startup
 	@echo ""
 	@echo "============================================================"
 	@echo "Now run 'make bench-no-reboots' or 'make bench-with-reboots'"
--- a/startup.krun
+++ b/startup.krun
@@ -1,8 +1,5 @@
-import os
-from krun.vm_defs import (PythonVMDef, LuaVMDef, JavaVMDef, GraalVMDef,
-    PHPVMDef, JRubyTruffleVMDef, V8VMDef, NativeCodeVMDef,
-    find_internal_jvmci_java_bin, GenericScriptingVMDef)
-from krun import EntryPoint
+execfile("warmup.krun", globals()) # import the warmup definitions completely
+
 
 # this file is a minimally hacky way to do startup measurements with all the
 # krun goodies, but without changing krun.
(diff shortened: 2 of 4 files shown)


vext01:
```
# sudo -u krun nice -20 taskset 0x2 env <some variables> \
# outer_startup_runner_c jruby startup_runner.rb
```

This is now out of date. We don't pin, so I wouldn't expect `taskset`. I would just elide this part perhaps(?)

cfbolz:
Yeah, as this is just to illustrate the approach, just I just kill everything up to the call to `env`?

vext01:
Works for me.

vext01:
This looks good.

This probably doesn't matter, but why does `patch_platform` accept an instance. Why don't we just lookup the class directly?

vext01:
P.S. I'm going to test this now.

cfbolz:
We could patch all platform classes, I guess?

vext01:
I guess I expected to patch the `krun.platform` class directly in such a way that all instances are patched. If it's annoying, never mind though.

cfbolz:
But it _is_ the class that is patched. However, to find out which class to patch (Linux, BSD, whatever), I run platform detection first.

vext01:
Ah yes. Very good :)

cfbolz:
so, we wait for the test, then merge?

vext01:
Yeah, just testing it...

vext01:
I might be being dumb, but the results should be deltas of monotonic clock values (i.e. seconds), right? The results look too large:

```
 "data": {
  "binarytrees:C:default-c": [
   [
    1176.679015,
    1176.679245
   ],
   [
    1182.930025,
    1182.930286
   ]
  ]
...
```

cfbolz:
no, you need to substract the two numbers to get the seconds

cfbolz:
(as the comment explains)

vext01:
I get it. So each "execution" is really a start and stop time.

It was unfortunate that I chose a value of 2 for `N_EXECS`. Managed to confuse myself.

cfbolz:
ah :-)

vext01:
Add `build-startup` to the `all` target in the `Makefile` and we can merge.

cfbolz:
like this?

vext01:
You got it.

Want to squash?

cfbolz:
done

vext01:
Cheers!

cfbolz:
cool, thanks

PR #137, opened 2016-08-05
Make outliers more visible.

snim2:
This PR embiggens all the blobs that mark outliers, as in the attached example.

embiggen_the_blobs


ltratt:
I'll do this one (since it was my moaning that caused it!).

ltratt:
I like this.

snim2:
Thanks!

PR #140, opened 2016-08-08
Add 20pc margin to x-axis.

### LINKED ISSUE #138: Graph padding
In some graphs -- most commonly noticed if there's an outlier at position 0 -- it's hard to see what's going on. I wonder if we should take a leaf from Rebecca's book and add internal padding to the graphs?

snim2:
Add 20pc margin to x-axis. Fixes #138 

This proved to be a little tricky, because the matplotlib `axes.margins()` method does not work with Seaborn. However, this PR does work, and the code currently adds no margins to the y-axes and a 20% margin to the x-axis. In run sequences where the first dataum is an outlier, the outlier marker can now be seen.

Example:
twenty_percent_margin

### DIFF
--- a/plot_krun_results.py
+++ b/plot_krun_results.py
@@ -323,6 +323,20 @@ def draw_subplot(axis, data, title, x_range, y_range, window_size, outliers,
     return handles, labels
 
 
+def add_margin_to_axes(axis, x=0.01, y=0.01):
+    """Seaborn-friendly way to add margins to axes (default 1% margin).
+    """
+
+    if x > .0:
+        xlim = axis.get_xlim()
+        xmargin = (xlim[1] - xlim[0]) * x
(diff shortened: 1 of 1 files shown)


ltratt:
Should we also add the same padding to the y axis (I'm sure there'll be some graphs where we have the same problem we do with x axes)?

snim2:
The `y-limits` are always adjusted when we give all process executions the same `y-limits`, so there is always a bit of space at the top and bottom of the charts. Also, we can't have an outlier right next to the x-axis, so the x-axis can never obscure the markers, a marker at the top of the range will always have a bit of headroom because the `y-limits` are generous.

This was only tricky because we don't want to start the `x-limit` at `-1`, we needed a margin between the zeroth data point and the y-axis.

Below is an example:

marker-at-top

ltratt:
Point taken!

PR #142, opened 2016-08-11
TSR support

vext01:
Let's start looking at TSR support. Please don't merge yet, as discussion is required, and a few small tweaks.

There is going to be an accompanying krun change, which I will link shortly.
EDIT: https://github.com/softdevteam/krun/pull/234

In short: We need to be able to get at the TSR register from all of the VMS.

Getting at the TSR means ultimately calling inline assembler. There is a C function in libkrun to get at the TSR. If convinient (as is for Python and Lua) we simply call to libkrun. For other VMs where it is not so easy to call libraries, I've built the C function into the VM itself.

Notes:
- The TSR is a 64-bit unsigned value, but the numeric representaions of some languages can't represent this.
  - Javascript and Lua only have doubles -- we _should_ be OK (?)
  - PHP has platform specific signed integers only. In HHVM I've attempted to work around this by only ever working on TSR values in C++ code. I'm not happy with it -- need to think more -- worst case, return a double.
  - Java has no unsigned 64, but `long` (64-bit signed) can be interpreted as 64-bit unsiogned  if we are careful.
- We should use the synchronising variant of the TSR instruction -- This will be the last change I make.

I've reformatted the graphs, checked `--one-page` and `--interactive` work, but due to the massive TSR values, the Y-axis is kinda annoying:

tsr

Note the weird axis offset/scale on the top right. (Also note the crossover of the wallclock and TSR time early on. This seems to be a recurring trend, interesting...). Not sure how we could present the graph better...

We all need to check this super-carefully.

Later I'm going to run a small experiment and upload a result file for your inspection.

Let the discussion begin.


cfbolz:
It's cycle count since the last reboot. A double has 52 bits of mantissa, so 2 *\* 52 can still be represented precisely in a double. The cycle count is around 3e9 per second. `2**52 / 3e9` is about 416 hours. Since we reboot regularly, we should be fine.

vext01:
That's a nice quote for the paper, Carl. Thanks.

patches/hhvm_exts.diff
```diff
@@ -46,32 +54,70 @@ index 6e94fa9..26a8a35 100644
...
++ * $OpenBSD: pctr.h,v 1.5 2014/03/29 18:09:28 guenther Exp$
++ *
++ * PHP has no way to represent a unsigned 64-bit integer. We return a signed
++ * 64. Any app-level code using this must know that this is really a unsigned
++ * value.
++ */
++int64_t HHVM_FUNCTION(read_ts_reg) {
++    u_int32_t hi, lo;
++
++    __asm volatile("rdtsc" : "=d" (hi), "=a" (lo));
```
ltratt: Should be rtdscp I think.

patches/hhvm_exts.diff
```diff
@@ -46,32 +54,70 @@ index 6e94fa9..26a8a35 100644
...
  Array HHVM_FUNCTION(sys_getloadavg);
  Array HHVM_FUNCTION(token_get_all, const String& source);
  String HHVM_FUNCTION(token_name, int64_t token);
 +double HHVM_FUNCTION(clock_gettime_monotonic);
++int64_t HHVM_FUNCTION(read_ts_reg);
++void HHVM_FUNCTION(print_ts_delta, const int64_t start, const int64_t stop);
  String HHVM_FUNCTION(hphp_to_string, const Variant& v);
- extern const double k_INF;
- extern const double k_NAN;
+ Variant HHVM_FUNCTION(SystemLib_max2, const Variant& arg1, const Variant& arg2);
```
ltratt: Minor thing, but shouldn't it be `const Variant &arg1` (i.e. the ampersand next to the variable name)?

plot_krun_results.py
```diff
@@ -255,9 +261,14 @@ def transform(coord):
     return fig.add_axes([fig_left, fig_bottom, fig_width, fig_height])
 
 
-def draw_subplot(axis, data, title, x_range, y_range, window_size, outliers,
+def draw_subplot(axis, data, tsr_data, title, x_range, y_range, window_size, outliers,
                  unique, common, mean, sigma):
+
+    # We present the TSR data as a ration of wall-clock time
```
ltratt: "Ratio"

patches/hhvm_exts.diff
vext01: Just copying what I see around this area, e.g.

```
Variant HHVM_FUNCTION(unpack, const String& format, const String& data);
```

ltratt: OK, let's stick with the local culture.

plot_krun_results.py
vext01: Oops, did we want to plot the ratio?

ltratt: I don't know. But I'm pretty certain we didn't want it to be a ration of streaky bacon ;)

vext01: I think we want to plot the raw TSR times for now.

vext01: Mystery solved:

noratio

vext01:
Looks like we are confused as to the meaning of the timestamp register:

```
The processor monotonically increments the time-stamp counter MSR every
clock cycle and resets it to 0 whenever the processor is reset.
```

It's nothing to do with the number of instructions executed. It's actually simply a measure of time.

The P variant waits for the pipeline to empty. I'm not sure we want that, since the pipeline includes future instructions. Am I right?

It's also good to know the TSR is zeroed on reboot.

ltratt:
Yes, it's number of cycles and yes you can think of it as a measure of time, though since most humans perceive time in seconds, I'm reluctant to confuse people by phrasing it in that way.

Neither `rtdsc` or `rtdscp` is perfect, and in our cases, it _probably_ doesn't make much difference... but it's arguably better to flush the pipeline than not.

vext01:
I agree that there is unlikely to be much difference, but wouldn't using the contents of the pipeline include (a tiny amount of) work after the iteration?

ltratt:
Have a look at Section 3.2.1 (p16 of the PDF) http://www.intel.com/content/dam/www/public/us/en/documents/white-papers/ia-32-ia-64-benchmark-code-execution-paper.pdf -- we should probably be doing something a little more sophisticated again.

vext01:
I see.

Well it wouldn't be a total PITA to implement that. Can I recommend we make that enhancement as another PR, since this PR (and the Krun one) is getting large at this point. WHat do you think?

vext01:
This one is good to squash too?

ltratt:
Yep.

vext01:
Picked this one apart. Ready.

PR #146, opened 2016-08-16
patch pypy to fix log nesting

cfbolz:
equivalent to pypy revision 6a6545b4a915


vext01:
Looks good to me. Did you test it with Krun?

cfbolz:
nope, but tonight's pypy tests ran through

vext01:
Well, I guess any problems will be caught by our initial short runs, so OK.

PR #147, opened 2016-08-17
New TSR interfaces for VMs.

vext01:
In response to: https://github.com/softdevteam/krun/pull/238

Note that the new TSR functions now all use libkruntime instead of duplicating code in the VM itself. The support code was getting large and I was worried that we might end up in a situation where the code falls out of sync between VMs and libkruntime.

There is still duplication WRT `clock_gettime_monotonic`, but I will raise a separate PR to fix that next (make them use libkrun too).

Note the ruby support is special. It now always returns a Ruby `Bignum`.

OK?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -479,7 +479,11 @@ build_hhvm() {
     ln -sf ${OUR_CXX} `dirname ${OUR_CXX}`/g++ || exit $?
     HHVM_PATH=`dirname ${OUR_CC}`:${PATH}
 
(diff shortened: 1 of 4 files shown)


vext01:
And how about this one? We should merge the two PRs at the same time.

cfbolz:
looks good to me. do you want to splat?

vext01:
I reckon it is OK as-is.

Let Laurie OK before merge.

vext01:
Or not :P

ltratt:
The Truffle part looks a bit heavyweight, calling `getcontext` and so on -- is that fair? I wonder if we can cache the libkruntime somewhere, and then just call the TSR function?

vext01:
Raise a bug and I will look tomorrow.

ltratt:
It's not a bug necessarily...

vext01:
Might be able to cache it locally in the constructors.

ltratt:
Yes, that's the sort of thing I was thinking of.

Issue #150, opened 2016-08-19
Remove duplication from plotting scripts


snim2:
Once more urgent tasks have been completed, duplication between the plotting scripts needs to be removed. The important functions that calculate outliers, 5-sigmas and so on are all duplicated. This makes it increasingly likely that bugs will creep in as we make changes.

PR #152, opened 2016-09-06
New outlier calculations

### LINKED ISSUE #139: Change the way we calculate outliers
Outliers are currently defined as data points that lie outside a sliding 5-sigma. For a window size of `W`, the first `W-1` data do not have a full window of data with which to calculate the rolling mean and 5-sigma.

Currently, we fill partially empty windows with zeroes. This means that any benchmark that warms-up in  the traditional way will have its pre-warmup data points marked as outliers. This causes a problem for changepoint analysis, which ignores _all_ outliers.

To resolve this we need to find a more appropriate way to 'fill' the first window of data.

### LINKED ISSUE #149: Outlier-per-threshold script
This script (see #141 ) needs some visual improvement. In particular, the markers are not all visible.

snim2:
**DO NOT MERGE**

This is ready for review, but cannot be merged because the code in `master` cannot read the 0.6 version of the data.

This PR replaces the old outlier calculations (mean and 5 sigma) with this formula from Tukey:

> +/- 3 (90^th percentile - 10^th percentile)

Command-line options for relevant scripts have been changed to match (e.g. we now pass `--median` rather than `--mean` to the plot_krun script).

Some visual improvements to the outliers-by-threshold chart are also included. Run sequences now look like this:

current_per_threshold

Needs squashing.

Fixes #149 
Fixes #139

### DIFF
--- a/mark_outliers_in_json.py
+++ b/mark_outliers_in_json.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python2.7
 """
 Determine which iterations in Krun data are outliers, where an outlier is
-greater than 5 sigma above from a rolling mean.
+greater than 6 sigmas above / below a rolling mean.
 
 usage: Write lists of outliers into Krun results file(s).
 Example usage:
@@ -76,16 +76,24 @@ def get_outliers(all_outliers, window_size, threshold=1):
     return common, unique
 
 
-def get_all_outliers(data, window_size):
+def _tuckey_all_outliers(data, window_size):
+    # Ignore windows that do not have a full set of data.
     all_outliers = list()
     for index, datum in enumerate(data):
         l_slice, r_slice = _clamp_window_size(index, len(data), window_size)
+        if l_slice == 0 and r_slice < window_size:
+            continue
         window = data[l_slice:r_slice]
-        mean = numpy.mean(window)
-        five_sigma = 5 * numpy.std(window)
-        if datum > (mean + five_sigma) or datum < (mean - five_sigma):
+        median = numpy.median(window)
+        pc_band = (3 * (numpy.percentile(window, 90.0, interpolation='nearest')
+                        - numpy.percentile(window, 10.0, interpolation='nearest')))
+        if datum > (median + pc_band) or datum < (median - pc_band):
             all_outliers.append(index)
     return all_outliers
+
+
+def get_all_outliers(data, window_size):
+    return _tuckey_all_outliers(data, window_size)
 
 
 def _clamp_window_size(index, data_size, window_size=200):
--- a/plot_krun_results.py
+++ b/plot_krun_results.py
@@ -105,7 +105,7 @@ DPI = 300
 
 
 def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
-         xlimits, with_outliers, unique_outliers, mean=False, sigma=False,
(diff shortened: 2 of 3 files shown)


ltratt:
Is this ready for review now?

snim2:
It is ready for review, but it can't be merged because master can't read the 0.6 data. I've raised this as a PR now because an subsequent changes to the scripts will need to follow this, and I didn't want to bundle everything into one very confusing PR.

mark_outliers_in_json.py
```diff
@@ -76,18 +76,26 @@ def get_outliers(all_outliers, window_size, threshold=1):
     return common, unique
 
 
-def get_all_outliers(data, window_size):
+def _tuckey_all_outliers(data, window_size):
+    # Ignore windows that do not have a full set of data.
     all_outliers = list()
     for index, datum in enumerate(data):
         l_slice, r_slice = _clamp_window_size(index, len(data), window_size)
+        if (r_slice - l_slice) < window_size and l_slice == 0:
```
ltratt: Isn't this the same as `if l_slice == 0 and r_slice < window_size`?

plot_krun_results.py
```diff
@@ -261,21 +255,32 @@ def transform(coord):
     return fig.add_axes([fig_left, fig_bottom, fig_width, fig_height])
 
 
-def draw_subplot(axis, data, tsr_data, title, x_range, y_range, window_size, outliers,
-                 unique, common, mean, sigma):
+def _no_first_window_get_window(index, window_size, data):
+    # Ignore windows that do not have a full set of data.
+    l_slice, r_slice = _clamp_window_size(index, len(data), window_size)
+    if (r_slice - l_slice) < window_size and l_slice == 0:
```
ltratt: Same comment as above. [Maybe refactor into a common method?]

mark_outliers_in_json.py
snim2: I don't think so. `r_slice` and `l_slice` are array indices, so we want to know whether they are a window size apart. So, this is saying "have we failed to fill a window on the left hand side of the chart"

ltratt: If `l_slice == 0` then `(r_slice - l_slice)` must also be 0. That's why I suggested rewriting it to do the `l_slice` check first, to make it a bit clearer.

snim2: For example, if the window size is 100 and we are looking at the 5th data point, `l_slice` will be 0 and `r_slice` will be 55, so `r_slice - l_slice < window_size` is true, `l_slice == 0` is true, and this partial window should be ignored.

I'll swap the tests round.

plot_krun_results.py
snim2: Issue #150 covers refactoring

mark_outliers_in_json.py
ltratt: Doh, I made a typo. "If l_slice == 0 then (r_slice - l_slice) must also be 0" should be "If l_slice == 0 then (r_slice - l_slice) must be equal to `r_slice`".

plot_krun_results.py
ltratt: OK, although we can do things as we go along if that makes sense.

snim2:
Finished squashing...

ltratt:
Looks good. I can now merge?

snim2:
Yes please!

PR #154, opened 2016-09-07
Refactor scripts

### LINKED ISSUE #150: Remove duplication from plotting scripts

snim2:
This PR removes the problematic duplication we have in the scripts, and places common code in a package called `lib/`.

It also removes the `outlier_count` script which generated a latex document.

There was some discussion on a previous PR about creating a `bin/` directory and renaming the scripts, so this PR probably requires some discussion before we squash.

Fixes #150


ltratt:
I think I'd like to see the executable scripts stripped of their filename extensions and put in a `bin/` directory. I don't feel super-super strongly about this though.

ltratt:
[I should have said: the rest of the PR looks fine to me!]

cfbolz:
looks good, but `lib` is a super-generic name, and can lead to name clashes. How about naming the dir something specific?

ltratt:
I was imagining mirroring the standard unix `bin`/`lib` layout -- so indeed, it's generic ;) Then the Python executables in `bin` add the (absolute) `lib` path to their `sys.include`. I don't think this can clash with anything.

vext01:
I'm not too fussed by naming as long as it's obvious what is in which dir.

For scripts, I'm ok with (e.g.): scripts, tool, bin.
For the library dir, warmup, is probably as good a module name as any.

snim2:
OK, I can never remember how Python deals with relative paths, but this seems to work.
Anything else?

bin/mark_outliers_in_json
```diff
@@ -21,11 +21,13 @@
...
 
 import argparse
-import bz2
-import json
-import numpy
 import os
 import os.path
+import sys
+
+sys.path.insert(0, '../warmup')
```
cfbolz: I am a bit sceptical about this. I can never remember the details either, but I think it adds the path relative to the current working directory, so the script won't work if ran from somewhere else. you could turn this into an absolute path relative to the file, with code roughly like this:

`os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "warmup")`

vext01: Didn't Python have some relative import syntax which we could use (I never used it)?

ltratt: CF's mechanism is the most reliable.

snim2: Python does have relative import syntax, it would be `from ..warmup.module import name` but this won't work in our case because `..` is not a package (although `warmup` is).

cfbolz:
looks good to me now. squash, then I'll merge.

snim2:
Squashed, thanks!

cfbolz:
mergd, thanks!

PR #155, opened 2016-09-07
Fix xlimits and ticks.

snim2:
This PR fixes a bug that has crept into the outlier plotting code. 

This commit changes this:

old_xlim

to this:

new_xlim

### DIFF
--- a/bin/plot_outliers_by_threshold
+++ b/bin/plot_outliers_by_threshold
@@ -85,7 +85,7 @@ def plot_results(outliers_per_thresh, filename):
         window = WINDOWS[index]
         axis = axes[row, col]
         axis.ticklabel_format(useOffset=False)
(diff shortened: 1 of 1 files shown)


ltratt:
Not that I'm saying this PR is the right place to fix it, but I don't quite understand the y-axis scale. It doesn't start at 0 (which is fine), but the interesting data is entirely in the bottom half of the graph (so finishing at 5594 seems a bit weird).

snim2:
This is because the chart above is a small part of a larger file, and we give all the y-axes the same y-limits. This chart from the same PDF:

win400

is the one that causes the large range in the y-limits.

ltratt:
Ah, OK. Then I think we're good here.

snim2:
Thanks

PR #156, opened 2016-09-09
Support for core-cycle counts.

vext01:
These changes are paired with the core-cycles-msr PR on krun.

(It also includes the V8 fix @cfbolz encountered)


vext01:
This is now synced with @snim2's changes on master.

build.sh
```diff
@@ -62,11 +62,23 @@ echo "===> Working in $wrkdir"
 PATCH_DIR=`pwd`/patches/
 
 # XXX when we stabilise, fix the krun revision.
-fetch_krun() {
+build_krun() {
```
ltratt: Should this be something like `build_initial_krun`?

patches/jruby.diff
```diff
@@ -51,66 +51,73 @@ index 50a155d..0234ef7 100644
...
+@@ -1985,4 +1986,56 @@ public abstract class KernelNodes {
  
      }
  
-+    @CoreMethod(names = "read_ts_reg_start", needsSelf=false)
-+    public abstract static class ReadTSRegStartNode extends CoreMethodArrayArgumentsNode {
++    @CoreMethod(names = "read_core_cycles", needsSelf=false)
++    public abstract static class ReadCoreCyclesNode extends CoreMethodArrayArgumentsNode {
 +
 +        private static final BigInteger negAdjust = BigInteger.ONE.shiftLeft(64); // 2^64
```
ltratt, line 61: Did the indentation level just jump to 8?

ltratt:
Looks pretty good to me (apart possibly from wonky indentation in the Java part?).

patches/jruby.diff
vext01: Note that this is the first line of a new block. Looks OK to me, unless I am mistaken.

vext01:
Addressed those bits.

patches/jruby.diff
ltratt: Nope, you're right (oops!).

ltratt:
OK, thanks. Let's wait for @cfbolz to do a pass.

vext01:
and can we merge this and the krun change at the same time too?

ltratt:
Yes, although I don't think we need to be super-worried about pressing the merge buttons completely simultaneously :)

vext01:
I think these two PRs are good to merge.

ltratt:
Still waiting for @cfbolz's OK on this PR.

cfbolz:
Sorry for the delay, this looks good to me.

vext01:
OK to squash?

ltratt:
Yep, let's squash then I can merge.

vext01:
splat

ltratt:
Can we fix the typo in the "fix import path" commit? [Assuming it needs to be a separate commit.]

vext01:
splatted into one

PR #157, opened 2016-09-15
add an "empty" benchmark

cfbolz:
It is used to get an estimate of the granularity and the overheads of the timers.

This was ready since a while, but I wanted to test it with the msr work. I just did, it simply works.


vext01:
Looks OK. I wonder why you bothered adding tracing variants though :P

cfbolz:
for completeness ;-)

vext01:
Fair enough. Merging.

PR #158, opened 2016-09-15
Fix startup after msr

cfbolz:
The msr work broke startup, this pull request fixes it. In addition, I fixed the problem with running a rebooting version of the startup benchs.

### DIFF
--- a/startup.krun
+++ b/startup.krun
@@ -60,7 +60,10 @@ def patch_platform(platform):
         if "platform_sanity_check" in args[-3]:
             return prepend_args + args
 	return prepend_args + ["startup_runners/outer_startup_runner_c"] + args
-    assert Cls.__dict__.get('bench_cmdline_adjust') is None
+    old_bench_cmdline_adjust = Cls.__dict__.get('bench_cmdline_adjust')
+    assert (old_bench_cmdline_adjust is None or
+                old_bench_cmdline_adjust._startup_patch)
+    bench_cmdline_adjust._startup_patch = True
     Cls.bench_cmdline_adjust = bench_cmdline_adjust
 
 patch_platform(detect_platform(None, None))
--- a/startup_runners/startup_runner.c
+++ b/startup_runners/startup_runner.c
@@ -26,6 +26,6 @@ main(void)
     double    start_time = -1;
 
     start_time = clock_gettime_monotonic();
-    fprintf(stdout, "%f]\n", start_time);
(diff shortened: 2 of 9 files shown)


startup_runners/startup_runner.c
```diff
@@ -26,6 +26,6 @@ main(void)
     double    start_time = -1;
 
     start_time = clock_gettime_monotonic();
-    fprintf(stdout, "%f]\n", start_time);
+    fprintf(stdout, "%f], [-1.0, -1.0]]\n", start_time);
```
vext01, line 29: -1.0 is not a valid MSR value. Thoughts? Should we be using 0 instead? Or it doesn't matter?

startup.krun
```diff
@@ -60,7 +60,10 @@ def patch_platform(platform):
         if "platform_sanity_check" in args[-3]:
             return prepend_args + args
 	return prepend_args + ["startup_runners/outer_startup_runner_c"] + args
-    assert Cls.__dict__.get('bench_cmdline_adjust') is None
+    old_bench_cmdline_adjust = Cls.__dict__.get('bench_cmdline_adjust')
```
vext01, line 63: So why is bench_cmdline_adjust non-None the second time around? The patching somehow persists between runs?

vext01 (commented):
a comment

startup_runners/startup_runner.c
cfbolz: well, I am intentionally not using a valid MSR value, to prevent any other code from thinking it is one.

startup.krun
cfbolz: No, when resuming the code config is loaded twice: once, from the .krun when starting krun, and then a second time via `Results.read_from_file`. If that should be changed, that's a different discussion.

startup_runners/startup_runner.c
vext01: Should be fine.

startup.krun
vext01: Oh weird. Can you raise a PR? Let's roll with this workaround for now.

vext01 (approved):

vext01:
OK as is? Or squash?

cfbolz:
I'd like to leave them separate, they do very different things.

vext01:
Yep, I tend to agree, but polite to ask anyway :) #howbritish

PR #162, opened 2016-09-20
Plot aperf/mperf data.

### LINKED ISSUE #159: Deal with APERF/MPERF in plotting
Reported by @vext01 

We need a way to visualise the APERF/MPERF data in our plotting script.

Some background:
- MPERF is a continually incrementing counter.
- APERF is a counter incrementing only if the CPU is _on_ (in C0 C-state)

From the Intel manual:

```
By default, the IA32_MPERF counter counts during forced idle periods as if the logical
processor was active. The 
IA32_APERF counter does not count during forced idle state. This counting convention allows
the OS to compute 
the average effective frequency of the Logical Processor between the last MWAIT exit and the
next MWAIT entry 
(OS visible C0) by ΔACNT/ΔMCNT * TSC Frequency.
```

Using the above equation we can compute the average effective frequency of the CPUs on a per-in-process-iteration basis.

I'm currently working on having Krun emit two more lists into the results file:
- `aperf_data`: The sum of the differences between the APERF counters at the beginning and end of the iteration for all cores.
- `mperf_data`: as above, but for MPERF.

Note that it would be possible to get a clock speed higher than the actual system clock speed, as we have summed N-cores worth of APERF/MPERF.

Note that we don't yet know how to get the TSC frequency (probably an MSR or via CPUID), but I don't think we care, as it only scales the ratio, which we just want to know is flat over time.

NaN cannot happen. The values come from `uint64_t`s in libkrun.

snim2:
Plot the aperf/mperf data on a second y-axis. 

Example:

aperf-mperf

Fixes #159


snim2:
@vext01 to merge.

ltratt:
The good news is that the graph and axes are quite readable (the legend at the very top of the graph probably needs to move up by a few pixels, so that it doesn't obscure the benchmark name, but that's minor).

Now to the actual data... interesting! IMHO, we need to think carefully about how to report something which only has variation in the 4th significant digit.

One issue that we _might_ be seeing to an extent is that the APERF/MPERF numbers are ever so slightly off, because depending on which we execute first, one will have a slightly higher count than the other (because it counts the execution of that instruction too). That, I think, is why we never see a ratio of 1.0. Related to that is the fact that the shorter the iteration the more the discrepancy between APERF/MPERF will show up.

So, we can either say "we have to explain this to readers" or we can slightly fix the data. One possibility is to work out the difference (in cycles) and modify the data. Another is to say that we'll chop things off at (say) 2DP, which is how Linux (for example) reports CPU MHz back to the user.

I'll be interested in people's thoughts on this -- I don't have a firm opinion on this yet myself.

ltratt:
I should have been clearer. Linux multiplies the ratio by the expected CPU MHz, so it reports back a number like 4000.02MHz. Which means that it's reporting things to 6 significant figures. Hmm.

vext01:
My thoughts:
- Although the a/m perf looks like a zigzag, it's just noise. We are pretty much at 1. Which is good.
- Laurie is right about the non-atomic reads. In my experience, the values are so close that changing the order of the reads is enough to make one higher than the other.
- There are too many decimal places.
- The grids are doubled up -- intentional?

vext01:
>  Linux multiplies the ratio by the expected CPU MHz

I don't understand this. The measurement is not a clock speed.

ltratt:
On bencher2:

``` sh
$ grep MHz /proc/cpuinfo
cpu MHz     : 800.000
cpu MHz     : 3401.000
cpu MHz     : 800.000
cpu MHz     : 800.000
```

3DP even... which I didn't remember!

vext01:
I'm confused. I thought we were discussing APERF and MPERF?

ltratt:
Yes, but APERF and MPERF are typically used to report what speed (in MHz) the CPU has been running at. We don't _have_ to do that, but that's typically what's done.

vext01:
If you wanted a clock speed, we would multiply by the TSC frequency and divide by 4, but I don't see any need. We just want to see that the value is close to 1.

Since we know that the CPU isn't clocking down by much, we could scale the aperf/mperf ratio between 0.9 and 1. This would make the data more interpretable. Thoughts?

bin/plot_krun_results
```diff
@@ -126,13 +135,16 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
     # a flat list of all run sequences.
     if one_page:
         # XXX revise this
```
vext01: Any ideas what this XXX is about?

bin/plot_krun_results
```diff
@@ -211,10 +229,13 @@ def add_inset_to_axis(axis, rect):
     return fig.add_axes([fig_left, fig_bottom, fig_width, fig_height])
 
 
-def draw_subplot(axis, data, cycles_data, title, x_range, y_range, window_size, outliers,
-                 unique, common, median, tukey):
+def draw_subplot(axis, data, cycles_data, aperf_data, mperf_data, title,
```
vext01: I think if we find some time, we should pack all of these data points into some data structure, so we can pass fewer arguments, but now now ;)

ltratt:
If we really only care about the binary "is this data trustworthy or not" question, then another possibility is that we could say that rather than plotting the ratio, we simply check that it's between (say) 0.9980 and 0.9999. If it is, we say "CPU did not throttle" somewhere; if it isn't we say "CPU throttled" and assume the whole graph is suspect.

bin/plot_krun_results
```diff
@@ -391,23 +432,9 @@ def draw_page(is_interactive, executions, cycles_executions, titles, window_size
...
-        if not all([x == None for x in unique]):
-            unique_plotted = True
-    elif unique is not None:
-        unique_plotted = True
-
-    if median or tukey or outliers_plotted or unique_plotted:
-        fig.legend(handles, labels, loc='upper center',
-                   fontsize=LEGEND_FONTSIZE, ncol=10,)
+    # One legend per page.
+    fig.legend(handles, labels, loc='upper center',
```
vext01, line 600: Minor: The placement of the legend looks a little odd in the screenshot.

vext01:
I like the idea, but I think plotting the data is more transparent for a reader. We should do both :)

ltratt:
I think that could make sense.

vext01:
Ah, another thing -- which we can probably tackle later -- we would need a different line style so that when printed in black and white the reader can still make sense of it.

I'm worried that adding two more lines onto the graph (gc/tracing) will make the already busy graph indecipherable. I wonder if we should have separate graphs or switches to turn on/off lines.

ltratt:
If we do the "CPU did/did not throttle" binary thingy, we should make the ratio as tight as is possible. One way to work this out is to look at the speedstep values for the machine, and work out what ratio that implies. For example on bencher6:

```
cpu0: Enhanced SpeedStep 3592 MHz: speeds: 3600, 3400, 3200, 3000, 2900, 2700, 2500, 2300, 2100, 1900, 1700, 1500, 1400, 1200, 1000, 800 MHz
```

I think that means that a ratio <= ~0.944 means that the CPU has throttled to the next speed-step value. So we could be super-cautious and say that a ratio > 0.985 and < 1.000 means "everything is good".

[BTW, I think a ratio > 1 means that turbo was used! So we can add a check for that too. And say "Turbo was used, so data is suspect".]

bin/plot_krun_results
snim2: No, you wrote it.

ltratt:
At the moment, let's not worry about B&W etc. The main thing we need to immediately care about is that _we_ can see what's going on. Once we're satisfied, we can tweak things for publication.

[BTW, I agree that having two horizontal grids for the 2 main y axes is confusing. Fixing that will be hard, unless we can make the intervals on both axes line up.]

snim2:
A couple of quick things:
- There _are_ two line styles here, although they are not easy to tell apart.
- I really do not want to "fix" the issue with spacing around the legend. On the 10-charts-per-page runs there is far too much spacing around the legend, on this test run there is too little. A while ago I tried to deal with this and didn't come up with anything that worked. If we do find a way, we should raise a separate PR and test with both smaller and larger data sets.

vext01:
A ratio of >1 would mean that the actual performance was greater than the reference performance, which would be indeed, fishy.

As for the speed step stuff: assuming P-states can only switch to these discreet values (i'm not sure), you could compute the minimum/maximum theoretical ratio and use those values as y-min/max.

> we should make the ratio as tight as is possible

I'm not sure what you mean. Do you mean "as zoomed in a possible"? That's what we have at current.

ltratt:
By "as tight as possible", I mean that we should be cautious. Just because I think 0.944 is the threshold, we might as well use something much more conservative than that (e.g. 0.98). I agree that we don't know for sure that the speedstep values are discreet (though I _think_ they are, I'm not going to stake my life on it).

vext01:
> [BTW, I agree that having two horizontal grids for the 2 main y axes is confusing. Fixing that will be hard, unless we can make the intervals on both axes line up.]

I'd just remove the grids for non wall-clock data.

ltratt:
Removing the grids is misleading: readers will assume the grid for one axis matches the other, so they'll incorrectly read the values for one y axis.

cfbolz:
or give the second grid a different color (yellow?)

vext01:
>  we might as well use something much more conservative than that (e.g. 0.98)

Let's use [0.98, 1.02] -- therefore allowing us to see if a ratio >1 occurs.

Another thought -- The grid for this line needs only one tick at 1.0 -- is that possible?

ltratt:
In the case of the specific example, the RHS y axis (core cycles) has nice intervals, so we would like to keep those. The LHS y axis has fiddly intervals, so we could perhaps change those to match the RHS. [So e.g. the LHS y axis might be rescaled slightly to be between 0.887 and 1.0399.]

ltratt:
I'm 99% certain the ratio _must_ be below 1.0 (otherwise turbo mode is on, at which point the machine has been setup incorrectly). I'm fine with 0.98 as the lower bound though.

vext01:
I think matching up the axes like that is going to be a matplotlib nightmare

vext01:
We could try something more radical like "sharing both axes" here: http://matplotlib.org/examples/pylab_examples/subplots_demo.html

This solves many of our issues all at once, but requires more engineering.

vext01:
(Obviously, we would have different Y-scales)

ltratt:
Yes, that axis sharing is what I'm suggesting. If matplotlib doesn't do it well automatically, I believe we can specify the intervals ourselves, which isn't going to be too hard for us to do. [I'm lazy though, so if it does it well automatically, I won't complain!]

vext01:
Oh, I thought you were suggesting that we try to make the grids line up. To be clear I am suggesting a new sub-plot for each line where only the x-axis is shared. This will take a while to implement as it would involve a total overhaul of the plotting script.

vext01:
@snim2 We would have to try and use a "gridspec" to make plots appear in clusters of 3 (and later 5)

http://matplotlib.org/users/gridspec.html

bin/plot_krun_results
snim2: I tend to think we should leave this until we have a better idea of what we will be passing around, since there are more events to be plotted, I think.

vext01:
Or maybe if the subplots are created with `sharex=True` the clustering is automatic?

vext01:
I'll have a play with he branch and see what happens.

snim2:
@vext01 just to be clear, when you added the first `twinx` you did not try to unify the `ymin/ymax` like we do with the left hand side y-axis. So, once we have ten plots per page each plot will have a different `yscale` on the RHS axes. Was that intended?

ltratt:
OK, we're talking at cross-purposes then ;) Yes, I'm talking about making "everything on one graph" readable with the y-axes being aligned so that the grid isn't confusing.

cfbolz:
actually, it shouldn't be too hard. isn't it enough to make sure that we have the same number of grid lines on both sides? right now it's 7 vs 8, which causes the problems

ltratt:
Yes, that may do it.

vext01:
> Was that intended?

Not really.

>  I'm talking about making "everything on one graph" readable with the y-axes being aligned so that the grid isn't confusing.

I believe what I am suggesting will solve more issues.

I am suggsting we have 10 clusters of graphs, each one of 10  is one process execution. Each cluster would be 3 vertically stacked graphs sharing an x-axis only.

I like this idea because:
- Colours and line styles are not required.
- Y-axis (and therefore grids) are independent
- No lines can ever overlap
- Can easily add two more lines without making the plots look overloaded.

But it will require a lot of hacking.

Do you follow?

snim2:
OK, so for one process execution, will the measurements appear on each subplot in the stack?

ltratt:
I think I follow, but let's hold off on doing that for the time being. [For one thing, it's going to require a lot more space, and that might be a great trade-off for us, for more than one reason.] I think that a few tweaks to what Sarah's done here is the way to go for now.

vext01:
It's true that it would take more space.

>  will the measurements appear on each subplot in the stack?

Imagine you had 10 of these, where the top axes is the wallclock time, the middle one the core cycles count, and the bottom is the amperf ratio: And everyhing is black.

subplots_demo_03

> I think I follow, but let's hold off on doing that for the time being

OK, but I think we will end up revisiting this when we need to add on GC and tracing -- that's next on my list. I really don't think we have enough real estate for two more lines on the same graph. It's already cluttered.

ltratt:
I'm very happy to revisit this in the future when we've got more stuff. I think what we've got at this moment is pretty good, modulo the changes (grid lines, adding the binary (OK, ternary) "is the CPU throttled/turbo" check) we've suggested.

snim2:
I am dealing with the grids now. @vext01 can you raise an issue for the gridspec subplots please?

vext01:
Thanks @snim2 will do.

vext01:
This closes #159

snim2:
Gridlines / ticks now look like this:

unified_yticks

vext01:
Wow, well done. I was expecting that to be really hard.

vext01:
So just the y-axis scale change now: [0.98, 1.00].

snim2:
This is the aperf/mperf only right? The cycles ticks can be left alone?

vext01:
Correct.

snim2:
After the latest commit the charts look like this:

fixed_yrange

snim2:
OK, are there any more changes to make, apart from the `FIXME` on line 583?

ltratt:
The #core cycles must be to the power of some number I guess? Otherwise, it looks good!

snim2:
If you look at the top right of the chart it says `1e9`

vext01:
The upper bound on amperf should be 1.00 -- aside from that -- I think that is all. @ltratt?

vext01:
> must be to the power of some number I guess

1*10^9 -- see the top right on th axis.

ltratt:
Oh, crumbs, that's tiny -- I missed that! Probably it would be better in "#core cycles (10^9)" in the longer term, but this is fine for now.

snim2:
Hold off on merging this please, I just want to check a couple of things

vext01:
@snim have you tried `:` as a linestyle to see if it is easier to differentiate from the dashed line?

snim2:
Can do

ltratt:
Ah, for the aperf/mperf ratio, I think it should be between 0.98-1.02 for graphing purposes (not 0.975). This might also make the intervals on that axis a bit tidier.

snim2:
OK, I am no longer convinced that we can fix the y-range, here is an example where our yrange does not work:

low-perf

vext01:
This is a case where the CPU was clocked down. Indeed we need to cover this.

It's the empty benchmark -- i suspect it doesn't do enough work to get the CPU clocked up.

vext01:
Although --- having said that, it should just be a tight empty loop. Which _should_ spin up at least one CPU.

snim2:
I think you must be right, the empty ones seem to be low

vext01:
That's it, isn't it -- PyPy is single threaded, so 3 cores are sat twiddling their thumbs.

vext01:
Although you might expect a value of 0.25.

So if we revise our outlook, we hope to see a mostly straight ratio < 1.

If it is close to 1, then all four cores are pegged to the max.

vext01:
Is it possible to define a maximum zoom level on a y-axis?

ltratt:
Ah, I see. This makes visualising things even trickier because we have 4 separate ratios to report (assuming 4 cores)?

vext01:
@ltratt the four ratios are not recorded separately. We would need to go back and re-work libkrun if that is what we want.

vext01:
Sorry, I mean to say: the aperf and mperf values are not collected on a per-core basis in libkrun. They are summed in-library.

ltratt:
Ah, so they're squashed together. I get it.

OK, in which case I would have assumed the ratio should still be 1 (or very, very close to 1). I don't quite understand why we might have ratios of (say) 0.5.

vext01:
Why would it be close to 1?

As i understand mperf increments proportional to the TSC and aperf only increments if the CPU is on (i guess lower clock speeds are achieved by turning the cpu off periodically)

When we sum as we do, we effectively have a CPU with a clock speed 4 times that of the clock speed of the CPU. If 3 cores are off, then that's a ratio of 0.25.

ltratt:
Ah, interesting. That means, I think, that we can't differentiate 4 cores running at 25% speed vs. 1 core running at 100% and 3 cores idle?

vext01:
Correct.

ltratt:
OK, that means we can't detect an important case where a CPU has been throttled. In which case we're going to have to record the cores separately I fear :/ Once we've separated them out, I think visualising so many is going to be super-annoying, at which point just adding a "did/did not throttle" boolean seems the only likely route?

vext01:
So we need to go back to the drawing board then. We need to record per-core amperf counters.

I did raise this in an earlier conversation, but I guess we didn't appreciate the impact back then.

vext01:
So @snim, I think we need to put this on hold.

@ltratt what about core-cycles. Do we need those per-core?

ltratt:
Don't know. Perhaps we should record them per core in case it's useful later. We can always aggregate them in the visualisation, but we can't tear them apart.

snim2:
OK, if 1.0 is the top of the y-range then it's difficult to see the line. Since we unify the gridlines we end up with 1.2 being y-max if we go over 1, but the result is much clearer:

perf-done

snim2:
Before adding more commits here I'm going to bring this up to date with master; the reason for this is that now we have the `csv_to_krun_json` script, we need to make sure that the changes here work with the output of `csv_to_krun_json`.

snim2:
Example page with gridspec:

[multiplot-example.pdf](https://github.com/softdevteam/warmup_experiment/files/503967/multiplot-example.pdf)

ltratt:
That is really interesting -- and really shows that other cores do have short, sharp spikes!

ltratt:
I wonder if we should not use dashed lines for the other graphs, now that we've separated them out? [Or use much closer-together dashes, so we can more easily interpret the curve?]

snim2:
Yes, we should probably use something other than colour. We could also put subplot titles between the stacks, but I'm not so sure how readable that would be.

snim2:
New linestyles:

new_linestyles

vext01:
This is looking good. Well done Sarah -- I know matplotlib is a complete pain!

Some observations and comments:
- We need to indicate what each graph is showing, perhaps on the right hand side of each axes? E.g. "CC core 1" to mean "Core cycles Core 1", "AMR Core 1" to mean "IA32_APERF/IA32_MPERF ration core 1"
- Once you have the above, I would remove line colours and dashing altogether. I find the dashes pretty distracting, and the gap in a dash could hide a spike. The colours won't work once printed black and white, and each graph has only one line now anyway.
- Notice that core 0 tends to have a different behaviour to the others. This is the only "tickful" core on the system. Coincidence? Maybe -- let's wait for more data.
- We should label the axes.
- Do we have the ability to show only a subset of data. Maybe at some point in the paper we only want to show the WCT and the one core where the single-threaded benchmark ran.
- Does `--xlimits` and `-b` and friends still work?
- Does interactive mode still work?

vext01:
I basically said "label the axes" twice in my above comment. Sorry.

vext01:
I'm still very surprised that the amperf ratio is 1 for all cores for a single-threaded benchmark. Either the idle cores are running flat out, the MPERF counter does not increment when the core is off (contrary to what the manual says), or there is a bug. Will investigate.

ltratt:
OK, I'm now pretty convinced that we should be using solid lines for all of the "split apart" graphs.

I suspect we should also normalise the y axis between the various core counts -- at the moment a large spike on (say) core 2 might not actually correspond to _that_ much work.

vext01:
@ltratt I don't think we should normalise the Y-axes. The reason we split the graphs out is so that each individual core could have an independent axes. Remember there was a graph where one core was very busy and three cores were idle, and this made the graph scales very "artificial" and made it hard to see what any one core was doing in isolation as the scale becomes so zoomed out.

Perhaps unifying the scales on all core zeroes and all core ones, etc. might make sense though. I'm not really sure.

ltratt:
I have a cunning plan! On each core, we can have two y axes: the LHS y axis (with, say, a solid black colour) is the normalised y axis (i.e. same for all cores); the RHS y axis is a "local" y axis (with, say, a dashed line). That way we can see both the "global" and "local" context.

snim2:
@ltratt To be clear, you want two lines on each core-count plot?

ltratt:
That's my cunning plan. Of course, it might be a bit too Baldrick-esque...

snim2:
It seems that we can't easily have a gridspec plot inside another plot, which affects our "thumbnail" plots. Am investigating...

snim2:
hotspot_gc_events

ltratt:
Could we attach that as a PDF? I can't really read the PNG (it's too lo-res).

snim2:
Page 4
[hotspot_test.pdf](https://github.com/softdevteam/warmup_experiment/files/518397/hotspot_test.pdf)

ltratt:
Thanks! What units are the GC events given in?

ltratt:
[This is also a very clear example of a single-core benchmark migrating over 3 cores in its lifetime, which is interesting.]

snim2:
Seconds (they are written into the JSON as milliseconds, but I converted them because they would be the only chart in ms rather than s).

ltratt:
Am I being a dummy, or does that mean that the scale is 25000 seconds then? [I must admit, even if the y axis is ms, it still doesn't make sense to me. I was wondering if it's ns or something.]

snim2:
It does seem odd. The original data is cumulative time in ms, and this is the start of the list:

`[21, 21, 21, 22, 22, 22, 23, 23, 23, 23, 23, ...`

So, I read that as their being 21 ms of GC activity before the first iteration has finished.

ltratt:
So it's "how much time (in ms) has each iteration spent doing GC"? Right, that would make more sense. So then the top of the y axis should be 0.025s?

ltratt:
Ah, I think we've _multiplied_ by a thousand, not _divided_ by a thousand.

snim2:
Oops.
[hotspot_test2.pdf](https://github.com/softdevteam/warmup_experiment/files/518403/hotspot_test2.pdf)

ltratt:
Right, that makes a lot more sense! Should we call the axis "GC (secs)"?

Now, the actual data is also interesting, and not totally what I expected, but that's a separate issue :)

snim2:
The cunning plan has been implemented. Results a bit mixed: 
[cycles_test.pdf](https://github.com/softdevteam/warmup_experiment/files/518445/cycles_test.pdf)

ltratt:
Ah, this still has the wrong data for PyPy, right?

That said, I suppose the other alternative is to simply plot one line for each core, and normalise it to the other cores (so cores that only do a little smidgen of work will have a flat line around 0.0)?

snim2:
Yes, this is from before the PyPy fix was pushed. OK, I think this is what you are after: 
[cycles_normalise_per_pexec.pdf](https://github.com/softdevteam/warmup_experiment/files/518472/cycles_normalise_per_pexec.pdf)

ltratt:
Yes, I think that's more-or-less it, although I'm not sure what the core Y axis is. Have we just taken the maximum value and set that as 1.0? I guess we could do that. I was originally thinking -- but it may or may not be a good idea! -- that we'd take the maximum cycle count and use that as the same for the 4 core graphs. So the shape of the graphs would stay the same as what you have now, but the axis labelling would change to something a little more meaningful (I think).

snim2:
I think it depends whether you want to show which cores are active or something more subtle. For reference, this is what you get if you just change the ylimits: 
[cycles_just_change_ylim.pdf](https://github.com/softdevteam/warmup_experiment/files/518474/cycles_just_change_ylim.pdf)

ltratt:
Perfect! That's what we want. This then means we can compare cycle counts across machines (should we want to).

snim2:
In that case, I think we should start reviewing this before it gets any larger!

ltratt:
OK, so if you think this is ready for reviewing now, I'm going to suggest you do some pre-squashing, and see if you can chunk this up into smaller commits that can be reviewed on their own, almost as mini-PRs. Is that possible?

snim2:
Sure, do you want me to bring it up to date with master now, or after reviewing?

ltratt:
Probably best to bite the bullet and do it at the same time, but that's up to you.

snim2:
Ready for review.

bin/plot_krun_results
```diff
@@ -8,26 +8,29 @@ import datetime
 import math
 import matplotlib
 matplotlib.use('Agg')
-from matplotlib.collections import LineCollection
-import matplotlib.pyplot as plt
+import matplotlib.gridspec as gridspec
```
ltratt: Simplify to `from matplotlib import gridspec, pyplot`?

bin/plot_krun_results
ltratt: Let's delete the comment, since we have no idea why it's there.

bin/plot_krun_results
```diff
@@ -202,6 +229,9 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
         fig, export_size = draw_page(is_interactive,
```
ltratt: Compress function call args.

bin/plot_krun_results
```diff
@@ -238,36 +268,74 @@ def add_inset_to_axis(axis, rect):
...
+    if instr_data:
+        n_rows += len(instr_data_narrays)
+    inner_grid = gridspec.GridSpecFromSubplotSpec(n_rows, 1, grid_cell, hspace=0.2)
+    # Collect handles and labels, so that we can draw one legend per page.
+    handles_labels = [list(), list()]
+    # Plot wall clock times.
+    axis1 = pyplot.subplot(inner_grid[n_rows - 1, 0])
+    axis1.plot(iterations, data_narray, label='Measurement', color=LINE_COLOUR,
+               zorder=ZORDER_DATA)
+    # If there are outliers, highlight them.
```
ltratt, line 309: There seems to be a common function here which is something like:

``` python
def f(l, x_bounds):
    out1 = [val for val in l if val >= x_bounds[0] and val < x_bounds[1]]
￼    out2 = [val - x_bounds[0] for val in out1 if val >= x_bounds[0] and val < x_bounds[1]]
    return (out1, out2)
```

If I've got that right, then it's probably best separated out.

bin/plot_krun_results
```diff
@@ -384,26 +529,29 @@ def draw_page(is_interactive, executions, cycles_executions, titles, window_size
...
 
-    fig, axes = plt.subplots(n_rows, n_cols, squeeze=False)
+    fig = pyplot.figure()
+    outer_grid = gridspec.GridSpec(n_rows, n_cols)
 
     index, row, col = 0, 0, 0
     while index < n_execs:
         data = executions[index]
-        cycles_data = cycles_executions[index]
+        cycles_data = cycles_executions[index] if cycles_executions is not None else None
```
ltratt: Let's please not use Python's horrible ternary operator -- it leads to buggy code.

bin/plot_krun_results
```diff
@@ -727,6 +873,11 @@ def create_cli_parser():
                         help=('Name of the PDF file to write to. If no file is '
                               'specified, charts will be displayed interactively '
                               'and not written to disk.'))
+    parser.add_argument('--wallclock-only',
```
ltratt, line 884: Compress function args.

warmup/vm_instruments.py
```diff
@@ -0,0 +1,71 @@
...
+            for index in xrange(1, len(cumulative_times)):
+                times_secs.append((cumulative_times[index] - cumulative_times[index - 1]) / 1000.0)
+            assert len(times_secs) == len(cumulative_times)
+            self.chart_data.append([ChartData('GC (secs)', times_secs, 'GC events')])
+
+
+# Mapping from VM name -> parser class.
+# This enables the main scripts to parse instrumentation data based only
+# on the vm:bench:language triplets found in Krun data files.
+INSTRUMENTATION_PARSERS = {'Hotspot': HotspotInstrumentParser}
```
ltratt: Not sure if this is a typo, or if the bug is elsewhere in the repo, but the capitalisation should be `HotSpot`.

bin/plot_krun_results
snim2: There is >100 lines of code to set up the CLI parser, do you want it all "compressed" in this next commit, or just the lines that have changed in this PR?

warmup/vm_instruments.py
vext01: That's likely the internal name of the key in the config file. In hindsight, I wish I had made all the keys lower case.

snim2: This is a JSON key, so it can't be changed.

ltratt: OK, let's leave the json key as is, but fix the class name.

bin/plot_krun_results
ltratt: Maybe let's do a separate PR and get this fixed later.

bin/mark_outliers_in_json
```diff
@@ -41,9 +41,9 @@ def main(in_files, window_size, threshold):
         all_outliers = dict()
         unique_outliers = dict()
         common_outliers = dict()
-        for bench in krun_data[filename]['data']:
```
vext01, line 44: Sorry for not updating this with the per-core branch.

bin/plot_krun_results
snim2: In that case I will raise an Issue

bin/plot_krun_results
```diff
@@ -75,6 +78,9 @@ UNIQUE_SIZE = 80
 COMMON_COLOR = 'r'
 COMMON_MARKER = '*'
 COMMON_SIZE = 80
+CYCLES_COLOR = 'green'
```
vext01, line 80: Not sure colours will work in print, but we can look at that later.

bin/mark_outliers_in_json
snim2: No worries

bin/plot_krun_results
vext01: I think it was because in an earlier change I used multi-page mode first, and came back to single page later.

bin/plot_krun_results
vext01: Yeah, it's been on my list to try and box these args up into some data structure(s?). But it's probably very low priority.

bin/plot_krun_results
vext01: I reckon later, we may find ourselves needing (e.g.) only core 3's cycle graph. Cross that bridge when we fall off of it.

snim2: Possibly, but let's find out if we have a use case first. To be clear: `--wallclock-only` is something that is needed to chart data produced by the `csv_to_json_krun` script.

warmup/vm_instruments.py
vext01: Here is the code to parse the PyPy instrumentation data:
https://gist.github.com/vext01/3b685296d938cd78363dc6cd4e22754d

snim2:
Good to squash?

ltratt:
Fine for me. You should wait for @vext01 to confirm as well though.

vext01:
Fine for me.

snim2:
Squashed.

ltratt (approved):

ltratt:
Ah, I think you might quickly need to rebase against `origin/master` (after doing `git fetch`) to avoid a merge commit.

snim2:
OK, done (I think)

PR #167, opened 2016-09-23
Changepoint analysis

### LINKED ISSUE #166: Add changepoints to chart

snim2:
Needs a lot of squashing!

Fixes #166


snim2:
Note to self: plot means rather than changepoints

snim2:
This is the current output:

current_changepoint

there are some improvements to make to the charts (e.g. some of the gridlines appear in front of the plots) but these need to be dealt with on the aperf/mperf PR.

ltratt:
I like this. One thing, however, I think we're going to have to address is changepoints that are really short. We need some way to say to the user "actually, there's a changepoint under here that's only 1 or 2 iterations long, so you can't even see it". Maybe a big blob or ...

[Admittedly, a possible alternative is to say that changepoints have to be at least _n_ iterations long for us to want to plot them. We might need to ask Rebecca whether something like that is a good idea though.]

snim2:
I'm not keen to make that sort of change in this particular PR. We need to make sure that the Python code is producing the same results as the R code, so we need to keep the two as identical as possible until we've made careful checks. After that I'm sure there will be a number of changes to make to both the changepoint marker and the classifier.

ltratt:
OK, I'm happy leaving that to later.

ltratt:
Is this ready for review and possible merging?

snim2:
Yes please.

bin/mark_changepoints_in_json
```diff
@@ -0,0 +1,144 @@
...
+import rpy2.interactive.packages
+import rpy2.robjects
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+from warmup.krun_results import read_krun_results_file, write_krun_results_file
+
+
+def main(in_files, window_size):
+    cpt = rpy2.interactive.packages.importr('changepoint')  # Import once.
```
ltratt: I'm not sure I understand the comment here.

bin/mark_changepoints_in_json
```diff
@@ -0,0 +1,144 @@
...
+        root_name = os.path.splitext(basename)[0]
+    base_out = root_name + '_changepoints.json.bz2'
+    return os.path.join(directory, base_out)
+
+
+def create_cli_parser():
+    """Create a parser to deal with command line switches.
+    """
+    script = os.path.basename(__file__)
+    description = ('Write changepoints and classifications into Krun '
```
ltratt: Let's use a `"""` for this (it's really hard to read ATM).

bin/mark_changepoints_in_json
```diff
@@ -0,0 +1,144 @@
...
+                   'changepoints and this will be written into the name\nof '
+                   'the output file, for example if the input file is:\n'
+                   'results_outliers_200.json.bz2 and the window size is 200, '
+                   'the output of this\nscript will be stored in file: '
+                   'results_outliers_w200_changepoints.json.bz2.'
+                   '\n\nExample usage:\n\n\t$ python %s results1.json.bz2\n'
+                   '\t$ python %s  --window 250 results1.json.bz2 '
+                   'results2.json.bz2\n' % (script, script))
+    parser = argparse.ArgumentParser(description)
+    parser.add_argument('json_files',
```
ltratt: Arguments can be compressed onto fewer lines.

ltratt:
A few minor comments -- nothing too much.

bin/mark_changepoints_in_json
snim2: We import the R package once, and pass it to the method where it is used, rather than importing it inside the method.

ltratt: Ah, OK, so it is for the obvious reason (I wondered if there was something deeper I was missing). I suspect the comment's probably best removed.

ltratt:
Looks good to me! Squash as you see fit, and then I can merge.

snim2:
Cool. Will also bring up to date with master...

ltratt:
Good point.

snim2:
OK, squash done.

ltratt:
This will seem silly... but is there any reason that "AND" is capitalised in the commit message? If not, can we quickly fix that? Thanks!

snim2:
Probably left over from an old commit message or something. Fixed.

ltratt:
Thanks!

ltratt (approved):

PR #168, opened 2016-09-27
Convert CSV files to Krun JSON files

snim2:
New `bin/csv_to_krun_json` script. Requires the user to specify language, VM name and uname string on the command line.

Needs squashing.

Results from the preliminary Da Capo run are here:

[dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/495581/dacapo.pdf)

### DIFF
--- a/bin/csv_to_krun_json
+++ b/bin/csv_to_krun_json
@@ -1,0 +1,100 @@
+#!/usr/bin/env python2.7
+"""
+Convert a CSV file of timing measurements to JSON, suitable for using with the
+other scripts here.
+"""
+
+import argparse
+import csv
+import os
+import os.path
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', 'warmup'))
+from warmup.krun_results import write_krun_results_file
+
+
+def main(in_files, language, vm, uname):
+    for filename in in_files:
+        data_dictionary = { 'data': dict(), # Measurement data.
+                   'core_cycles_data': dict(),
+                   'aperf_data': dict(),
+                   'mperf_data': dict(),
+                   'instr_data': dict(),
+                   'audit': dict(),
+                   'config': '',
+                   'reboots': 0,
+                   'starting_temperatures': list(),
+                   'eta_estimates': list(),
+                   'error_flag': list(), }
+        with open(filename, 'r') as fd:
+            data_dictionary['audit']['uname'] = uname
+            reader = csv.reader(fd)
+            reader.next()  # Skip first row, which contains column names.
+            for row in reader:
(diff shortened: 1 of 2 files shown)


snim2:
Ugh, I think I branched from dacapo rather than master :(

ltratt:
Am I correct in thinking this is based off of CF's current branch? If so, we should probably rebase against master, as they're kind-of separate branches.

ltratt:
Ah, OK :) So if we just rebase against master and force push here (not against master ;) ), all should be good.

snim2:
I should probably rebase and remove the dacapo commits from this branch; do you want me to do that now?

ltratt:
I think we should, otherwise it might mess up merging the dacapo PR later. [This PR will almost certainly be merged first.]

snim2:
Done!

bin/csv_to_krun_json
```diff
@@ -0,0 +1,103 @@
...
+                   'audit': dict(),
+                   'config': '',
+                   'reboots': 0,
+                   'starting_temperatures': list(),
+                   'eta_estimates': list(),
+                   'error_flag': list(), }
+        with open(filename, 'r') as fd:
+            data_dictionary['audit']['uname'] = uname
+            reader = csv.reader(fd)
+            firstrow = True
```
ltratt: Wouldn't it be easier to say something like `reader.next()` (or `reader.readline()` or whatever it's called) to munch the first line, rather than having to have a variable and an `if` check inside the `for` loop?

bin/csv_to_krun_json
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env pypy
```
ltratt: I don't think this script is particularly slow? If so, we should probably just use `python2.7` (or whatever version of CPython is capable of running this), simply because more people are likely to have it installed already.

ltratt:
OK, let's squash this, then I'll merge.

bin/csv_to_krun_json
cfbolz: it would be something like:

```
it = iter(reader)
next(it)
for ... in it:
```

ltratt: Sarah's subsequent commit has a minor variation of this.

cfbolz: ah, makes sense. (why doesn't it show me that this is an outdated part of the diff any more?)

ltratt: I was wondering the same thing about 30 minutes ago. Dunno, guess it's part of the (overall, much better) new UI.

snim2:
Squashed...

ltratt (approved):

snim2:
Thanks

PR #169, opened 2016-09-28
add DaCapo as an external benchmark

cfbolz:
- download the DaCapo .jar file in build.sh
- add a script to run the benchmarks the right number of times
- convert the resulting code to a krun json file
- fix the path insertion code in csv_to_krun_json


cfbolz:
`run_shell_cmd` is code from krun, where we this is properly implemented, right @vext01?

vext01:
It returns a tuple `stdout, stderr, returncode`. It uses `subprocess.Popen()` then `communicate()` (which blocks) in order to capture the outputs.

I don't know the exact details of how `communicate()` consumes file descriptors, but I imagine it reads the pipes as they become ready. You should read the source to find out.

If it doesn't do what you want, then you could do something like we do for benchmarking in Krun:
https://github.com/softdevteam/krun/blob/master/krun/vm_defs.py#L242

vext01:
(In Krun we do not use `run_shell_cmd` to run benchmarks, only to issue various other commands, e.g. to change CPU governor)

ltratt:
Do we think it's OK for running benchmarks?

vext01:
It depends how it is implemented. I recommend reading the CPython source code.

FWIW, the code I linked to above came about because:
- I wanted to be able to see debug output as it appeared, rather than all in one go at the end.
- Because of some I/O related concern which @ltratt had, but I don't remember the details of.

Do you remember Laurie?

cfbolz:
Ah, so the careful code for benchmarks is in `krun/vm_defs.py` and I can't get at it easily, right? Should I factor that out into `utils.py` so that we have solved the problem once and for all?

ltratt:
I'd probably be happier if we did whatever normal Krun does (since we put a lot of thought into that).

cfbolz:
Yes, sounds like a plan, apart from the fact that right now that's tied in with the iteration runners, so I can't use it in this context.

vext01:
I'd be happy for it to move into `util.py`.

cfbolz:
Ok, I'll prepare a pull request for krun then.

ltratt:
Moving it out works for me.

Out of curiosity, will the Dacapo benchmarks run with rebooting? [Hopefully "yes" :)]

ltratt:
Thanks!

vext01:
Perhaps give the function a more appropriate name in doing so.

vext01:
I'm a little confused as to why we aren't krunifying this, if we want stuff like rebooting.

cfbolz:
rebooting: no, they aren't tied into krun that much (which as far as I understood you was not the plan).

ltratt:
@cfbolz Ok, a missing bit of context from yesterday's workshop. Alastair Reid (ARM) pointed out that memory address spaces tend to change benchmark results if you repeatedly run things. So rebooting might be useful, but there again we can justify things on the basis of "this is what people normally do" (and that's a powerful argument IMHO.)

cfbolz:
Integrating it more fully into krun is going to be quite messy: krun assumes you use the iteration runner, which we can't easily do, because dacapo has its own (and the point is anyway to use that).

ltratt:
I agree that fully Krun-ifying it might be too hard (and might, ironically, cause people to say "you accidentally messed up the bencmark's naturally nice behaviour). Rebooting is a nice optional extra that's semi-independent of full Krun-ification though.

vext01:
Let's take a step back. What is the aim of this?
- a) To show what happens when you do what others do?
- b) To run someone else's benchmarks in a controlled fashion?

if a:
  Just do it simple without reboots

if b:
  krunify_if_possible()

cfbolz:
I'm going to focus on factoring out the krun code to run subprocesses first, then we can see what to do next.

ltratt:
Let's go with a).

cfbolz:
Ok, now the new function is used, so Laurie's concerns are fixed. What do you think?

snim2:
lgtm

ltratt:
If Edd's OK, then squash, and then I'll merge.

vext01:
fine.

ltratt (approved):

PR #174, opened 2016-10-04
Integrate Octane.

ltratt:
_Not ready for merging yet_

Here's my first attempt at merging the Octane benchmark suite in. It's not Krun-ified (and we may never do so), and this PR will only tackle V8 (spidermonkey to come later, hopefully).

Octane benchmarks are, mostly, very quick to run so each benchmark says how many times it should be run to be "long enough". We then wrap that in a second loop of 2000 iterations. So the patch makes things look a bit like:

``` python
for i in range(2000):
  for j in range(benchmark.howmany):
    run_bench()
```

At the moment the output is of the form:

```
<bench>
<iter 1 in ms>
<iter 2 in ms>
...
```

e.g.:

```
richards
727
598
599
```

What's the right format to make it easy to graph the resulting data?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,8 @@ bench-da-capo:
 	bin/csv_to_krun_json dacapo.csv
 
 bench-octane:
-	cd ${PWD}/extbench/octane && LD_LIBRARY_PATH=${PWD}/krun/libkrun ${PWD}/work/v8/out/native/d8 run.js
+	PYTHONPATH=krun/ ${PYTHON} extbench/runoctane.py
+	bin/csv_to_krun_json octane.v8.results
 
 # XXX target to format results.
 
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,9 @@ bench-da-capo:
 	PYTHONPATH=krun/ JAVA_HOME=${JAVA_HOME} ${PYTHON} extbench/rundacapo.py | tee dacapo.csv
 	bin/csv_to_krun_json dacapo.csv
 
+bench-octane:
+	cd ${PWD}/extbench/octane && LD_LIBRARY_PATH=${PWD}/krun/libkrun ${PWD}/work/v8/out/native/d8 run.js
+
 # XXX target to format results.
 
 clean: clean-benchmarks clean-krun
--- a/build.sh
+++ b/build.sh
@@ -516,6 +516,21 @@ fetch_dacapo_jar() {
     wget "http://downloads.sourceforge.net/project/dacapobench/9.12-bach/dacapo-9.12-bach.jar?r=https%3A%2F%2Fsourceforge.net%2Fprojects%2Fdacapobench%2Ffiles%2F&ts=1474888492&use_mirror=freefr" -O ${HERE}/extbench/dacapo-9.12-bach.jar  || exit $?
 }
 
+
+OCTANE_V=4852334f
+fetch_octane() {
+    echo "\n===> Download Octane\n"
+
+    if [ -d "${HERE}/extbench/octane" ]; then return; fi
+
+    cd ${HERE}/extbench
+    git clone https://github.com/chromium/octane || exit $?
+    cd octane
+    git checkout ${OCTANE_V} || exit $?
+    patch < ${PATCH_DIR}/octane.diff || exit $?
+}
+
+
 fetch_external_benchmarks() {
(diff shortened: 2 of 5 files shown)


vext01:
Would you want the graphs in the same format as krunified results?

ltratt:
We should just do whatever we did for the dacapo branch I think?

snim2:
So, the `bin/csv_to_json` script takes data of this form:

```
processnum,bmark,0,1,2...
0,spectralnorm,1.2,1.5,2.6...
...
```

The first row should contain column headings, every subsequent row should contain a process execution number, followed by a benchmark name, followed by timings for each iteration.

When you run `bin/csv_to_krun_json` you will also need to pass in a VM name and a `uname -a` string. The `bin/csv_to_krun_json --help` should explain.

ltratt:
OK, I think I can deal with that.

snim2:
When you are ready for a review, if you could attach an example .csv file here that would be very helpful.

ltratt:
Are timings in seconds?

snim2:
Yes

ltratt:
OK, this now produces output in the correct format, and I think is ready for review and possible merging.

snim2:
Have you got an example .csv you can attach to a comment?

ltratt:
It's too big to sensibly upload here (which is why I sent you one by email earlier).

ltratt:
Gah.

snim2:
OK, and the format hasn't changed since then?

ltratt:
"Too big" is wrong, but the general point stands. Yes, the format is identical.

Makefile
```diff
@@ -64,6 +64,10 @@ bench-da-capo:
 	PYTHONPATH=krun/ JAVA_HOME=${JAVA_HOME} ${PYTHON} extbench/rundacapo.py | tee dacapo.csv
 	bin/csv_to_krun_json dacapo.csv
 
+bench-octane:
+	${PYTHON} extbench/runoctane.py
+	bin/csv_to_krun_json octane.v8.results
+
 export-graphs:
 	${PYTHON} export_all_graphs.py
```
snim2: Well, this is interesting. I wasn't aware that `export-graphs` was a Makefile target, and I don't _think_ we have an export_all_graphs script. Also, all the scripts have changed, so this should target should probably be deleted, in this PR or another one.

ltratt: It's not part of this PR, so should be done in a separate one IMHO.

[I think the makefile needs a bit of general re-love anyway, so we should relook at it once the next couple of PRs are merged in.]

snim2:
OK, this looks OK to me, but @vext01 knows more about the benchmark builders than I do.

ltratt:
Probably the most critical part is the diff to the Octane suite itself.

build.sh
```diff
@@ -566,6 +581,7 @@ case `uname` in
     build_initial_krun
     fetch_external_benchmarks
     fetch_dacapo_jar
+    fetch_octane
```
vext01, line 584: Will we run octane on BSD? If so add to the BSD case.

ltratt: I was thinking of doing that as a separate PR, as I think I might have to do a bit of VM patching.

ltratt: [Plus at the point that I tried to do it at the weekend, krun didn't build on OpenBSD, which sort-of discouraged me ;)]

extbench/runoctane.py
```diff
@@ -0,0 +1,58 @@
...
+                # execution.
+                csvf.flush()
+                os.fsync(csvf.fileno())
+                time.sleep(3)
+
+                stdout, stderr, rc = run_shell_cmd_bench(jsvm_cmd, platform)
+                if rc != 0:
+                    sys.stderr.write(stderr)
+                    sys.exit(rc)
+                times = None
```
vext01, line 39: If I were to nitpick, I'd add a comment to each of the cases here. I think one consumes the next measurement, and the other flushes a whole execution's worth of measurements to disk. Am I right?

vext01:
This looks fine to me.

ltratt:
If you're OK with that recent commit, I think we're ready to squash and merge?

snim2:
Sounds good

vext01:
Thanks. That's all my comments. Go ahead.

ltratt:
Squashed and ready to merge.

vext01 (approved):
OK

PR #175, opened 2016-10-04
Per-core VM changes

vext01:
These are changes for the warmup experiment that correspond with https://github.com/softdevteam/krun/pull/251

Mostly updates to patches to expose the libkruntime functions, but also I noticed that depot_tools was being fetched as a movingn target. Fixed it to the code of the same date as v8.


vext01:
Don't forget this one too.

ltratt:
Is there a reason we use a C `int` in the V8 patch, vs. a specifically sized int?

vext01:
Can you point out which one?

ltratt:
`DOUBLE_ARG_TO_CINT_CHECKED` and friends.

vext01:
Ah, those ints are only used for mdata indicies and for core indicies. They don't get very big.

ltratt:
Ah, could we maybe clarify that somewhere? I didn't fully pick that up. Maybe say something like "These macros are only intended for converting numbers that are known to be small (smaller than 32 bits)"?

vext01:
The name of the macro says it all doesn't it?

ltratt:
It says what it _does_, but not _why_ it does it that way.

ltratt:
OK, so maybe my suggested edit wasn't a good one. Maybe it should say "This is used in cases where we know we're dealing with very small numbers that can be represented in a few bits (certainly less than 32)"?

vext01:
I didn't see this comment before my last push. Is it OK, or should I revise?

ltratt:
Let's maybe revise it a bit further to save dumbies like me being confused in the future.

vext01:
Fixed.

[I think you would have realised what was going on when you saw the result only being passed to libkrun as index arguments of type int.]

ltratt:
OK, I'm now fine with this. If Sarah's OK, squash as you wish, and then I'll merge.

snim2:
Sure, let's merge

vext01:
Squashed.

ltratt (approved):

PR #176, opened 2016-10-04
Make the dacapo runner tee-less.

ltratt:
Tee can do unpredictable things at unpredictable times, so we'd prefer it not to
be running at the same time as our benchmarks. Instead write the CSV to a file,
and flush and fsync at known safe points to lessen the chances of the OS doing
nasty things whilst we're benchmarking.

Whilst I'm here, make this look a bit more like the (upcoming) octane runner
which a) gives a minimalistic progress meter b) makes it easier to add new VMs
to the runner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,9 +60,9 @@ bench-startup-no-reboots: build-startup build-benchmarks
 bench-startup-with-reboots: build-startup build-benchmarks
 	${PYTHON} krun/krun.py --reboot startup.krun
 
-bench-da-capo:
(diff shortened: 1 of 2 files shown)


ltratt:
[I think the timings in this file are in ms. Should we convert them to seconds?]

snim2:
Yes, we have assumed wallclock times are in seconds in all our scripts, so we probably want to keep that consistent.

ltratt:
OK, will raise a second PR after this one.

cfbolz (approved):
This looks good to me. In the long run, I wonder whether you should pull out the repeated "flush, fsync, sleep" into a small library, to not have to remember the precise steps for every new external bench.

PR #177, opened 2016-10-04
Remove redundant target.

snim2:


ltratt:
Can we just quickly justify why this is redundant? [I think the script it's referenced has disappeared, but maybe there's some other context I don't know about.]

snim2:
If that Python script was ever in the repo it isn't now. Also the scripts have been refactored, so even if we had the file, it would no longer work with master.

ltratt:
OK, makes sense to me. Will merge now.

ltratt (approved):

PR #182, opened 2016-10-05
Add Spidermonkey to Octane.

ltratt:
To get things to build on OpenBSD, we need to import some more GCC patches so
that gcc can deal with -rdynamic.

Note, on bencher6, this currently creates a binary which gets in an endless sequence of:

```
 30072 js       RET   sched_yield 0
 30072 js       CALL  sched_yield()
```

(i.e. doesn't do anything useful). This might be fixed by upgrading to a more recent version of OpenBSD?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -28,9 +28,13 @@ check_for java
 check_for javac
 check_for xzdec
 check_for wget
+check_for virtualenv
 
 case `uname` in
-    Linux*) PATCH_ARGS=--backup
+    Linux*)
+        PATCH_ARGS=--backup;;
+    OpenBSD*)
+        check_for gm4;;
 esac
 
 which pypy > /dev/null 2> /dev/null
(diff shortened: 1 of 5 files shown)


ltratt:
I'm currently seeing if this builds a useful binary on a more recent version of OpenBSD, so don't merge yet please.

build.sh
```diff
@@ -508,6 +512,40 @@ build_hhvm() {
 }
 
 
+# autoconf-2.13 is needed to build spidermonkey
```
vext01, line 515: I still can't believe this! :8ball:

build.sh
```diff
@@ -508,6 +512,40 @@ build_hhvm() {
...
+build_autoconf() {
+    echo "\n===> Download and build autoconf-2.13\n"
+    if [ -d ${wrkdir}/autoconf-2.13 ]; then return; fi
+    cd ${wrkdir} || exit $?
+    wget http://ftp.gnu.org/gnu/autoconf/autoconf-2.13.tar.gz || exit $?
+    tar xfz autoconf-2.13.tar.gz || exit $?
+    cd autoconf-2.13
+    ./configure --prefix=${wrkdir}/autoconf-inst || exit $?
+    make install || exit $?
+    cd ${wrkdir}/autoconf-inst/bin
```
vext01, line 525: Is it worth installing into a dir with a version number, in case should need a different autoconf for something else?

ltratt: I don't think it's worth being that fiddly, unless it turns out that we really need to do so. I mean, none of the other VMs is remotely fussy about this ATM.

build.sh
```diff
@@ -508,6 +512,40 @@ build_hhvm() {
...
+}
+
+
+SPIDERMONKEY_VERSION=465d150bc8be # FIREFOX_AURORA_50_BASE
+build_spidermonkey() {
+    echo "\n===> Download and build SpiderMonkey\n"
+    if [ -d ${wrkdir}/spidermonkey ]; then return; fi
+    cd ${wrkdir} || exit $?
+    wget -O spidermonkey.tar.bz2 http://hg.mozilla.org/mozilla-central/archive/${SPIDERMONKEY_VERSION}.tar.bz2 || exit $?
+    bunzip2 -c - spidermonkey.tar.bz2 | tar xfp - || exit $?
```
vext01, line 537: Why not `tar jxf spidermonkey.tar.bz2`?

vext01: `jxfp` maybe

ltratt: I don't know that all tars have that option, but I do know that the above command works reliably on a large number of systems.

vext01: It used to be the case that OpenBSD didn't support `-j` but it's been there for a while now.

I'm happy either way tbh.

vext01:
This looks OK, but I can't vouch for the scary GCC patch. Looks like it came from OpenBSD-current, whereas we are using 5.8 on bencher6, which might have linker differences to -current.

build.sh
ltratt: I'd prefer to stick to the known portable route, rather than being a bit clever, and then finding out down the line that it's not always supported.

ltratt:
Yes, the GCC patch came from OpenBSD ports (as did the others, I assume?). The GCC patch is relatively benign, as far as I can tell, and seems to work fine... without it, we can't compile SpiderMonkey as it uses `-rdynamic` which our GCC doesn't define. The patches mean that it then works (or, at the very least, builds).

vext01:
The other patches came from ports, yeah. But have you checked the patch is the same in the 5.8 branch, is what I mean. In all likelihood it is...

ltratt:
We won't be using 5.8 for the final run anyway...

vext01:
The patch in current changed 2 months ago, for example, which might impact 5.8, but would not impact 6.0.

I'm going to build on my -current and see if I get the same.

ltratt:
OK, so we get a build of SM on OpenBSD, but it enters an infinite spin loop. For the time being "don't run it on OpenBSD" is good enough.

If you're OK with the rest of the PR, I think I'm happy to start squashing.

vext01:
Discussed elsewhere. Octane gets stuck in a spinlock on both 5.8 and openbsd-current.

We agreed to merge this, but not run it on OpenBSD.

Once you disabled running on OpenBSD, squash and I can merge.

ltratt:
Agreed. Squashed and ready for merging.

vext01 (approved):

PR #183, opened 2016-10-07
Speed up builds by having make run multiple simultaneous jobs.

ltratt:
This speeds up builds significantly. I know it's a rounding error compared to the time to run experiments, but when one has to rebuild to test something, saving a couple of hours definitely helps.

Should only be considered for merger after #182 has been merged.

### DIFF
--- a/bin/mark_changepoints_in_json
+++ b/bin/mark_changepoints_in_json
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+"""
+Write changepoint and classification information into JSON files.
+"""
+
+import argparse
+import copy
+import os
+import os.path
+import rpy2
+import rpy2.interactive.packages
+import rpy2.robjects
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+from warmup.krun_results import read_krun_results_file, write_krun_results_file
+
+
+def main(in_files, window_size):
+    cpt = rpy2.interactive.packages.importr('changepoint')
+    print 'Using changepoint library version %s' % cpt.__version__
+    assert cpt.__version__ >= '0.2.2', 'Please update the changepoint library.'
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -52,6 +52,20 @@ case `uname` in
 esac
 check_for ${GMAKE}
 
+# How many cores does this machine have (so we can pass a sensible number to
+# gmake -j)?
+num_cores=1
+case `uname` in
+    OpenBSD) num_cores=`sysctl hw.ncpu|cut -d "=" -f 2`;;
+    Linux) num_cores=`cat /proc/cpuinfo|grep cores|head -n 1|cut -d ":" -f 2|sed "s/ //g"`;;
```
vext01: On linux, you can use `nproc`:

```
root@bencher3:~# nproc
4
```

build.sh
```diff
@@ -52,6 +52,20 @@ case `uname` in
...
+esac
+# Keep num_cores to safe limits for pretty much any machine out there.
+if [ $num_cores -le 0 ]; then
```
vext01: Does this case happen? Should we perhaps bomb out instead?

vext01:
Looks OK bar minor comments.

build.sh
ltratt: Aha!

build.sh
ltratt: This is just paranoia on my part: since we don't _need_ to increase the number of jobs, 1 is always a safe default.

ltratt:
If that latest commit works for you, I'll squash, and then we should be good for merging.

vext01:
Looks good. I'll do a quick test on a bencher.

ltratt:
OK.

vext01:
OK, i see gcc being built on 4 busy cores.

Squash?

ltratt:
OK, squashed.

PR #184, opened 2016-10-07
Add Graal to the DaCapo runner.

ltratt:
_Not yet ready for merging_ [and, when it is, shouldn't be merged until after #183]

Add Graal to the DaCapo runner so we can compare Graal and HotSpot on the DaCapo benchmark suite.

[The difficulty here was purely trying to work out from `warmup.krun` how to call `mx`!]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,8 @@ bench-startup-with-reboots: build-startup build-benchmarks
 
 bench-dacapo:
 	PYTHONPATH=krun/ JAVA_HOME=${JAVA_HOME} ${PYTHON} extbench/rundacapo.py
-	bin/csv_to_krun_json dacapo.hotspot.results
+	bin/csv_to_krun_json -u "`uname -a`" -v Graal -l Java dacapo.graal.results
+	bin/csv_to_krun_json -u "`uname -a`" -v HotSpot -l Java dacapo.hotspot.results
 
 bench-octane:
 	PYTHONPATH=krun/ ${PYTHON} extbench/runoctane.py
--- a/extbench/rundacapo.py
+++ b/extbench/rundacapo.py
@@ -3,6 +3,9 @@
 import csv, os, sys, time
 from krun.platform import detect_platform
 from krun.util import run_shell_cmd_bench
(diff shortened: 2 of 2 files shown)


ltratt:
Ready for review @vext01. [This branch ended up much simpler than I expected.]

vext01:
Looks ok to me. I take it you got the command from the graal vm def?

side-note: I know that command has changed upstream already >.<

ltratt:
Yes, I more-or-less got it from there.

ltratt:
OK, so if you're OK with this, I think it can be merged.

vext01:
more-or-less? Did you check the JIT is definitely enabled? It's easy to get wrong.

There's a sanity check you could lift a JIT check from.

ltratt:
You saw the resulting data -- I think it's pretty clear that it's at least somewhat working.

vext01:
If you get it wrong you will still see warmup, but from the standard java JIT.

Having gotten it wrong multiple times myself, I think we should quickly check (you can remove the JIT check after if you want).

It would be pretty unfair and potentially embarrassing if we reported graal performance with the wrong JIT.

ltratt:
For clarity, when I said "more-or-less", I meant that I had tweaked the path names -- the invocation is otherwise the same. If you see the data I emailed, it's very clear that we're not using the default HotSpot system.

vext01:
OK, if you are certain.

[I wouldn't be able to identify one JIT from another by looking at data]

ltratt:
Well, you are the reviewer, so you get to check this stuff ;)

vext01:
OK, I'll check it.

BTW:

`ltratt wants to merge 1 commit into faster_builds from dacapo_graal`

We want to merge to master, right?

vext01:
The JIT is enabled.

```
(Pdb++) stderr
'java.vm.version=25.66-b00-internal-jvmci-0.9-dev\nEnableJVMCI=true\nUseJVMCICompiler=true\n'
```

vext01:
I'm OK to merge, but I think the base branch is wrong?

vext01 (approved):

PR #186, opened 2016-10-09
Remove -i and PDF_FILENAME.

### LINKED ISSUE #185: Graph plotter uses a generic name and wipes over old files
At the moment, running `plot_krun_results` always creates a filed called `krun_results.pdf`, happily overwriting what was there before. I think we need to choose one or the other: either use a less generic name or prevent overwriting (both is overkill IMHO). My preference would be for the former, so that `plot_krun_results x.y.json.bz2` would create a file `x.y.pdf`.

snim2:
This PR changes the logic of some CLI options.

If the user passes in `-o OUTFILE` results are written to the given PDF file. If not, charts are displayed interactively and not written to disk. This means we no longer need the `-i` option.

Fixes #185


ltratt:
Works for me!

ltratt (approved):

PR #192, opened 2016-10-10
Speed up builds by having make run multiple simultaneous jobs.

ltratt:
Due to a misunderstanding on my part, we merged the original version of this into the wrong branch, and then deleted that branch. So the changes never made it to master. This PR simply puts back what was supposed to be there. Mea culpa.


snim2 (approved):
Looks OK to me!

PR #193, opened 2016-10-10
Consistent diff names

ltratt:
Initially this was supposed to just be *.patch -> *.diff. But, of those two files, one of them turned out to be no longer used, so I've deleted it.

I've tested a full build with this PR, and there were no issues.


snim2 (approved):
LGTM

PR #194, opened 2016-10-14
Add targets

### LINKED ISSUE #178: Add Makefile target to process results and plot charts

snim2:
Fixes #178 

Needs squashing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ all: build-benchmarks build-startup
 
 .PHONY: build-vms build-benchmarks build-krun build-startup bench clean
 .PHONY: plot-warmup-results plot-warmup-outliers-by-threshold
+.PHONY: plot-dacapo-results plot-octane-results
 .PHONY: clean-benchmarks clean-krun clean-plots
 
 build-vms:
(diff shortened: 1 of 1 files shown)


ltratt:
I'm not sure we want to hardcode the window size in here, at least not at this point. One way of doing things is to say something like `WINDOW_SIZE ?= 200` and then the user can override that with `make WINDOW_SIZE=100`.

snim2:
Ready to squash?

ltratt:
There are still hard-coded references to "200" left in the Makefile, which should depend on `WINDOW_SIZE` (I think).

Makefile
```diff
@@ -1,6 +1,10 @@
 PYTHON ?= python2.7
 PWD != pwd
 UNAME != uname
+WINDOW_SIZE ?= 200
+OUTLIER_THRESHOLD ?= 8
+PLOTS_NO_CPTS = plots_w200.pdf
```
vext01: `PLOTS_NO_CPTS = plots_w${WINDOW_SIZE}.pdf`?

vext01 (changes requested):

Makefile
vext01: And similar on the next line too.

snim2:
OK, I think that is fixed.

ltratt:
Yep, can be squashed, then I can merge.

snim2:
Splat.

vext01 (approved):

snim2:
Thanks!

PR #198, opened 2016-10-14
Recursively clone Krun.

vext01:
OK?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -83,7 +83,7 @@ PATCH_DIR=`pwd`/patches/
 build_initial_krun() {
     echo "\n===> Download and build krun\n"
     if ! [ -d "${HERE}/krun" ]; then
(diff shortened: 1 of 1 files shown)


snim2:
Is there any chance that someone might run `./build.sh` in an old clone of this repo? (i.e. should we say `git submodule update --init --recursive`)

ltratt:
No, they should be wiping old repos and rerunning. [Until last week, `make clean` didn't work properly.]

vext01:
The clone is guarded like this:

```
     if ! [ -d "${HERE}/krun" ]; then
```

For the final version we will clone a tag, so it won't matter. Right?

snim2:
OK, sounds good to me.

snim2 (approved):

PR #199, opened 2016-10-17
Update startup

### LINKED ISSUE #181: The startup scripts need to output data in the new dict format
At the moment (after the merger of #180) I get this exception:

```
make[1]: Leaving directory '/home/ltratt/warmup_experiment/benchmarks'
python2.7 krun/krun.py startup.krun
[2016-10-05 17:36:24: INFO] Wait 60 secs to allow system to cool prior to collecting initial temperature readings
  File "krun/krun.py", line 346, in <module>
    main(parser)
  File "krun/krun.py", line 209, in main
    inner_main(mailer, config, args)
  File "krun/krun.py", line 295, in inner_main
    sanity_checks(config, platform)
  File "krun/krun.py", line 68, in sanity_checks
    vm_info["vm_def"].sanity_checks()
  File "/home/ltratt/warmup_experiment/krun/krun/vm_defs.py", line 633, in sanity_checks
    self._check_truffle_enabled()
  File "/home/ltratt/warmup_experiment/krun/krun/vm_defs.py", line 629, in _check_truffle_enabled
    "jruby_check_graal_enabled.rb", force_dir=VM_SANITY_CHECKS_DIR)
  File "/home/ltratt/warmup_experiment/krun/krun/util.py", line 250, in spawn_sanity_check
    _ = check_and_parse_execution_results(stdout, stderr, rc)
  File "/home/ltratt/warmup_experiment/krun/krun/util.py", line 216, in check_and_parse_execution_results
    key_set = set(json_data.keys())
AttributeError: 'list' object has no attribute 'keys'
Makefile:58: recipe for target 'bench-startup-no-reboots' failed
make: *** [bench-startup-no-reboots] Error 1
bencher3%
```

snim2:
Fixes #181

### DIFF
--- a/startup.krun
+++ b/startup.krun
@@ -42,6 +42,7 @@ def patch_runner(vm_def):
         assert "iterations_runner." in vm_def.iterations_runner
         _, suffix = vm_def.iterations_runner.rsplit(".", 1)
         vm_def.iterations_runner = os.path.join(STARTUP_RUNNER_DIR, "startup_runner." + suffix)
+    vm_def.instrument = False
     return vm_def
 
 
--- a/startup_runners/outer_startup_runner.c
+++ b/startup_runners/outer_startup_runner.c
@@ -6,6 +6,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <errno.h>
 #include <stdint.h>
 #include <unistd.h>
@@ -19,18 +20,37 @@ main(int argc, char **argv)
 {
     double start_time = -1;
     int result;
+    char fake_datum[] = "[0, 0], ";
+    char end_datum[] = "[0, 0] ";
+    int datum_size = (int)strlen(fake_datum);
+    krun_init();
+    char* fake_data = malloc(1);
+
+    if (krun_get_num_cores() > 1) {
+        memcpy(fake_data, fake_datum, datum_size);
+    }
+    if (krun_get_num_cores() > 2) {
+        for(int i=0; i < krun_get_num_cores() - 2; i++) {
+            memcpy(fake_data + datum_size * (i + 1), fake_datum, datum_size);
+        }
+    }
+    memcpy(fake_data + datum_size * (krun_get_num_cores() - 1), end_datum, datum_size - 1);
+    fake_data[datum_size * krun_get_num_cores() - 1] = '\0';
 
     /* silence gcc */
     argc = argc;
     argv = argv;
 
-    krun_measure(1);
-    start_time = krun_get_wallclock(1);
-    fprintf(stdout, "[[%f,\n", start_time);
+    start_time = krun_clock_gettime_monotonic();
+    fprintf(stdout, "{ \"core_cycle_counts\": [ %s ],", fake_data);
+    fprintf(stdout, " \"mperf_counts\" : [ %s ],", fake_data);
+    fprintf(stdout, " \"aperf_counts\" : [ %s ],", fake_data);
+    fprintf(stdout, " \"wallclock_times\" : [ %f, ", start_time);
     fflush(stdout);
     result = execv(argv[1], argv + 1);
     if (result) {
         perror("Starting subprocess failed:");
     }
+    free(fake_data);
     return (EXIT_SUCCESS);
 }
--- a/startup_runners/startup_runner.js
+++ b/startup_runners/startup_runner.js
@@ -1,3 +1,2 @@
-// NOTE: you need to provide clock_gettime_monotonic.
-print(clock_gettime_monotonic());
-print("], [-1.0, -1.0]]");
+print(krun_clock_gettime_monotonic());
+print("] }");
--- a/startup_runners/startup_runner.php
+++ b/startup_runners/startup_runner.php
@@ -1,10 +1,6 @@
 <?php
 
-/*
- * Note! You will need to provide an implementation of clock_gettime_monotonic()
- */
-
-echo clock_gettime_monotonic();
-echo "], [-1.0, -1.0]]";
+echo krun_clock_gettime_monotonic();
+echo "] }";
 
 ?>
--- a/startup_runners/startup_runner.rb
+++ b/startup_runners/startup_runner.rb
@@ -1,4 +1,4 @@
-# defined this way so we don't measure the conditional platform check.
+# Defined this way so we don't measure the conditional platform check.
 if /linux/ =~ RUBY_PLATFORM then
     def clock_gettime_monotonic()
         Process.clock_gettime(Process::CLOCK_MONOTONIC_RAW)
@@ -12,5 +12,5 @@ end
 if __FILE__ == $0
     start_time = clock_gettime_monotonic()
     STDOUT.write String(start_time)
-    STDOUT.write "], [-1.0, -1.0]]"
+    STDOUT.write "] }"
 end
--- a/patches/hhvm.diff
+++ b/patches/hhvm.diff
@@ -23,7 +23,7 @@ index 1d0822d..ad3e69b 100644
  # See https://github.com/facebook/hhvm/issues/5217
  INCLUDE(CheckCXXSourceCompiles)
 diff --git a/hphp/runtime/ext/std/ext_std_misc.cpp b/hphp/runtime/ext/std/ext_std_misc.cpp
-index 8b08cec..3347fbf 100644
+index 8b08cec..dcfc57b 100644
 --- a/hphp/runtime/ext/std/ext_std_misc.cpp
 +++ b/hphp/runtime/ext/std/ext_std_misc.cpp
 @@ -44,10 +44,14 @@
@@ -41,13 +41,14 @@ index 8b08cec..3347fbf 100644
  namespace HPHP {
  
  IMPLEMENT_THREAD_LOCAL(std::string, s_misc_highlight_default_string);
-@@ -192,6 +196,14 @@ void StandardExtension::initMisc() {
+@@ -192,6 +196,15 @@ void StandardExtension::initMisc() {
      HHVM_FE(sys_getloadavg);
      HHVM_FE(token_get_all);
      HHVM_FE(token_name);
 +    HHVM_FE(krun_init);
 +    HHVM_FE(krun_done);
 +    HHVM_FE(krun_measure);
++    HHVM_FE(krun_clock_gettime_monotonic);
 +    HHVM_FE(krun_get_num_cores);
 +    HHVM_FE(krun_get_wallclock);
 +    HHVM_FE(krun_get_core_cycles_double);
@@ -56,7 +57,7 @@ index 8b08cec..3347fbf 100644
      HHVM_FE(hphp_to_string);
      HHVM_FALIAS(__SystemLib\\max2, SystemLib_max2);
      HHVM_FALIAS(__SystemLib\\min2, SystemLib_min2);
-@@ -966,5 +978,37 @@ Variant HHVM_FUNCTION(SystemLib_min2, const Variant& value1,
+@@ -966,5 +979,41 @@ Variant HHVM_FUNCTION(SystemLib_min2, const Variant& value1,
  #undef YYTOKEN_MAP
  #undef YYTOKEN
  
@@ -70,6 +71,10 @@ index 8b08cec..3347fbf 100644
 +
 +void HHVM_FUNCTION(krun_measure, int mdata_index) {
 +    krun_measure(mdata_index);
++}
++
++double HHVM_FUNCTION(krun_clock_gettime_monotonic) {
++    return krun_clock_gettime_monotonic();
 +}
 +
 +int HHVM_FUNCTION(krun_get_num_cores) {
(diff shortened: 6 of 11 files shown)


startup_runners/startup_runner.rb
```diff
@@ -1,3 +1,5 @@
+# NOTE: you need to provide krun_measure() and krun_get_wallclock().
+
 # defined this way so we don't measure the conditional platform check.
 if /linux/ =~ RUBY_PLATFORM then
     def clock_gettime_monotonic()
```
vext01: I think this might be dead?

vext01:
Both the inner and outer runners are using `krun_measure(1)`. This means we end up doing something like this:

```
        data->wallclock = krun_clock_gettime_monotonic();                       
        for (int core = 0; core < krun_num_cores; core++) {                     
            data->core_cycles[core] = krun_read_core_cycles(core);              
            data->aperf[core] = krun_read_aperf(core);                          
            data->mperf[core] = krun_read_mperf(core);                          
        }                                                                       
        krun_check_mdata();

        // VM starts up here

        data->wallclock = krun_clock_gettime_monotonic();                       
        for (int core = 0; core < krun_num_cores; core++) {                     
            data->core_cycles[core] = krun_read_core_cycles(core);              
            data->aperf[core] = krun_read_aperf(core);                          
            data->mperf[core] = krun_read_mperf(core);                          
        }                                                                       
        krun_check_mdata();
```

In other words, we end up measuring taking un-needed msr readings on top of startup. We also measure the integrity check.

At the very least, the outer runner should use index 0, but I wonder if it would be better to expose `krun_clock_gettime_monotonic` and use it directly?

startup.krun
```diff
@@ -42,6 +42,7 @@ def patch_runner(vm_def):
         assert "iterations_runner." in vm_def.iterations_runner
         _, suffix = vm_def.iterations_runner.rsplit(".", 1)
         vm_def.iterations_runner = os.path.join(STARTUP_RUNNER_DIR, "startup_runner." + suffix)
+    vm_def.instrument = False
```
ltratt, line 45: I'm not totally sure if this should be in this PR. I'm happy to be convinced though!

snim2:
We are taking timing readings twice. Once in the outer running and once in the inner runner. Presumably we want to turn measurements off in between, so how do we do that?

startup_runners/startup_runner.js
```diff
@@ -1,3 +1,4 @@
-// NOTE: you need to provide clock_gettime_monotonic.
-print(clock_gettime_monotonic());
-print("], [-1.0, -1.0]]");
+// NOTE: you need to provide krun_measure() and krun_get_wallclock()..
```
ltratt: Let's delete this comment (the old one wasn't very useful either).

startup_runners/startup_runner.php
```diff
@@ -1,10 +1,11 @@
 <?php
 
 /*
- * Note! You will need to provide an implementation of clock_gettime_monotonic()
+ * Note! You will need to provide an implementation of krun_measure() and krun_get_wallclock()
```
ltratt: Similarly, let's kill this comment.

startup_runners/startup_runner.rb
```diff
@@ -1,3 +1,5 @@
+# NOTE: you need to provide krun_measure() and krun_get_wallclock().
```
ltratt: Similarly, let's kill this comment.

ltratt:
Best to ask @cfbolz about the methodology.

startup.krun
snim2: If PyPy instrumentation is on in `warmup.krun` and we don't turn it off in `startup.krun` then I get a parse error of sorts. So, either we need to decide that we keep this in and all instrumentation is off in the startup runner, or we remove this line and fix the startup runners to work with all the instrumentation we have.

ltratt: We can consider startup and instrumentation modes to be mutually exclusive. Maybe this should be an `assert` or something? [I'm not saying I have the answer here... just trying to work out what the right thing is!]

ltratt: We can consider startup and instrumentation modes to be mutually exclusive. Maybe this should be an `assert` or something? [I'm not saying I have the answer here... just trying to work out what the right thing is!]

snim2: I'm not sure what we want here either. Maybe @vext01 has a view?

vext01:
Like I say, we only care about wallclock time, so maybe we should call `krun_clock_gettime_monotonic()` directly and thus skip the other measurements.

snim2:
OK, I will push a commit which does that...

snim2:
OK, what I have now works for every language except Java, and every VM except HHVM and V8. In all three cases the VMs need to be patched to allow access to `krun_clock_gettime_monotonic()`

ltratt:
If Edd's happy with the current approach (it looks sensible to me, but this isn't my parish), then I'd suggest going ahead on that basis.

vext01:
This looks fine.

I guess do Java next.

Do you need help patching VMs, or shall I do that bit? It can be fiddly.

snim2:
I'm not really sure where to start with patching the VMs, happy to have a go if you give me some step by step instructions...

vext01:
In the interest of getting it done quickly, shall I go ahead and do it?

snim2:
That would be great if you can!

vext01:
sure, no worries. It will take a while, as building hhvm takes an age and a day.

ltratt:
If you're using a vaguely recent master, HHVM now builds with `num_cores`, which speeds it up a _lot_.

vext01:
@ltratt ah yeah.

@snim2 before I go ahead, did you look at the output of a startup run to check it looks saneish?

snim2:
Before we moved to `krun_clock_gettime_monotonic()` I checked that we could produce charts form the output file:

startup_chart

The chart itself isn't very useful, but clearly the output file works with our other scripts.

Should we raise a new issue to write a separate script to make nice plots from the startup results?

ltratt:
We don't, I think, want plots from the startup scripts: we want simple tables with means and confidence intervals. Our working assumption is that they're normally distributed.

vext01:
What is the y-axis on that graph? The clock is measured in seconds. Those values look like garbage, no?

Re plots, what @ltratt said.

vext01:
Probably because krun_get_wallclock returns an interval based on two internal measurements, but we only take one measurement in each inner/outer runner.

This should go away after we use krun_clock_gettime_monotonic.

snim2:
Hmm, this still seems a little odd:

``` python
>>> from warmup.krun_results import read_krun_results_file
>>> data = read_krun_results_file("startup_results.json.bz2")
>>> data["wallclock_times"]["binarytrees:JRubyTruffle:default-ruby"]
[[4559160.027431, 4559163.433388]]
>>> 
```

What units does `krun_clock_gettime_monotonic()` return? Seconds?

vext01:
Yes, but you were not using krun_clock_gettime_monotonic() directly above, right?

vext01:
And that's seconds since boot. We should be computing an interval at some point.

snim2:
The chart was old data, the Python session was new data with `clock_gettime_monotonic()`. Where should an interval be computed?

vext01:
I'm not sure. Let me finish patching the VMs and I'll have a look for you.

vext01:
> OK, what I have now works for every language except Java, and every VM except HHVM and V8. 

Did you mean JRuby btw? We shouldn't need to patch the JVM. Just update the JNI bindings.

snim2:
No, I meant that `clock_gettime_monotonic` needs to be added to the JNI bindings.

vext01:
Oh that's right, JRuby already provides access to the clock.

vext01:
The JNI bindings are easyish btw.

startup_runners/outer_startup_runner.c
```diff
@@ -24,9 +24,11 @@ main(int argc, char **argv)
     argc = argc;
     argv = argv;
 
-    krun_measure(1);
-    start_time = krun_get_wallclock(1);
-    fprintf(stdout, "[[%f,\n", start_time);
+    start_time = krun_clock_gettime_monotonic();
+    fprintf(stdout, "{ \"core_cycle_counts\": [[0, 0], [0, 0], [0, 0], [0, 0]],");
+    fprintf(stdout, "  \"mperf_counts\" : [[0, 0], [0, 0], [0, 0], [0, 0]],");
+    fprintf(stdout, " \"aperf_counts\" : [[0, 0], [0, 0], [0, 0], [0, 0]],");
```
snim2: Is it worth refactoring this so that we don't assume the CPU has four cores?

vext01: You can use krun_get_num_cores(), which is exposed in all VMs.

vext01:
VMs patched.

hhvm:

```
vext01@bencher5:~/research/warmup_experiment$ ./work/hhvm/hphp/hhvm/hhvm -a     
Welcome to HipHop Debugger!                                                     
Type "help" or "?" for a complete list of commands.                             

Note: no server specified, debugging local scripts only.                        
If you want to connect to a server, launch with "-h" or use:                    
  [m]achine [c]onnect <servername>                                              

hphpd> echo krun_clock_gettime_monotonic();                                     
4502111.3383156
```

v8:                                                                                

```
vext01@bencher5:~/research/warmup_experiment$ ./work/v8/out/native/d8                                          
V8 version 5.1.281.65                                                           

d8> krun_clock_gettime_monotonic;                                               
function krun_clock_gettime_monotonic() { [native code] }                       
d8> print(krun_clock_gettime_monotonic());                                      
4502287.607179753                                                               
```

snim2:
Thanks, will fix the fake data in the outer loop and test.

cfbolz:
It makes sense to me to try to use the timer only directly, otherwise you always measure the time it takes to read the MSR and other stuff.

About the output file: It is not meant to be directly compatible with krun output files. After all, it just produces a single measurement (not 2000) per process iteration. The reason why you are seeing two numbers like this: `[4559160.027431, 4559163.433388]` is because it simply prints the "just before" and "just inside" times of the process start. So what you have to do is take the difference to get the time it starts up.

snim2:
This is just waiting on me to test it out on bencher2, then I'm expecting we can merge later today

vext01:
Thanks for clarifying @cfbolz

snim2:
@ltratt I have done a full run on bencher2 and everything seems fine, so this is ready for review.

startup_runners/outer_startup_runner.c
```diff
@@ -25,7 +25,9 @@ main(int argc, char **argv)
     argv = argv;
 
     start_time = krun_clock_gettime_monotonic();
-    fprintf(stdout, "{ \"core_cycle_counts\": [], \"mperf_counts\" : [], \"aperf_counts\" : [],");
+    fprintf(stdout, "{ \"core_cycle_counts\": [[0, 0], [0, 0], [0, 0], [0, 0]],");
+    fprintf(stdout, "  \"mperf_counts\" : [[0, 0], [0, 0], [0, 0], [0, 0]],");
```
ltratt: Two spaces before the `\=`?

snim2: I think to make sure there is a space after the commas in the output.

startup_runners/outer_startup_runner.c
```diff
@@ -6,28 +6,45 @@
...
 
 #define BENCH_FUNC_NAME "run_iter"
+#define LEN_DATUM 8
 
 int
 main(int argc, char **argv)
 {
     double start_time = -1;
     int result;
+    char fake_datum[LEN_DATUM] = "[0, 0], ";
```
ltratt: I think there's a buffer overrun here: `LEN_DATUM` is length 8, but the RHS string is of length 9 (with the null terminator). I think this probably impacts other parts of the code too.

startup_runners/outer_startup_runner.c
```diff
@@ -6,28 +6,45 @@
...
     /* silence gcc */
     argc = argc;
     argv = argv;
 
     start_time = krun_clock_gettime_monotonic();
-    fprintf(stdout, "{ \"core_cycle_counts\": [[0, 0], [0, 0], [0, 0], [0, 0]],");
-    fprintf(stdout, "  \"mperf_counts\" : [[0, 0], [0, 0], [0, 0], [0, 0]],");
-    fprintf(stdout, " \"aperf_counts\" : [[0, 0], [0, 0], [0, 0], [0, 0]],");
+    fprintf(stdout, "{ \"core_cycle_counts\": [ %s ],", fake_data);
+    fprintf(stdout, "  \"mperf_counts\" : [ %s ],", fake_data);
```
ltratt: Double space before `\"`.

ltratt:
OK, various minor comments.

snim2:
Ready to squash?

ltratt:
Let's rebase against master, then I can review. Don't squash though please.

ltratt:
Squash as you see fit, then I can merge.

ltratt (approved):

snim2:
Splat.

PR #203, opened 2016-10-18
Fix issues with changepoints

snim2:
/cc @rkillick

### DIFF
--- a/bin/mark_changepoints_in_json
+++ b/bin/mark_changepoints_in_json
@@ -4,7 +4,6 @@ Write changepoint and classification information into JSON files.
 """
 
 import argparse
-import copy
 import os
 import os.path
 import rpy2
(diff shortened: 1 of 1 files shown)


snim2:
On Ubuntu, using CRAN to install `r-base` and `r-base-dev`: https://www.datascienceriot.com/how-to-install-r-in-linux-ubuntu-16-04-xenial-xerus/

ltratt:
Obvious dumb question: why prefer deepcopy? After all, it's much more expensive! Is there a semantic reason why it's necessary?

snim2:
This was what we were playing with at the last thrash. It doesn't seem to make any difference to the results, but you thought it might be the source of a bug. Before this branch is ready for review I'll check this again and revert if we can.

ltratt:
The suggestion was simply because copy.copy on a list is equivalent to the more normal `[:]`, so it did seem a possible thing to look at. Looking at it again once everything's fixed sounds sensible to me.

snim2:
Squashed and ready for review

ltratt (approved):

PR #204, opened 2016-10-19
Bring the network down while benchmarking

vext01:
This implements Laurie's idea to bring the network down while benchmarking.

Tested with real reboots on VMs, including hitting ctrl+c in the middle of a benchmark (post-hooks run just fine!).

OK?

### DIFF
--- a/warmup.krun
+++ b/warmup.krun
@@ -141,9 +141,14 @@ if sys.platform.startswith("linux"):
         "sudo service cron stop",
         "sudo service atd stop",
         "sudo service postfix stop",
+        "sudo service networking stop",
     ]
 
+    PING_HOST = "bencher2.soft-dev.org"
     POST_EXECUTION_CMDS = [
+        # The network doesn't always come up properly on debian. We keep trying
+        # until we can ping a host on the network.
+        "while true; do ping -c 3 -q %s && break; sudo service networking stop; sleep 5; sudo service networking start; done" % PING_HOST,
         "sudo service cron start || true",
         "sudo service atd start || true",
         "sudo service postfix start || true",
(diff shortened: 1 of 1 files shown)


warmup.krun
```diff
@@ -141,9 +141,14 @@ if sys.platform.startswith("linux"):
         "sudo service cron stop",
         "sudo service atd stop",
         "sudo service postfix stop",
+        "sudo service networking stop",
     ]
 
+    PING_HOST = "192.168.1.245"
```
snim2: Is it wise to have IP addresses hard-coded like this? I'm just thinking that if someone else tries to clone and run this repo....

vext01: Oops! That was not supposed to be committed.

warmup.krun
```diff
@@ -166,12 +173,20 @@ else:
...
+# directory, and that the corresponding public key is installed on the correct
+# user on the remote machine.
+#
 # We allow failure, otherwise Krun will halt the experiment if (e.g.) the
 # SSH server hostname is temporarily unavailable.
 HOSTNAME = platform.node().split(".")[0]
-SCP_CMD = ("tar czf - ${KRUN_RESULTS_FILE} ${KRUN_LOG_FILE} | "
-           "ssh -i id_rsa " "vext01@bencher2.soft-dev.org "
-           "'cat > research/krun_results/%s.tgz'" % HOSTNAME)
+REMOTE_LOGIN = "vext01@bencher2.soft-dev.org"
```
snim2, line 183: Similarly, hard-coded soft-dev specifics here.

vext01: I think it's ok to have soft-dev hostnames in the warmup experiment. This is not new in this commit, I just moved it into a variable.

snim2: OK

ltratt: Sarah's right, it is a bit weird to have these things in here... or, at least, to have them lurking a bit randomly later in the file. Maybe we can bring all the "softdev specific variables" to the top of the file, so that other people stand a chance of running this on their machines?

vext01: In a sense all of the hooks are soft-dev specific. Do you really want to move them all to the top? Perhaps a better option is to put our local stuff in another module, and import it as/where it feels right?

ltratt: Hmm, for now, let's keep things as-is. At some point we can maybe factor things out a little bit, but it's not critical.

vext01: Agreed. It's certainly a separate PR

vext01:
Can I squash?

ltratt:
Yep, I think we're all agreed this is ready for squashing.

vext01:
splat

ltratt (approved):

PR #205, opened 2016-10-19
Default to building one-job-at-a-time.

ltratt:
One problem with parallel jobs is that we then end up compiling VMs non-deterministically, which may or may not cause unpredictable builds. If "-f" is specified, we do fast builds, otherwise we do one-job-at-a-time "probably safer" builds.

The idea is that, during testing, it's fine to use "-f" but for the real experiment we use the default (slower) setting.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -37,6 +37,36 @@ case `uname` in
         check_for gm4;;
 esac
 
+
+usage() {
+    echo "Usage: build.sh [-f]" >&2
+    exit 1
+}
+
+# How many cores does this machine have (so we can pass a sensible number to
+# gmake -j)?
+case `uname` in
+    OpenBSD) max_jobs=`sysctl -n hw.ncpu`;;
+    Linux) max_jobs=`nproc`;;
+esac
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -37,6 +37,36 @@ case `uname` in
...
+
+usage() {
+    echo "Usage: build.sh [-f]" >&2
+    exit 1
+}
+
+# How many cores does this machine have (so we can pass a sensible number to
+# gmake -j)?
+case `uname` in
+    OpenBSD) max_jobs=`sysctl hw.ncpu|cut -d "=" -f 2`;;
```
vext01: Use `-n` instead of cut?

vext01:
Looks ok.

build.sh
ltratt: I didn't know you could do that. Good idea!

ltratt:
Ready for rereview.

vext01:
Squash and I can merge.

ltratt:
Done.

vext01 (approved):

PR #212, opened 2016-10-28
Obtain instr_data from files.

### LINKED ISSUE #163: Add  GC time and compilation time to plots.
As discussed before, we need to add two more lines to our graphs, one for GC time, one for compilation time.

If we add these on top of the graph we have, I suspect the graph will be too busy.

One idea is to use a vertical stack of 5 axes, one for each data point, a bit like this:

subplots_demo_03

Each of the axes is a subplot created with `sharex=true`.

The problem is, one figure can have only one set of subplots. Ideally we would need a way to put many figures on one page. If we can't find a way to do that in matplotlib itself, we could generate a PDF externally using tex, or even make a HTML page. This would however break the interactive mode the script currently has.

I'm open to new ideas too. It seems multi-dimensional data visualisation is hard.

### LINKED ISSUE #201: Bring scripts up to date with separate instr_data files
Related to Krun PR 268.

`instr_data` will soon be removed from results files. In the new scheme, if you have a config file named `test.krun`, you get a directory called `test_instr_data` with files prefixed with the job key (with `:` replaced with `__`) and then a process exec number suffix + `.json.bz2`.

For example:

```
$ python -mpdb ./krun/krun.py test.krun --quick --debug debug
...
[2016-10-17 11:40:58 INFO] Done: Results dumped to test_results.json.bz2
[2016-10-17 11:40:58 INFO] Session completed. Log file at: 'test_20161017_114049.log'
Don't forget to disable Krun at boot.
$ ls test_
test_20161017_114049.log  test_instr_data/          test_results.json.bz2     
vext01@bencher5:~/research/warmup_experiment$ ls test_instr_data/
empty_PyPy_default-python__0.json.bz2  empty_PyPy_default-python__5.json.bz2
empty_PyPy_default-python__1.json.bz2  empty_PyPy_default-python__6.json.bz2
empty_PyPy_default-python__2.json.bz2  empty_PyPy_default-python__7.json.bz2
empty_PyPy_default-python__3.json.bz2  empty_PyPy_default-python__8.json.bz2
empty_PyPy_default-python__4.json.bz2  empty_PyPy_default-python__9.json.bz2
```

snim2:
This PR adds all our available instrumentation data to the plots, in the new file format.

The PyPy instrumentation parser is based on [this gist](https://gist.github.com/vext01/3b685296d938cd78363dc6cd4e22754d).

Please check very carefully (particularly units of measurement, these are definitely wrong for PyPy, but I'm not sure what they should be).

**Test case**: [test.pdf](https://github.com/softdevteam/warmup_experiment/files/559088/test.pdf)

Fixes #163
Fixes #201


bin/plot_krun_results
```diff
@@ -687,15 +740,32 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
...
                         data_dictionary['aperf_counts'][key][machine] = None
                         data_dictionary['mperf_counts'][key][machine] = None
+                        if instr_data:
+                            data_dictionary['instr_data'][key][machine] = None
                     else:
                         data_dictionary['cycles_counts'][key][machine] = data['core_cycle_counts'][key]
                         data_dictionary['aperf_counts'][key][machine] = data['aperf_counts'][key]
                         data_dictionary['mperf_counts'][key][machine] = data['mperf_counts'][key]
+                        if instr_data:
+                            _, vm, _ = key.split(':')
```
ltratt: Might be clearer as `vm = key.split(':')[1]`?

ltratt:
I like this!

I think that some of the PyPy numbers may have semi-meaningless units (maybe some sort of cycle count? I'm not sure). @cfbolz will probably know.

OK, so now some other comments:
- We end up with some fun things like a y axis which is all "0.0000" (p4). I suspect in these extreme case, we should just do the "add a scale to the y axis" thing (even though I don't generally like it).
- I suspect we should normalise the y axes of all the data across process executions. e.g. look at p8. In terms of the overall execution time, it's immediately clear that proc #1 is much slower. But when you look at the GC time, it takes a lot more mental effort to work out what's going on.
- One interesting thing is that it's very hard to relate the GC spikes to anything in the execution time. Maybe not for this PR, but I wonder if we should say something like "if there's a GC spike (using the same sort of outlier stuff we do now), plot a red blob on the execution time" so that we can say "ah, the GC spike is highly likely to be the cause of that slow iteration". [Same for JIT time as well -- you can see an interesting example on p14 where JIT compilation happens part of the way through the benchmark, causing spikes.]

[As an aside, bencher2 is excellent fun for this, as you can see the effects of turbo boost. It shows that the core measurement stuff appears to be working fine!]

cfbolz:
yes, the pypy numbers are have arbitrary units (if it becomes important, I can probably investigate what they really mean, but a brief look didn't tell me).

snim2:
Many thanks @cfbolz

bin/plot_krun_results
```diff
@@ -672,6 +724,17 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
         else:
             machine_name = machine
 
+        # Is there instrumentation data with this results file?
+        instr_data = False
+        instr_dir = os.path.dirname(filename) + filename.split('_')[0] + '_instr_data'
```
vext01: Isn't the field separator in these filenames `__`?

vext01: (double underscore)

snim2: I think it's in the call to `split()`

vext01: Ah, it's the dir name. You would be better off stripping off the ending `.json.bz2` part in case the config filename contained underscores.

vext01:
Other than that one small item, nothing jumps out as wrong here.

snim2:
Next test case: [test.pdf](https://github.com/softdevteam/warmup_experiment/files/562413/test.pdf)

Hopefully answers all the comments above.

bin/plot_krun_results
```diff
@@ -729,7 +729,8 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
 
         # Is there instrumentation data with this results file?
         instr_data = False
-        instr_dir = os.path.dirname(filename) + filename.split('_')[0] + '_instr_data'
+        filename_root = filename[:-17]  # Remove _results.json.bz2
```
vext01: This is fine, but if i were nitpicking:

```
filename[:len("_results.json.bz2")]
```

vext01: minus len, that is.

ltratt:
@cfbolz We really could do with having some vague idea of what the units are, because of the problem I'm just going to point out below.

@snim2 This looks nice, but one thing looks like it might a _bit_ wrong: have a look at the y axis for JIT times on p4 for example. 3.978×10^7 seconds is probably not correct, unless that benchmark took a _really_ long time to run ;) I think we should merge this PR now, but remove the RHS label (and, maybe, LHS y axis numbers) from those graphs until we know what the units are. At the moment, the numbers are probably more confusing than anything else.

snim2:
@ltratt do we know what units the Hotspot GC events data is measured in?

ltratt:
I have no idea. Edd might know.

snim2:
OK, let's just remove them all for now. 

Ready to squash?

ltratt:
Fine with me. Let's get @vext01's OK before doing so. Once we've got that, squash, and I'll merge with pleasure!

warmup/vm_instruments.py
```diff
@@ -62,10 +71,55 @@ def parse_instr_data(self):
             for index in xrange(1, len(cumulative_times)):
                 times_secs.append((cumulative_times[index] - cumulative_times[index - 1]) / 1000.0)
             assert len(times_secs) == len(cumulative_times)
-            self.chart_data.append([ChartData('GC (secs)', times_secs, 'GC events')])
+            self.chart_data.append([ChartData('GC', times_secs, 'GC events')])
+
+
+class PyPyInstrumentParser(VMInstrumentParser):
+    """Parser for PyPy instrumentation data.
```
vext01: Join lines?

snim2:
New test case: 
[test.pdf](https://github.com/softdevteam/warmup_experiment/files/563914/test.pdf)

I have put the "seconds" back in the Hotspot charts because these are described in the [documentation for our iterations runner](https://github.com/softdevteam/krun/blob/master/iterations_runners/iterations_runner.java) (so presumably came from the JVM docs).

N-Body Hotspot looks a bit strange to me, but we only have 200 iterations here.

vext01:
This is so insightful! I love it!

fannkuch:pypy we can see the steps are GC steps! Oddly, there are two steps on other cores at that time too. But why? PyPy is single threaded? Or does it release the GIL and do something in C/the kernel at some point in GC?

Similarly for fasta/pypy.

Why is N-Body Hotspot strange? Because there was no GC? Maybe it will GC later.

Spikes on richards/pypy are clearly GC.

Spectralnorm/hotspot is the JVMs fault alone.

Will now review the code.

warmup/vm_instruments.py
```diff
@@ -66,12 +66,24 @@ def parse_instr_data(self):
...
-            for index in xrange(1, len(cumulative_times)):
-                times_secs.append((cumulative_times[index] - cumulative_times[index - 1]) / 1000.0)
-            assert len(times_secs) == len(cumulative_times)
-            self.chart_data.append([ChartData('GC', times_secs, 'GC events')])
+            jit_cumulative_times = [self.instr_data[p_exec][i][1] for i in xrange(iterations)]
+            gc_cumulative_times = list()
+            # Sum GC times over all collectors that ran in each iteration.
+            for i in xrange(iterations):
+                gc_iteration = 0
+                for collectors in self.instr_data[p_exec][i][2]:
```
vext01: de-pluralise this name? Each iteration works on one collector, right?

warmup/vm_instruments.py
```diff
@@ -66,12 +66,24 @@ def parse_instr_data(self):
...
+                    gc_iteration += collectors[-1]
+                gc_cumulative_times.append(gc_iteration)
+            jit_times_secs = [jit_cumulative_times[0] / 1000.0]
```
vext01: This part is computing the difference between the cumulative counts, right?

Would it be clearer to have two variables, `last_gc` and `last_jit`, which you subtract from the next values, before re-assigning the last values?

snim2: Lines 79--82 (inclusive) turn the cumulative times into per-iteration times, they also convert milliseconds into seconds. Lines 77 and 78 convert the first datum into milliseconds, because it isn't handled by the loops on lines 79--82.

snim2:
Hang about, that last commit is definitely not right

snim2:
Fixed.

ltratt:
Looks good to me. @vext01? If Edd's OK, then I think we can squash and merge.

vext01:
LGTM

ltratt:
OK, let's squash, then I can merge.

snim2:
@ltratt have squashed...

ltratt (approved):

PR #213, opened 2016-10-29
First stab at a better Octane runner.

ltratt:
_Not ready for merging until a full test run has completed in a few days time, but ready for review now_

The original Octane runner is written in a convoluted continuation-passing style, for no obvious gain, but at a substantial cost of readability. It also has a couple of features that we don't need. So, first, making things simpler, and removing those features makes our lives a lot better. The new runner also uses a microsecond monotonic timer rather than a millisecond non-monotonic timer, which should give us much better quality data.

[The most important code is probably the new runner in `octane_run_we.js`]

### DIFF
--- a/patches/octane_run_we.js
+++ b/patches/octane_run_we.js
@@ -1,0 +1,51 @@
+load('base.js');
+load('richards.js');
+load('deltablue.js');
+load('crypto.js');
+load('raytrace.js');
+load('earley-boyer.js');
+load('regexp.js');
+load('splay.js');
+load('navier-stokes.js');
+load('pdfjs.js');
+//load('mandreel.js');
+//load('gbemu-part1.js');
+//load('gbemu-part2.js');
+//load('code-load.js');
+//load('box2d.js');
+//load('zlib.js');
+//load('zlib-data.js');
+load('typescript.js');
+load('typescript-input.js');
+load('typescript-compiler.js');
+
+
+function main() {
+  var suites = BenchmarkSuite.suites;
+  for (var i = 0; i < suites.length; i++) {
+    var suite = suites[i];
+    var benchmarks = suite.benchmarks;
+    for (var j = 0; j < benchmarks.length; j++) {
+      var benchmark = benchmarks[j];
+      print(benchmark.name);
+      for (var k = 0; k < 2000; k++) {
+        BenchmarkSuite.ResetRNG();
+        benchmark.Setup();
+        var start = performance.now();
+        // Octane benchmarks consist of a (generally fast) "inner" iteration;
+        // each benchmark then says "running me for
+        // benchmark.deterministicIterations times makes for an outer
+        // iteration". We only care about the outer iterations.
+        for (var l = 0; l < benchmark.deterministicIterations; l++) {
+          benchmark.run();
+        }
+        var elapsed = performance.now() - start;
+        print('  ' + elapsed);
+        benchmark.TearDown();
+      }
+    }
+  }
+}
+
+
+main();
--- a/build.sh
+++ b/build.sh
@@ -596,6 +596,7 @@ fetch_octane() {
     cd octane
     git checkout ${OCTANE_V} || exit $?
     patch < ${PATCH_DIR}/octane.diff || exit $?
+    cp ${PATCH_DIR}/octane_run_we.js run_we.js || exit $?
 }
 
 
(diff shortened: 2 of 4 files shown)


ltratt:
This is now ready for review @vext01.

patches/octane_run_we.js
```diff
@@ -0,0 +1,47 @@
...
+load('richards.js');
+load('deltablue.js');
+load('crypto.js');
+load('raytrace.js');
+load('earley-boyer.js');
+load('regexp.js');
+load('splay.js');
+load('navier-stokes.js');
+load('pdfjs.js');
+//load('mandreel.js');
```
vext01, line 11: These are skipped on purpose? I suppose they crash, or are not useful?

patches/octane_run_we.js
```diff
@@ -0,0 +1,47 @@
...
+//load('zlib.js');
+//load('zlib-data.js');
+load('typescript.js');
+load('typescript-input.js');
+load('typescript-compiler.js');
+
+
+function main() {
+  var suites = BenchmarkSuite.suites;
+  for (var i = 0; i < suites.length; i++) {
```
vext01, line 25: I was about to suggest that if `i` is only used to fetch the right array index out, use a foreach, but it seems javascript doesn't have one. Or one that everyone can agree on, anyway...

patches/octane_run_we.js
```diff
@@ -0,0 +1,47 @@
...
+    var suite = suites[i];
+    var benchmarks = suite.benchmarks;
+    for (var j = 0; j < benchmarks.length; j++) {
+      var benchmark = benchmarks[j];
+      print(benchmark.name);
+      for (var k = 0; k < 2000; k++) {
```
vext01, line 31: Can you add comments saying what each loop does here? Presumably this innermost one is in-process iterations? Which would mean the next ourmost loop should be process executions, but it can't be, as we are in the same process still...

patches/octane_run_we.js
```diff
@@ -0,0 +1,47 @@
...
+        BenchmarkSuite.ResetRNG();
+        benchmark.Setup();
+        var start = performance.now();
+        for (var l = 0; l < benchmark.deterministicIterations; l++) {
```
vext01: There's another loop here. I'm confused.

vext01: Maybe one of these loops is an artificial loop to make the benchmark runtime longer?

vext01:
Why `octane_run_we.js` btw? Because "we" made it? Or it stands for something?

patches/octane_run_we.js
ltratt: See the paper for more details.

patches/octane_run_we.js
ltratt: Precisely :/

patches/octane_run_we.js
ltratt: [Summary: yes, they're skipped because they're broken, but it makes it easier to see explicitly which are skipped.]

patches/octane_run_we.js
ltratt: Given that it's 2000, I think it's blindingly clear in this case.

patches/octane_run_we.js
ltratt: This one I agree on.

ltratt:
"we" == "warmup experiment" (nothing deep :))

patches/octane_run_we.js
vext01: Which means the next outer loop should be for process executions, yet a new process is not started? or?

vext01: Wait, does the octane runner run only one process execution? Or it runs all benchmarks in one process execution, which itself is repeated?

ltratt: Look at `runoctane.py`.

ltratt:
Are we ready to squash?

patches/octane_run_we.js
```diff
@@ -0,0 +1,51 @@
...
+load('typescript-input.js');
+load('typescript-compiler.js');
+
+
+function main() {
+  var suites = BenchmarkSuite.suites;
+  for (var i = 0; i < suites.length; i++) {
+    var suite = suites[i];
+    var benchmarks = suite.benchmarks;
+    for (var j = 0; j < benchmarks.length; j++) {
```
vext01, line 28: I'm still very confused. run_octane.py spawns new processes, so this loop can't be the process loop? So what does this loop do? It looks like we mighjt be running >1 benchmark per process. We don't want that, rigtht?

ltratt: The first two loops just deal with Octane's particular data-structure (benchmark suites have benchmarks). There's nothing very deep in this.

vext01:
Resolved via mattermost. Go ahead an squash.

ltratt:
Squashed and ready for merging.

vext01 (approved):

PR #217, opened 2016-11-01
Create summary tables from startup results.

### LINKED ISSUE #200: New script to generate useful information from the results of the startup experiment
We need to generate useful output from the results of the startup experiments. Our usual plots are not very useful, tables with means and confidence intervals would be more useful.

snim2:
This is a first bash at producing tables from the startup experiment. The result is a LaTeX file with averages of startup times listed per-vm and per-language variant.

I'm not sure whether that is really the most useful presentation of the data, so suggestions are welcome.

A short test from bencher2 can be found here: [table.pdf](https://github.com/softdevteam/warmup_experiment/files/564517/table.pdf)

Fixes #200

### DIFF
--- a/bin/table_startup_results
+++ b/bin/table_startup_results
@@ -1,0 +1,142 @@
+#!/usr/bin/env python2.7
+
+"""Create a LaTeX summary of a Krun results file from the startup experiment."""
+
+import argparse
+import os
+import os.path
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+from warmup.krun_results import read_krun_results_file
+from warmup.statistics import bootstrap_confidence_interval
+
+TITLE = 'Startup Experiment Results'
+
+__LATEX_HEADER = """
+\documentclass[12pt]{article}
+\usepackage{booktabs}
+\\title{%s}
+\\begin{document}
+\maketitle
+""" % TITLE
+
+__LATEX_SECTION = lambda section: """
+\\section*{%s}
+""" % section
+
+__LATEX_START_TABLE = """
+\\begin{tabular}{lr}
+\\toprule
+VM & Time (secs) \\\\
+\\midrule
+"""
+
+__LATEX_END_TABLE = """
+\\bottomrule
+\\end{tabular}
+"""
+
+__LATEX_FOOTER = """
+\\end{document}
+"""
+
+
+def main(data_dcts, latex_file):
+    summary_data = {'vms':dict()}
+    for machine in data_dcts:
+        keys = sorted(data_dcts[machine]['wallclock_times'].keys())
+        for key in keys:
+            wallclock_times = data_dcts[machine]['wallclock_times'][key]
+            if len(wallclock_times) == 0:
+                print ('WARNING: Skipping: %s from %s (no executions)' %
+                       (key, machine))
+            elif len(wallclock_times[0]) == 0:
+                print('WARNING: Skipping: %s from %s (benchmark crashed)' %
+                      (key, machine))
+            else:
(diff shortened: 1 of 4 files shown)


vext01:
Here's some thoughts.
- We need confidence intervals. You can use the bootstrap method from libkalibera. I'll look for examples for you soon.
- Remove `default-` from language variants. Readers wont care for that.
- That column should be just `Language` too.
- Shorten `Average startup time (seconds)` to `Mean startup (s)`
- I wonder if the second table is useful... for `default-java` you will be munging hotspot and Graall into one. Perhaps leave the table, and we can choose to use it, or not, later.
- Can we make a table keyed by benchmark and VM?

vext01:
Example of bootstrapping confidence intervals from pyhyp paper:
https://github.com/softdevteam/pyhyp_experiments/blob/master/latex_table.py#L200

vext01:
I wonder if the second table needs geomeans. The benchmark pair table will.

You should find geomean code in that same file.

snim2:
Is it useful to have a table keyed by benchmark, when we don't actually run the benchmarks? I did think about this then removed it, perhaps erroneously.

vext01:
Hah, you are completely right. My bad!

vext01:
but we do still need CIs.

snim2:
Will do that next

ltratt:
This is a good first start!

I think the first table should be "VM | Time(s)" (shorter column headings will be vital given the layout we have to hit) and we explain in the caption that it's the mean (with 99% confidence intervals). I don't think we want the second table at all?

Some LaTeX nerdery: we use `booktabs` and its `toprule`/`midrule`/`bottomrule` stuff (and no vertical lines). The docs for it are annoyingly opinionated to the extent that they can be off-putting, but it's hard to disagree with the beautiful output it leads to.

I am a bit surprised by the HHVM numbers, though without any concrete reason for being so. Can someone maybe double check that we've measured the right thing?

snim2:
I have changed the column names. The CI stuff isn't ready to push yet. I have used `booktabs` from the start, does this need changing?

(BTW the data was from a very short run on Bencher2)

ltratt:
Ah, yes, I can see the booktabs thing now. Mea culpa! We just need to get rid of the ugly vetical bar in the table (i.e. should be `{lr}` not {l|r}). I'm not sure what "CI stuff" is?

snim2:
CI == "confidence interval"!

ltratt:
Aha!

snim2:
Next test case: [table.pdf](https://github.com/softdevteam/warmup_experiment/files/566930/table.pdf)

Confidence intervals are in, so this is now ready for another review. We need to merge this before #218 which contains some similar code factored out into a library module.

vext01:
Minor nits/thoughts.
- set the error in smaller font
- centre column headings?
- how many DPs do we want on error?

vext01:
FWIW, in the pyhyp paper, we used 3DP for measurements, and 4DP for error.

You will need more DP due to the nature of what we are measuring. How about 5/6?

ltratt:
I'm not quite sure why the VM heading is to the left of the VM names below?

snim2:
Test case again: [table.pdf](https://github.com/softdevteam/warmup_experiment/files/567014/table.pdf)

ltratt:
Works for me. If Edd is happy, then I think we can squash, then merge.

vext01:
Teensy nit. \pm should be small too, right?

vext01:
Other than that, nice!

vext01:
squash?

vext01:
done?

snim2:
Yup!

vext01 (approved):

PR #218, opened 2016-11-02
Summarise categories

### LINKED ISSUE #216: Summarise benchmark classifications in tables
Something like this:

```
  bench name  | warmup-style | iteration# steady state | steady state timings
  ------------+--------------+-------------------------+---------------------
  bench1      | warmup       | 2                       | 3.2s +- 0.1s
  bench2      | slowdown     | 123                     | 2.5s +- 0.2s
  bench3      | not stable   | n/a                     | n/a
```

snim2:
This is a first bash at summarising the classification and changepoint information in tables.

The test case is from the dacapo benchmark (which is why there is not much variation in the "styles"). However, the classifications will change soon in a later PR, so for this PR we just want to get the tabulation etc. right.

A couple of notes:
  * Data in the table is averaged over all the process executions for each benchmark. The "style" is the most commonly occurring style for all the process executions in each benchmark. Is this what we want?
  * The table headings are too long, and I'm not sure why they are incorrectly aligned.

As with #217 CIs are not in yet.

Fixes #216

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -20,7 +20,7 @@ from matplotlib import gridspec, pyplot
 from matplotlib.collections import LineCollection
 
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
-from warmup.krun_results import read_krun_results_file
+from warmup.krun_results import pretty_print_machine, read_krun_results_file
 from warmup.outliers import get_window
 from warmup.plotting import add_margin_to_axes, compute_grid_offsets
 from warmup.plotting import format_yticks_scientific, get_unified_yrange, style_axis
@@ -49,14 +49,6 @@ BENCHMARKS = {
     'spectralnorm':     'Spectral Norm',
     'fannkuch_redux':   'Fannkuch Redux',
     'nbody':            'N-Body',
-}
-
-# Machine names that need to be reformatted.
-MACHINES = {
-    'bencher3': 'Linux1/i7-4790K',
-    'bencher5': 'Linux2/i7-4790',
-    'bencher6': 'OpenBSD/i7-4790',
-    'bencher7': 'ARM',  # this machine has not been set up yet
 }
 
 GRID_MINOR_X_DIVS = 20
@@ -722,10 +714,7 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
         machine = data['audit']['uname'].split(' ')[1]
         if '.' in machine:  # Strip domain names.
             machine = machine.split('.')[0]
-        if machine in MACHINES:
-            machine_name = MACHINES[machine]
-        else:
-            machine_name = machine
+        machine_name = pretty_print_machine(machine)
 
         # Is there instrumentation data with this results file?
         instr_data = False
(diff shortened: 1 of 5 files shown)


snim2:
I forgot to post the test case: [summary.pdf](https://github.com/softdevteam/warmup_experiment/files/566204/summary.pdf)

snim2:
Test case: [summary.pdf](https://github.com/softdevteam/warmup_experiment/files/566379/summary.pdf)

snim2:
This is now waiting on #217 to be merged.

snim2:
This has been brought up to date with master, in order to use the code that was merged in #217 - hence the force-push. It is ready for another review.

Several things have changed since the last review:
- The startup-experiment summary code has been refactored to use the `warmup.latex` module, which avoids duplicate code in the two table generators.
- The latex module has been changed to take account of the comments and so on from #217 
- Confidence intervals have been added to the classification data

Test cases:
- Startup tables (just to check we don't have a regression): [table.pdf](https://github.com/softdevteam/warmup_experiment/files/567312/table.pdf)
- Classifications: [summary.pdf](https://github.com/softdevteam/warmup_experiment/files/567313/summary.pdf)

Some minor notes:
- The table headings are too long, but I'm not sure how best to re-word them.
- The tables show the average "number of iterations before the steady state was reached". This is the median value (with 99 CI) of the location of the last changepoint over all the process executions for each benchmark. This means that the "average" location of the last changepoint is now recorded as a float, which seems confusing. I'm not sure there is a way around this.

ltratt:
I think the columns could be something like {llll} "Benchmark | Classification | Steady iter | Mean" with longer explanations underneath. [Super long column names aren't going to work for this paper.] I think Classifications should be initial caps (e.g. "No steady state").

If the Classification is "No steady state" then we shouldn't give a steady iter or mean (they don't make any sense). [In both cases I'd suggest just leaving the cell totally blank.]

I'm unsure about non-integer values in the "Steady iter". I kind of feel like it's more sensible to round both the mean and CI, but I don't know if that's a good idea or not. Comments welcome!

snim2:
OK, next test case: [summary.pdf](https://github.com/softdevteam/warmup_experiment/files/568574/summary.pdf)

ltratt:
This is looking quite a bit better. Two things: I think the final column should be "Mean" and shouldn't the time (in seconds) be a real not an int?

snim2:
The time in seconds should be a real, yes. The average is not simple. The changepoint library gives us a list of means for each segment in each process-exec. The "steady-state time" is the mean of the last segment. We have many process exec for each benchmark, so we take an average over the last-segment-means. Because we do not know how the means are distributed, we use a non-parametric method (the bootstrap confidence interval) which gives us back a median and a measure of the error (the confidence interval). So, the number you are seeing in the table is a median of means. 

That is too long to explain in the table heading, but I think it is misleading to just call it a "mean".

snim2:
Again: [summary.pdf](https://github.com/softdevteam/warmup_experiment/files/568599/summary.pdf)

ltratt:
Obvious questoin: why use a median of means and not a mean of means? [There is a method for doing this that we've used in a paper somewhere. @cfbolz will remember more than me (including whether I'm wrong or not).]

snim2:
We are using the method from the PyHyp paper. The reason is, as above, it is a non-parametric method, and we do not know how the means are distributed. There may not be enough of them to apply the central-limit theorem.

ltratt:
OK.

ltratt:
[Though if we are using a median then "Median steady state" (or similar) might be better than "Average".]

snim2:
Next test case: [summary.pdf](https://github.com/softdevteam/warmup_experiment/files/568880/summary.pdf)

If you are both happy I would like to merge this sooner rather than later. No doubt there will be some fine-grained changes later on, but it would be helpful to have the summary tables available when I am working on the categorisation code.

ltratt:
I'm fine with merging this after squashing.

snim2:
@ltratt I think that's as much as I can sensibly squash!

ltratt (approved):

PR #219, opened 2016-11-02
Attempt to fix race condition between network and daemon pre-commands

vext01:
Bencher3 died mid-experiment in the night.

The reason it dies was because the network interface was coming up at the same time as we ran `service postfix stop`, causing the command to fail and krun to abandon ship.

Current pre commands:
```
PRE_EXECUTION_CMDS = [
        "sudo service cron stop",
        "sudo service atd stop",
        "sudo service postfix stop",
        "sudo service networking stop",
    ]
```

Corresponding systemd log (scroll right):

```
Nov 02 02:53:28 bencher3 sudo[1042]: kruninit : TTY=unknown ; PWD=/home/kruninit/warmup_experiment ; USER=root ; COMMAND=/usr/sbin/service cron stop
Nov 02 02:53:28 bencher3 sudo[1042]: pam_unix(sudo:session): session opened for user root by (uid=0)
Nov 02 02:53:28 bencher3 sudo[1042]: pam_unix(sudo:session): session closed for user root
Nov 02 02:53:28 bencher3 sudo[1062]: kruninit : TTY=unknown ; PWD=/home/kruninit/warmup_experiment ; USER=root ; COMMAND=/usr/sbin/service atd stop
Nov 02 02:53:28 bencher3 sudo[1062]: pam_unix(sudo:session): session opened for user root by (uid=0)
Nov 02 02:53:28 bencher3 sudo[1062]: pam_unix(sudo:session): session closed for user root
Nov 02 02:53:28 bencher3 sudo[1082]: kruninit : TTY=unknown ; PWD=/home/kruninit/warmup_experiment ; USER=root ; COMMAND=/usr/sbin/service postfix stop
Nov 02 02:53:28 bencher3 sudo[1082]: pam_unix(sudo:session): session opened for user root by (uid=0)
Nov 02 02:53:28 bencher3 dhclient[528]: DHCPDISCOVER on eth0 to 255.255.255.255 port 67 interval 8
Nov 02 02:53:28 bencher3 ifup[519]: DHCPDISCOVER on eth0 to 255.255.255.255 port 67 interval 8
Nov 02 02:53:28 bencher3 dhclient[528]: DHCPREQUEST on eth0 to 255.255.255.255 port 67
Nov 02 02:53:28 bencher3 dhclient[528]: DHCPOFFER from 137.73.8.8
Nov 02 02:53:28 bencher3 ifup[519]: DHCPREQUEST on eth0 to 255.255.255.255 port 67
Nov 02 02:53:28 bencher3 ifup[519]: DHCPOFFER from 137.73.8.8
Nov 02 02:53:28 bencher3 dhclient[528]: DHCPACK from 137.73.8.8
Nov 02 02:53:28 bencher3 ifup[519]: DHCPACK from 137.73.8.8
Nov 02 02:53:28 bencher3 dhclient[528]: bound to 137.73.9.137 -- renewal in 289421 seconds.
Nov 02 02:53:28 bencher3 ifup[519]: bound to 137.73.9.137 -- renewal in 289421 seconds.
Nov 02 02:53:28 bencher3 sshd[524]: Received SIGHUP; restarting.
Nov 02 02:53:28 bencher3 sshd[524]: Server listening on 0.0.0.0 port 22.
Nov 02 02:53:28 bencher3 sshd[524]: Server listening on :: port 22.
Nov 02 02:53:28 bencher3 systemd[1]: Unit postfix.service cannot be reloaded because it is inactive.
Nov 02 02:53:28 bencher3 postfix/master[717]: terminating on signal 15
Nov 02 02:53:28 bencher3 sudo[1082]: pam_unix(sudo:session): session closed for user root
Nov 02 02:53:28 bencher3 rc.local[530]: [2016-11-02 02:53:28: ERROR] Command failed: 'sudo service postfix stop'
Nov 02 02:53:28 bencher3 rc.local[530]: stdout:
Nov 02 02:53:28 bencher3 rc.local[530]: stderr:
Nov 02 02:53:28 bencher3 rc.local[530]: Job for postfix.service canceled.
Nov 02 02:53:28 bencher3 postfix[1141]: Stopping Postfix Mail Transport Agent: postfix.
Nov 02 02:53:28 bencher3 ifup[519]: run-parts: /etc/network/if-up.d/postfix exited with return code 1
Nov 02 02:53:28 bencher3 ifup[519]: Failed to bring up eth0.
```

I think the reason the command fails is because when the network comes up, systemd starts dependent services, or HUPs the existing ones, and we can't stop daemons while that happens. Here is a simple way to reproduce the error you see in the log:

```
# systemctl stop postfix & systemctl start postfix
[1] 4119
Job for postfix.service canceled.
[1]+  Exit 1                  systemctl stop postfix
```

But if we stop and start sequentially, no issue:
```
# systemctl stop postfix
# systemctl start postfix
# 
```

Now the fix. If we bring the network down first, then stop our daemons, then there should be no race between DHCP and us bringing services down. Also, be should use `systemctl` not the legacy `service`. I checked on my laptop that `systemctl stop networking` correctly stops the interface, and that `systemctl start networking` brings it back. The latter command takes quite some time to run (5 secs or so), so I suspect it is blocking for DHCP. For good measure added a sleep too.

This is untested. I need a code review before I try it on a bencher.

Thoughts, comments?

### DIFF
--- a/warmup.krun
+++ b/warmup.krun
@@ -140,21 +140,26 @@ SKIP = [
 N_EXECUTIONS = 10  # Number of fresh processes.
 
 if sys.platform.startswith("linux"):
+    # Assumes systemd is used for init.
     PRE_EXECUTION_CMDS = [
-        "sudo service cron stop",
-        "sudo service atd stop",
-        "sudo service postfix stop",
-        "sudo service networking stop",
+        # Stop network first. If an interface accepts a DHCP lease during one
+        # of the later commands below, it can cause the command to be
+        # "cancelled" by systemd. Bringing the network itself down can fail in
+        # the same way, so keep trying (with sleeps between).
+        "while true; do sudo systemctl stop networking; sleep 5; ping -q -c %s || break; done" % PING_HOST
+        "sudo systemctl stop cron",
+        "sudo systemctl stop atd",
+        "sudo systemctl stop postfix",
     ]
 
     PING_HOST = "bencher2.soft-dev.org"
     POST_EXECUTION_CMDS = [
         # The network doesn't always come up properly on debian. We keep trying
         # until we can ping a host on the network.
-        "while true; do ping -c 3 -q %s && break; sudo service networking stop; sleep 5; sudo service networking start; done" % PING_HOST,
-        "sudo service cron start || true",
-        "sudo service atd start || true",
-        "sudo service postfix start || true",
+        "while true; do ping -c 3 -q %s && break; sudo systemctl stop networking; sleep 5; sudo systemctl start networking; done" % PING_HOST,
+        "sudo systemctl start cron || true",
+        "sudo systemctl start atd || true",
+        "sudo systemctl start postfix || true",
     ]
 elif sys.platform.startswith("openbsd"):
     PRE_EXECUTION_CMDS = [


snim2:
+1 from me

warmup.krun
```diff
@@ -140,21 +140,24 @@ SKIP = [
...
 
 if sys.platform.startswith("linux"):
+    # Assumes systemd is used for init.
     PRE_EXECUTION_CMDS = [
-        "sudo service cron stop",
-        "sudo service atd stop",
-        "sudo service postfix stop",
-        "sudo service networking stop",
+        # Stop network first. If an interface accepts a DHCP lease during one
+        # of the later commands, it can cause them to fail.
```
snim2: I had to parse this comment a couple of times before I understood. Maybe replace "of the later commands, it can cause them to fail." with "of the later commands, it can cause those commands to fail."

vext01:
Argh, I realised bring the network down can fail too. Try my last commit...

warmup.krun
```diff
@@ -140,21 +140,27 @@ SKIP = [
...
-        "sudo service cron stop",
-        "sudo service atd stop",
-        "sudo service postfix stop",
-        "sudo service networking stop",
+        # Stop network first. If an interface accepts a DHCP lease during one
+        # of the later commands below, it can cause the command to be
+        # "cancelled" by systemd. Bringing the network itself down can fail in
+        # the same way, so try up to 5 times (with sleeps between) before
+        # finally giving up.
+        "sudo sh -c 'for i in 1 2 3 4 5; do systemctl stop networking && break; sleep 10; false; done'",
```
ltratt: I must admit, I don't understand how DHCP could fire up later. But, if it does, why not copy (with alterations) the `while` loop for restarting the network?

vext01: E.g. the lease is renewed, or the initial lease still didn't happen (I think the latter is what happened in the case we saw).

I pretty much did copy the loop, just I didn't want it to do this forever. Well, we could make it run forever, if we really wanted. Do we?

ltratt: Why 5? Why not infinite? Also, having a ping means that we can quit as soon as its worked.

vext01: I suppose infinite is fine.

vext01:
Try this. The loop exits when grep returns 1, thus indicating that no UP lines were found.

vext01:
Should we search for `RUNNING` instead? It's a longer string.

ltratt:
It looks reasonable to me. Have you been able to double-test the shell command? [I checked the precedences of the various operators against `man sh` on OpenBSD, but I wouldn't stake my life on it!]

ltratt:
I don't know what `RUNNING` means. `^[ \t]*UP` might be safer?

vext01:
I checked the failure case by inserting `true` in place of the down part, it goes round forever

vext01:
seems running an up are not well defined: 
https://www.quora.com/What-is-the-difference-between-the-UP-and-RUNNING-flags-of-an-ifconfig-output

Will try your regex idea

ltratt:
Let's try it for real on `bencher2`? If that goes down, we're not as dead as if `bencher3` goes down...

vext01:
i'm not keen on the regex. It assumes UP comes first.

vext01:
We can't risk bencher2 going down, or the other machines will fail to copy data onto it.

ltratt:
UP is at the beginning of the line on bencher2. But not for bencher3?

ltratt:
Alternatively, why not just use the ping command?

vext01:
It's the first flag on bencher3 at the moment, but always, not sure.

The problem with ping, is that it is a rough proxy for knowing if the interface is down. A router failure could cause the ping to fail, but the interface is not down.

ltratt:
Yes, but since we apparently can't reliably determine if an interface is up...

vext01:
I'm at a loss. So ping or UP? Up to you

ltratt:
Well, if we don't trust `^UP`, then we don't know how to reliably interpret `ifconfig`'s output. At which point, we can at least say that we know what `ping` does...

vext01:
OK. So we can use ping. But we must accept the small chance that the interface was not really down. I will try and find a way to fix this for the real run.

ltratt:
`ping` is imperfect, but in a way that we at least understand.

ltratt:
Assuming it works, I'm fine with squashing then merging this.

vext01:
Argh, it's not right. if systemctl fails the loop exits.

snim2:
OK, if this has been tested then let's squash

vext01:
I've tested it best I can. Squashing.

snim2:
All done?

snim2 (approved):

vext01:
yup

PR #224, opened 2016-11-03
Disable tradesoap.

ltratt:
It has, once, failed an internal validation check, so it's best disabled.


cfbolz (approved):
Makes sense to me (no sense to risk disturbing a run for it, imo).

PR #231, opened 2016-11-04
Plot sizes

vext01:
Various. See individual commits for details.

Example plot:
[slowdown_and_migrate.pdf](https://github.com/softdevteam/warmup_experiment/files/572190/slowdown_and_migrate.pdf)

This was generated with:
`python2.7 /home/vext01/research/warmup_experiment/bin/plot_krun_results --export-size 4,10 b5/warmup_results.json.bz2 --one-page -b bencher5:binarytrees:PyPy:default-python:0 -o paper-examples/slowdown_and_migrate.pdf`

I'll check it looks OK in the paper too.


ltratt:
Dumb comment: "0.000x10^9" is the same as 0 I think ;)

vext01:
Yes, but all the ticks have to be in the same format.

ltratt:
Really? That seems silly :/ It's just confusing. Oh well.

vext01:
I kinda wish the exponent could be shown once somewhere.

ltratt:
matplotlib has a way of putting the exponent at the top of the y axis, but then we always end up not noticing it's there...

vext01:
Will look into it

vext01:
page5

snim2:
@vext01 you could move the exponent into the y-label. I thought this would be easily missed, but you might prefer it.

vext01:
Don't suppose you know how that is done?

ltratt:
Let's not move the exponent. We know from past experience that we _always_ read such axes incorrectly.

vext01:
Alright, so in that case, any comments, can we merge?

snim2:
@vext01 just FYI in `warmup/plotting.py` there is a function which formats the exponent. It first draws the plot (because some artists will be `None`), it then finds the exponent, switches off displaying the exponent, switches on using latex, then adds appends the exponent to the ylabel.

vext01:
Thanks. I found that code just before Laurie said not to bother. So we are good to go?

snim2:
Have you tested with and without `-b` and `--one-page`?

vext01:
Not massively -- perhaps this needs more testing

snim2:
Incidentally, adding a `zorder` to the plots up the stack is probably best practice, but it is only necessary if there is more than one artist on the axes. On the measurement plot we can have outliers, a rolling median, the Tukey band, and the changepoints, so we need the `zorder` there. 

Leave this in though, it is good future-proofing.

vext01:
The line was being drawn under the axis grid, obscuring the plot.

snim2:
Ugh :( 

When you are happy that this has been tested with all the CLI options I'm happy for this to be squashed and merged.

vext01:
Just testing with all data from one machine... -b and --one-page seem to work still.

vext01:
OK, I didn't spot anything broken. Go ahead an merge. The commits are already isolated correctly.

snim2 (approved):

PR #232, opened 2016-11-07
Startup fixes

vext01:
Hi,

This fixes a buffer overrun and simplifies the code a lot. We also move the clock reading later, thus avoiding reading the majority of printfs. See individual commit messages for more detailed info.

No need to valgrind, as there are no allocations any more.

Output checked to be the same as before (bar emitting aperf before mperf for consistency with elsewhere). Please check carefully.

On OpenBSD (piped to Python json parser to check validity):
```
$ ./outer_startup_runner_c ./startup_runner_c | python -c "import json, sys; j = json.load(sys.stdin); print j"
{u'core_cycle_counts': [], u'wallclock_times': [4702.001296, 4702.002543], u'mperf_counts': [], u'aperf_counts': []}
```

On a bencher3/5, you woud see (hacked on OpenBSD to pretend there are 4 per-core measurements):
```
$ ./outer_startup_runner_c ./startup_runner_c | python -c "import json, sys; j = json.load(sys.stdin); print j"
{u'core_cycle_counts': [[0, 0], [0, 0], [0, 0], [0, 0]], u'wallclock_times': [4690.826458, 4690.827671],
u'mperf_counts': [[0, 0], [0, 0], [0, 0], [0, 0]], u'aperf_counts': [[0, 0], [0, 0], [0, 0], [0, 0]]}
```

OK?

### DIFF
--- a/startup_runners/outer_startup_runner.c
+++ b/startup_runners/outer_startup_runner.c
@@ -13,44 +13,47 @@
 
 #include "libkruntime.h"
 
-#define BENCH_FUNC_NAME "run_iter"
+void emit_dummy_per_core_data(char *, int);
+
+void
+emit_dummy_per_core_data(char *name, int num_per_core_data)
+{
+    int i;
+
+    printf("\"%s\": [", name);
+    for (i = 0; i < num_per_core_data; i++) {
+        printf("[0, 0]");
+        if (i < num_per_core_data - 1) {
+            printf(", ");
+        }
+    }
+    printf("], ");
+}
 
 int
 main(int argc, char **argv)
 {
-    double start_time = -1;
-    int result;
-    char fake_datum[] = "[0, 0], ";
-    char end_datum[] = "[0, 0] ";
-    int datum_size = (int)strlen(fake_datum);
+    int result, num_per_core_data;
+
+    if (argc != 2) {
+        printf("usage: ./outer_startup_runner_c <inner-runner>\n");
+        exit(EXIT_FAILURE);
+    }
+
     krun_init();
-    char* fake_data = malloc(1);
+    num_per_core_data = krun_get_num_cores();
 
-    if (krun_get_num_cores() > 1) {
-        memcpy(fake_data, fake_datum, datum_size);
-    }
-    if (krun_get_num_cores() > 2) {
-        for(int i=0; i < krun_get_num_cores() - 2; i++) {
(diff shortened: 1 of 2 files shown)


snim2:
This all looks good to me.

startup_runners/outer_startup_runner.c
```diff
@@ -13,44 +13,49 @@
...
-    /* silence gcc */
-    argc = argc;
-    argv = argv;
-
-    start_time = krun_clock_gettime_monotonic();
-    fprintf(stdout, "{ \"core_cycle_counts\": [ %s ],", fake_data);
-    fprintf(stdout, " \"mperf_counts\" : [ %s ],", fake_data);
-    fprintf(stdout, " \"aperf_counts\" : [ %s ],", fake_data);
-    fprintf(stdout, " \"wallclock_times\" : [ %f, ", start_time);
+    (void) argv;
```
ltratt: If we don't use `argv`, why not just make it `int main()`?

vext01: Oops, we do. Good catch.

vext01: Fixed

startup_runners/outer_startup_runner.c
```diff
@@ -13,44 +13,47 @@
...
 main(int argc, char **argv)
 {
-    double start_time = -1;
-    int result;
-    char fake_datum[] = "[0, 0], ";
-    char end_datum[] = "[0, 0] ";
-    int datum_size = (int)strlen(fake_datum);
-    krun_init();
-    char* fake_data = malloc(1);
+    int result, num_per_core_data = -1;
```
ltratt: Why is `num_per_core_data` initialised but not `result`? I don't think either needs initialising here.

vext01:
I ran this 10,000 times in a loop with no crash. Before I could get one in usually < 50

ltratt:
Modulo the new minor comment, I'm fine with this too.

startup_runners/outer_startup_runner.c
vext01: Also true.

vext01:
Fixed. Squash?

ltratt:
Yep, it sounds like we're both happy, so squash and then I can merge.

vext01:
There we go.

ltratt (approved):

PR #233, opened 2016-11-07
Ignore args 3 and above.

vext01:
When I tested before, I passed 2 args, krun passes more, but the startup runner does not use them.

OK?

### DIFF
--- a/startup_runners/outer_startup_runner.c
+++ b/startup_runners/outer_startup_runner.c
@@ -35,7 +35,11 @@ main(int argc, char **argv)
 {
     int result, num_per_core_data;
 
-    if (argc != 2) {
+    /*
+     * This startup runner needs one argument. Under normal operation, krun
+     * passes some other flags, which can safely be ignored.
+     */
+    if (argc < 2) {
         printf("usage: ./outer_startup_runner_c <inner-runner>\n");
         exit(EXIT_FAILURE);
     }


startup_runners/outer_startup_runner.c
```diff
@@ -35,7 +35,7 @@ main(int argc, char **argv)
 {
     int result, num_per_core_data;
 
-    if (argc != 2) {
+    if (argc < 2) {  // krun passs more args, which are ignored here.
```
ltratt: Typo: "passs".

Also the comment doesn't make sense to me: if Krun passes more than the expected number of args, this program will raise an error (rather than ignoring them).

vext01: Typo yes.

Disagree with the second statement though. _less than_.

ltratt: The comment still doesn't make sense. If Krun passes (say) 5 arguments, the program will raise an error. Something doesn't add up to me.

vext01: No it won't. If we pass 5 args, argc will be 6. `(6 < 2)` is false.

ltratt: Ah, I see what you mean. Can we simplify the comment to something like "Ensure we get the arguments we need, but we're OK with extra arguments being passed (such as by Krun)". [OK, that's a rubbish comment, but you get the idea.]

ltratt: Ah, I get it now.

vext01:
Try that.

ltratt:
Works for me! Squash and I'll merge?

vext01:
splat

ltratt (approved):

PR #234, opened 2016-11-07
Allow zooming cycle count yrange.

vext01:
Idle cores cause the axis on busy core to be too zoomed out. Add a switch.
OK?

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -137,7 +137,8 @@ def add_instr_data(data_dictionary, key, machine, instr_dir):
 
 def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
          xlimits, with_outliers, unique_outliers, changepoints, median=False,
-         tukey=False, inset_xlimit=None, one_page=False, legend_off=True):
+         tukey=False, inset_xlimit=None, one_page=False, legend_off=True,
+         cycles_ylimits=None):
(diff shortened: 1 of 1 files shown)


bin/plot_krun_results
```diff
@@ -1093,6 +1101,17 @@ if __name__ == '__main__':
         print("invalid --export-size argument")
         sys.exit(1)
 
+    # core cycles y-axis range
```
snim2: I think you could safely kill this comment

snim2:
Generally this is good, but this is a bit out of step with out we deal with `--xlimits`. For `--xlimits` we have an example of the expected input in the CLI parser:

``` python
                        help="Specify X-axis limits as a comma separated pair "
                             "'start,end'. Samples start from 0. e.g. "
                             "'-x 100,130' will show samples in the range "
                             "100 to 130.")
```

and we don't check that the user has passed in something sensible with a try/except, but we probably should!

vext01:
Will fix the usage, but the xlimit bug is for another day.

vext01:
Try that. Fixed a bug too.

snim2:
I haven't tested with a huge number of options, but this works for me. Want to squash?

vext01:
splat

snim2 (approved):

PR #237, opened 2016-11-10
DaCapo outputs timings in msec, but we need secs.

ltratt:


cfbolz (approved):
uh, I apologize, I missed this. looks good!

PR #239, opened 2016-11-12
Startup table

vext01:
Improved startup table, and a bootstrap fix. Bootstrap iterations != number of in-process iterations.

table

P.S. I think we should name the machines in the paper like `linux1`, `linux2` to save space, i.e. the i7_4790k part isn't that important.

OK?

### DIFF
--- a/bin/table_startup_results
+++ b/bin/table_startup_results
@@ -6,6 +6,7 @@ import argparse
 import os
 import os.path
 import sys
+from collections import OrderedDict
 
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
 from warmup.krun_results import read_krun_results_file
@@ -14,12 +15,12 @@ from warmup.latex import format_median_error, section, start_table
 from warmup.statistics import bootstrap_confidence_interval
 
 TITLE = 'Startup Experiment Results'
-TABLE_FORMAT = 'lr'
-TABLE_HEADINGS = 'VM & Time (secs)'
 
 
-def main(data_dcts, latex_file):
-    summary_data = {'vms':dict()}
+def main(data_dcts, latex_file, machine_map):
+    # machine -> vm -> times
+    summary_data = {machine: {} for machine in data_dcts.keys()}
+    all_vms = set()
     for machine in data_dcts:
         keys = sorted(data_dcts[machine]['wallclock_times'].keys())
         for key in keys:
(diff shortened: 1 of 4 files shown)


bin/table_startup_results
```diff
@@ -97,7 +97,8 @@ def create_cli_parser():
     parser.add_argument('json_files', action='append', nargs='+', default=[],
                         type=str, help='One or more Krun result files.')
     parser.add_argument('--outfile', '-o', action='store', dest='latex_file',
-                        type=str, help='Name of the LaTeX file to write to.')
+                        type=str, help='Name of the LaTeX file to write to.',
+                        default='startup.tex')
     return parser
```
snim2: Not having a default filename was deliberate (it was to avoid the problem we had with the plotting script, where the complaint was that a PDF could be overwritten without the user realising). The other table script will be the same (and presumably we want them both to have similar CLI behaviour.

So, I think if the user hasn't used `-o` we should probably write something to STDOUT and bail, rather than giving a default. Same with `table_classification summaries`.

warmup/statistics.py
```diff
@@ -10,7 +10,7 @@ def mean(seq):
     return math.fsum(seq) / len(seq)
 
 
-def bootstrap_confidence_interval(seq, confidence=CONFIDENCE):
+def bootstrap_confidence_interval(seq, confidence=CONFIDENCE, iterations=ITERATIONS):
```
snim2, line 13: OK, but this is also called in `table_classification_summaries`

bin/table_startup_results
```diff
@@ -71,15 +71,18 @@ def write_results_as_latex(summary, all_machines, tex_filename):
         fp.write(preamble(TITLE))
         for section_heading, summary in sections:
             fp.write(section(section_heading))
-            table_format = "l" + ("r" * len(all_machines))
+            table_format = "l" + ("r" * len(machine_map))
             table_headings1 = "&".join(
                 ['\multicolumn{1}{c}{\multirow{2}{*}{VM}}'] +
-                ["\multicolumn{%s}{c}{Machine}" % len(all_machines)])
-            table_headings2 = "&".join([''] + ["{\\small %s}" % x for x in  all_machines])
+                ["\multicolumn{%s}{c}{Machine}" % len(machines_map)])
```
snim2: `'` for quotes in the rest of this script (and the others)

bin/table_startup_results
```diff
@@ -91,15 +94,21 @@ def get_data_dictionaries(json_files):
     dictionaries of machine name -> JSON values.
     """
     data_dictionary = dict()
-    for filename in json_files:
+    machine_map = OrderedDict()
+    for arg in json_files:
+        machine_name, filename = arg.split(":")
```
snim2: likewise

snim2:
OK, apart from those comments (which I think GH turned into reviews or something?) this looks good to me.

ltratt:
I don't see any comments?

snim2:
Because flipping GH saved them as reviews. I'll try again

ltratt:
OK, the comments are there now. Thanks!

bin/table_startup_results
vext01: OK. Will fix.

warmup/statistics.py
vext01: Adding a kwarg won't affect the existing call-sites. The default is used. I only added this so that I could test with 1 iteration to make testing fast.

vext01:
I think that addresses the comments. Ready for review.

warmup/statistics.py
snim2: OK, sure.

snim2:
Do you want to add `required=True` to `-o` in the other table script in this PR?

vext01:
done, squash?

snim2:
lgtm

vext01:
@ltratt any comments before squash?

ltratt:
Nope, squash away!

vext01:
splat

ltratt (approved):

PR #244, opened 2016-11-12
Changepoints are blobs at ymin.

snim2:
First bash at blobby changepoints:

blobby

I'm not sure what you want to do about the segment means.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -367,10 +367,6 @@ def draw_process_exec(grid_cell, data, cycles_data, aperf_data, mperf_data,
         means = list()  # Draw means between ((x0, y0), (x1, y1)) pairs.
         assert len(changepoints) == len(changepoint_means) - 1
         for index, x_location in enumerate(changepoints):
-            # Draw changepoint.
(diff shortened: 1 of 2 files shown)


ltratt:
I think this is good. For the paper, I think we might need to simply remove the means, because they're probably going to be invisible. Or, maybe, we show them in a case where they're much clearer (i.e. when the data occupies a much greater portion of the y axis).

snim2:
So, are we saying we want different representations of changepoints for the paper and for our own use?

ltratt:
Probably (I can't say _for sure_, but I think so).

snim2:
OK, with this commit `--with-changepoint-means` produces this:

cpt_means

`--with-changepoints` produces this:

cpts

ltratt:
Excellent!

snim2:
OK, let's wait for @vext01 to review.

ltratt (approved):

ltratt:
I think it's fine for this to go in now -- we'll prboably need to do more fiddling in the future, but this is plenty good enough for now.

snim2:
OK, sure

PR #246, opened 2016-11-13
Use LaTeX machine names.

### LINKED ISSUE #245: Machine names need to be formatted as in the paper
Currently plots are in the "old" format (e.g. "Linux1/i7-4790K"). They should be converted to the new (e.g. "Linux$_\textrm{4790K}$").

snim2:
Fixes #245

Example:

[test_machine_name.pdf](https://github.com/softdevteam/warmup_experiment/files/587969/test_machine_name.pdf)

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -115,6 +115,7 @@ FONT = {
     'weight': 'regular',
     'size': BASE_FONTSIZE,
 }
+matplotlib.rc('text', usetex=True)
 matplotlib.rc('font', **FONT)
 
 MAX_SUBPLOTS_PER_ROW = 2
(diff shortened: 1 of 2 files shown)


ltratt:
Is it just me, or is the subscript the same size as the main text?

snim2:
It's in `textrm` right? So I think that's what we'd expect

ltratt:
The subscript should be in a smaller font (normally). I thought I was going mad, but I've checked and you can see the issue here:
image
The transparent text in the background is the paper, the black text in the foreground the PDF on this PR. Notice that the "Linux" text is exactly the same size in both PDFs, but the transparent subscript text is much smaller.

I wonder if matplotlib does a sort of latex simulation or something and gets the sizes wrong? It's not necessarily something we need to fix this second (we have bigger fish to fry), but it's probably worth noting and fixing at some point.

snim2:
This might be the problem:

http://stackoverflow.com/questions/23824687/text-does-not-work-in-a-matplotlib-label

ltratt:
Ah, yes, maybe we're supposed to use `mathrm`. Honestly, I can't remember all of LaTeX/TeX's arbitrary rules :(

snim2:
Second go:

[test_machine_name.pdf](https://github.com/softdevteam/warmup_experiment/files/587989/test_machine_name.pdf)

ltratt:
Ah, yes, I now see, this is a bug in matplotlib's rendering. `mathrm` is a bit uglier in the paper, but in a way we can live with. So let's go with this change, and I'll adjust the paper too.

ltratt:
So if you squash, I can merge?

snim2:
Splat

ltratt (approved):

PR #251, opened 2016-11-14
Table fixes

ltratt:
Fixes various minor issues, and makes the process a bit less error prone.

New output:

image


ltratt:
The new commit makes bencher6's table look like:

image

I think this is ready for merging if you're happy with it. It doesn't need squashing IMHO.

vext01 (approved):
Bring on the fudge.

PR #258, opened 2016-11-15
Don't overwrite the vm variable.

ltratt:
This is used in later iterations of the loop, causing subtle breakage in the
output. Mea culpa!


vext01:
Does this affect the other script? we might oberwrite `bench` not `vm` though due to column swaps.

vext01:
There's no map in the other one, so I think we are ok?

ltratt:
The other script is your purview, not mine ;)

vext01 (approved):

PR #259, opened 2016-11-15
Slowdown must always check for a steady state.

snim2:


ltratt:
How difficult would it to be to hoist the "no steady state" check to the beginning of the `if` block so that each branch doesn't have to check it?

snim2:
Please check that this is equivalent to what we had before...

ltratt:
It looks good to me, but we should double-check it on the plots to be doubly sure.

snim2:
OK, replotting b3 now

snim2:
Test case: [bencher3_warmup_07_plots.pdf](https://github.com/softdevteam/warmup_experiment/files/592015/bencher3_warmup_07_plots.pdf)

ltratt:
This looks good to me. Squash and I'll merge?

snim2:
Squashed

ltratt (approved):

PR #262, opened 2016-11-15
Get PyPy JIT times from the jit_times json key, not from the raw_vm_events.

vext01:
No need to merge today. Just raising so it is not lost.


vext01:
This appears to be implemented from another PR. Closing.

PR #263, opened 2016-11-18
Fix -b and add --core-cycles

snim2:
This PR is intended to help automate the paper figures. 

The first commit fixes `-b` when used with changepoints and outliers.

The second commit adds a new switch `--core-cycles` which allows the user to specify which core cycle counts should be plotted.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -653,15 +653,13 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
     for each subplot.
     """
 
-    data_dictionary = {'data': dict(), # Measurement data to be plotted.
-                       'cycles_counts': dict(),
+    data_dictionary = {'data': dict(),  'cycles_counts': dict(),
                        'instr_data': dict(),  # Per-VM information.
-                       'all_outliers': dict(),
-                       'common_outliers': dict(),
+                       'changepoints': dict(), 'changepoint_means': dict(),
+                       'all_outliers': dict(), 'common_outliers': dict(),
                        'unique_outliers': dict(),
-                       'changepoints': dict(),
-                       'changepoint_means': dict(),
                       }
+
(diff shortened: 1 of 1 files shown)


bin/plot_krun_results
```diff
@@ -138,7 +138,7 @@ def add_instr_data(data_dictionary, key, machine, instr_dir):
 def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
          xlimits, with_outliers, unique_outliers, changepoints, changepoint_means,
          median=False, tukey=False, inset_xlimit=None, one_page=False,
-         legend_off=True, cycles_ylimits=None):
+         legend_off=True, core_cycles=(0,1,2,3), cycles_ylimits=None):
```
vext01, line 141: This assumes we have 5 cores. Is that what we want? Does it do the right thing on openbsd, with no core data?

vext01: Should the default be `None` to mean "all cores"?

bin/plot_krun_results
```diff
@@ -502,7 +502,8 @@ def draw_process_exec(grid_cell, data, cycles_data,
 def draw_page(is_interactive, executions, cycles_executions,
               instr_executions, titles, window_size, xlimits,
               outliers, unique, common, changepoints, changepoint_means, median,
-              tukey, inset_xlimit=100, legend_off=True, cycles_ylimits=None):
+              tukey, inset_xlimit=100, legend_off=True, core_cycles=(0,1,2,3),
```
vext01, line 507: same here.

vext01:
Looks OK apart from that minor question.

snim2:
OK, will update

snim2:
OK, ready to squash?

vext01:
Looks good. If it's tested, go ahead.

snim2:
Squashed.

vext01 (approved):

PR #266, opened 2016-11-21
Remove export size

### LINKED ISSUE #264: Fix magic numbers with --export-size
The magic numbers we use to set font sizes, the space between plots and so on are hard-coded. Now that we allow the user to set the size of each figure, in inches, with `--export-size` the magic numbers can sometimes result in:

  * Font sizes that are too small to read
  * Font sizes that are too large, and so overlap with other artists
  * Spacing between plots that seems "too large" and wastes space
  * Spacing between plots that is so small that some artists overlap

snim2:
This PR removes `--export-size` and attempts to fix some of the magic numbers. 

Fixes #264 

For testing, attached to this PR are:

  * A .zip of examples from the other repo, and the script which generates them 
[examples.zip](https://github.com/softdevteam/warmup_experiment/files/603480/examples.zip)
  * A page from a full set of 5x2 charts from one of the benchers [bencher3_without_exportsize.pdf](https://github.com/softdevteam/warmup_experiment/files/603479/bencher3_without_exportsize.pdf)
  * A screenshot of interactive mode interactive_mode


ltratt:
If I look at the one page PDF a) the horizontal plots are *really* close together b) the titles are *really* big. That almost certainly means that longer benchmark titles will merge together (which is a problem we had last week). Even as it is, the axes on the cycle-cores all but bump into each other.

Personally, I think the font sizes used in the plots are way too big here. Dropping them by something like 1/3 would probably fix this problem immediately.

vext01:
There's plenty of horizontal space isn't there? Looks like a good inch or so.

I think if you made the font sizes much smaller, they would not be visible when printed.

The plots are too wide to fit into the paper I think. You would have to scale them down in tex to get them in one column, then you would have squashed fonts?

Sarah, can you import one of these into the paper to give us an idea?

vext01:
Eh, sorry I was looking at the screenshot.

The fonts look quite big in the PDF yeah, and the width is ok here I think.

snim2:
This next test case is *just* with the commit that shrunk the font sizes: 
[bencher3_without_exportsize_smaller.pdf](https://github.com/softdevteam/warmup_experiment/files/603515/bencher3_without_exportsize_smaller.pdf)

vext01:
Please still put one in the paper and show a screenshot, so we can see where we stand.

snim2:
I think we can accept that interactive mode is never really going to be perfect, it's really only for us t use in testing.

vext01:
Can we get away with shortening 'proc. exec.' more? Can you use tex symbols in the classifications without using tex=True? I guess not.

ltratt:
OK, this is starting to look better w.r.t. font sizes.

I have one further suggestion: let's reduce the font size of the LHS y axis ticks of the CC plots. The units are close to meaningless, so making them so large suggests they're as important as the run-sequence plot. [In an ideal world, we'd also make "0.000x10^9" the same as "0" ;) but I don't know how easy that is.]

snim2:
Like this? [bencher3_without_exportsize_smaller_ytick.pdf](https://github.com/softdevteam/warmup_experiment/files/603643/bencher3_without_exportsize_smaller_ytick.pdf)

vext01:
test123 @snim2

snim2:
New bencher3 page: [bencher3_without_exportsize_smaller_ytick.pdf](https://github.com/softdevteam/warmup_experiment/files/603805/bencher3_without_exportsize_smaller_ytick.pdf)

vext01:
FWIW, I'm using `--export-size` to make plots of different sizes for slides.

snim2:
Different sizes in a way that couldn't be automated?

vext01:
Not sure what you mean. FWIW, the script I used is here:
https://github.com/softdevteam/softdev/blob/master/talks/warmup/plots_0_7/plot.sh

snim2:
I mean, if you are using a fixed number of sizes for specific purposes (for example, one size when you are using `--wallclock-only` and another size when you are not) then we could hard-code that and remove the CLI option.

vext01:
I see.

From experience it's really hard to get a plot containing the sub-plots you want and to have it the right size.  The plot script I used has about 5 different  `--export-size` values in use. We wouldn't want to hard code them all would we?

snim2:
OK, in that case I will close this PR and remove the branch.

PR #267, opened 2016-11-21
Mem fix

vext01:
Fix memory efficiency of plot script, and make the script resilient to missing instr data files.

The memory fix basically just makes it so that only the instrumentation data we need is loaded (a ChartData is for one pexec only now, not all pexecs for a given key) and that only one instr file is loaded in at a time.

memory consumption for the data we have typically bounces between around 4 and 16GB, whereas before it would outright crash.

Tested with and without -b, with and without --onepage, and with missing instr files. Seems ok.

OK?

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -133,7 +133,13 @@ def get_instr_data(key, machine, instr_dir, pexec_idxs):
             file_ = os.path.join(instr_dir, "%s__%s__%s__%s.json.bz2" %
                                  (bench, vm, variant, pexec_idx))
             print("Loading: %s" % file_)
-            js = read_krun_results_file(file_)
+            try:
+                js = read_krun_results_file(file_)
+            except IOError:
+                print("WARNING: Missing instrumentation data for: %s:%s:%s" % \
+                      (machine, key, pexec_idx))
+                ret.append(None)  # missing instr data
+                continue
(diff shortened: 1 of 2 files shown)


bin/plot_krun_results
```diff
@@ -122,17 +122,28 @@ EXPORT_SIZE_INCHES = [12, 10]
...
-        parser = INSTRUMENTATION_PARSERS[vm](merge_instr_data(pexec_data))
-        data_dictionary['instr_data'][key][machine] = parser.chart_data
+        for pexec_idx in pexec_idxs:
+            file_ = os.path.join(instr_dir, "%s__%s__%s__%s.json.bz2" %
+                                 (bench, vm, variant, pexec_idx))
+            print("Loading: %s" % file_)
+            try:
+                js = read_krun_results_file(file_)
+            except IOError:
+                print("WARNING: Missing instrumenataion data for: %s:%s:%s" % \
```
ltratt: Typo "instrumenataion".

snim2:
Other than the typo this seems OK to me.

vext01:
Squash?

snim2:
Yup!

vext01:
splat

vext01:
test mattermost notify, ignore

snim2 (approved):

PR #273, opened 2016-11-27
Optional table preamble

snim2:
This PR adds a new CLI option `--with-preamble` to the table scripts. 

By default, the output of the script will be (as before) just the table, suitable for copying and pasting into a larger document. 

With `--with-preamble` on, a full latex document will be written out, with a preamble, `\begin{document}` and `\end{document}`.


snim2:
@ltratt ready for review.

ltratt (approved):

PR #275, opened 2016-12-02
Report inconsistent benchmarks with steady states

snim2:
This is the latest output of the table_main script on the b3 data (you will have to rename the files):

  * Table for paper: [bencher3.table.txt](https://github.com/softdevteam/warmup_experiment/files/628374/bencher3.table.txt)
  * Full document generated by `--with-preamble`: [b3_table.tex.txt](https://github.com/softdevteam/warmup_experiment/files/628378/b3_table.tex.txt)

The full document does not correctly centre the table, I think this belies a deeper problem, and maybe the table header still isn't correct (it looks like it has two too many columns, but LaTeX fails if you remove a couple!

### DIFF
--- a/bin/table_classification_summaries_main
+++ b/bin/table_classification_summaries_main
@@ -153,6 +153,8 @@ def write_results_as_latex(machine, all_benchs, summary, steady_state, tex_file,
             fp.write(preamble(TITLE))
             fp.write('%s' % get_latex_symbol_map())
             fp.write('\n\n\n')
+            fp.write('\\begin{table*}[t]\n')
+            fp.write('\\centering\n')
         # emit table header
         heads1 = TABLE_HEADINGS_START1 + '&'.join([TABLE_HEADINGS1] * num_splits)
         heads2 = TABLE_HEADINGS_START2 + '&'.join([TABLE_HEADINGS2] * num_splits)
@@ -235,6 +237,7 @@ def write_results_as_latex(machine, all_benchs, summary, steady_state, tex_file,
             split_row_idx += 1
         fp.write(end_table())
         if with_preamble:
+            fp.write('\\end{table*}\n')
             fp.write(end_document())
     return
 
--- a/bin/table_classification_summaries_main
+++ b/bin/table_classification_summaries_main
@@ -111,7 +111,7 @@ def main(data_dcts, window_size, latex_file, with_preamble=False):
                     median_t, error_t = bootstrap_confidence_interval(time_to_steadys)
                     time_to_steady = format_median_error(median_t, error_t, brief=True)
                 else:  # No changepoints in any process executions.
-                    mean_last_cpt = format_median_error(median, 0, as_integer=True)
+                    mean_last_cpt = format_median_error(0, 0, as_integer=True)
                     time_to_steady = ''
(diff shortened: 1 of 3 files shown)


ltratt:
Can you post a PDF of the page with the main table? That way it's much easier to judge if the changes work in context. Thanks!

snim2:
* Page from the paper: [warmup.pdf](https://github.com/softdevteam/warmup_experiment/files/630783/warmup.pdf)
  * Document from `--with-preamble`: [b3_table.pdf](https://github.com/softdevteam/warmup_experiment/files/630786/b3_table.pdf)

ltratt:
I think this looks almost perfect.

Two quick comments. I reckon we should probably marginally increase the padding between the two columns (i.e. add whitespace to the left of where it says "n-body"). We could drop a dummy column with padding in there?

I also just noticed that LuaJIT/Richards is flat but the steady iter is 2?

snim2:
It can't be right though, the full document version is using `article, 10pt` just like the paper. It is centered and placed in a `table*` environment, just like the paper, but it isn't centered at all...?

ltratt:
Not sure I understand?

snim2:
The table in both PDFs should look the same, but it doesn't

ltratt:
Oh, I see. TBH, I only care about the one in the paper...

snim2:
The last two commits fixed the margins in `--with-preamble` and the reported "steady iter" for flat benchmarks. Anything else or are we ready to squash?

bin/table_classification_summaries_main
```diff
@@ -111,7 +111,7 @@ def main(data_dcts, window_size, latex_file, with_preamble=False):
                     median_t, error_t = bootstrap_confidence_interval(time_to_steadys)
                     time_to_steady = format_median_error(median_t, error_t, brief=True)
                 else:  # No changepoints in any process executions.
-                    mean_last_cpt = format_median_error(median, 0, as_integer=True)
+                    mean_last_cpt = format_median_error(0, 0, as_integer=True)
```
ltratt, line 114: Dumb question: how did `median` end up as something non-zero before? Surely it should always have been 0 anyway? I'm wondering if we're accidentally burying a more subtle bug here...

snim2: It isn't a subtle bug, but unfortunately GitHub has folded the relevant line of code. Before the if/else, `median` is set to the median of the last segment means, on line 108, so this is a mean time, not a mean iteration, but you wouldn't notice because it is printed as an integer.

ltratt: So are you saying that before it was showing a value of something (in essence) a bit random?

snim2: yes, it was showing the integer part of "steady state perf", if you look in the b3 table, this is 1.64 for PyPy / n-body (steady iter shows as 1) and 2.6 for Richards / LuaJIT (steady iter shows up as 2)

ltratt: OK, got it.

ltratt:
OK, I think this is ready for squashing.

snim2:
Squashing done.

ltratt (approved):

PR #276, opened 2016-12-06
Fix insets

### LINKED ISSUE #191: Fix --inset-xlimits
This has been broken by introducing `GridSpec` style charts, rather than the old calls to `subplot()`. The problem we have is that our code calls `axis.transAxes.transform()` to make the size of the inset plot small enough to fit inside the wallclock time chart. 

Maybe we can find a way around this, but if not we may have to add the "thumbnail" plot to the stack of larger plots, rather than embedding it inside the wallclock plot.

Other ideas welcome...

snim2:
Needs squashing.

This isn't a very tidy solution, but we will refactor soon anyway, so it will do for now!

Test case: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/634169/test_insets.pdf)

Fixes #191

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -5,7 +5,6 @@ Plot data from Krun results file(s).
 
 import argparse
 import datetime
-import glob
 import math
 import matplotlib
 matplotlib.use('Agg')
@@ -24,7 +23,7 @@ from warmup.krun_results import pretty_print_machine, read_krun_results_file
 from warmup.outliers import get_window
 from warmup.plotting import add_margin_to_axes, compute_grid_offsets
 from warmup.plotting import format_yticks_scientific, get_unified_yrange, style_axis
-from warmup.vm_instruments import INSTRUMENTATION_PARSERS, merge_instr_data
+from warmup.vm_instruments import INSTRUMENTATION_PARSERS
 
 seaborn.set_palette('colorblind')
 seaborn.set_style('whitegrid', {'lines.linewidth': 1.0, 'axes.linewidth': 1.0})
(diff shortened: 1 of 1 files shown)


vext01:
Looks good. Well done!

Remind me, are the visibility and xrange of the insets still configurable?

snim2:
Syntax is: `--inset-xlimits START_ITER,END_ITER` 
Placement is not configurable.

vext01:
And if you don't specify `--inset-xlimit`, you get no inset, right?

I think that's OK for now.

snim2:
That's right, default is not to have the inset at all

vext01:
Go ahead and squash.

ltratt:
This looks nice. I think, by default, 200 iterations is too many to show though, as the shape is lost a little bit. I'd like to suggest trying the following: default to displaying 50 iterations; if all of the changepoints in the file are <= iteration 100, then display up to the final changepoint + 25 iterations; but the absolute max is (say) 125 iterations.

We'll probably need to fiddle to get this looking good, but something like this seems plausibly useful.

snim2:
The number of iterations and the extent of them is passed in on the command line, e.g. `--inset-xlimits 0,100`, the attachment was just an example.

ltratt:
A fixed number of iterations won't often be useful, so let's default to doing something automatic that does a decent job in most cases.

vext01:
Suggest making that a separate PR?

ltratt:
It's a pretty small feature. I don't see that it's worth creating a new PR for.

ltratt:
On reflection, I see what you both mean. OK, let's squash this, then create a new PR for the "auto-sizing" and "auto-placement" stuff.

ltratt (approved):

snim2:
Squashed

PR #277, opened 2016-12-06
Automatically compute iterations which should be placed in an inset

snim2:
Algorithm as follows:
  * Use the first 2.5% of the iterations (50 iterations on a run of 2000)
  * If all changepoints occur in the first 5% of iterations (100 iterations on a run of 2000) then the inset plots up to the last changepoint + 2% of the iterations.
  * Clamp the above to a largest extent of 0->150 iterations.

Test case: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/634409/test_insets.pdf)

Needs squashing.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -100,6 +100,8 @@ PERF_YTICK_FORMAT = '%.1f'
 INSET_DEFAULT_WIDTH = 0.4
 INSET_DEFAULT_HEIGHT = 0.25
 INSET_PADDING = 0.035
+INSET_MIN_ITERS = 50
+INSET_MAX_ITERS = 150
 INSET_TICK_FONTSIZE = 8
 
 # Default (PDF) font sizes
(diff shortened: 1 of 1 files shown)


vext01:
Sounds reasonable, but I'll let Laurie give the final word.

ltratt:
On p1 the insets x axis only goes up to 45, but actually seems to plot 43 or 44 iterations?

vext01:
Maybe an example PDF would help?

snim2:
Hang about, I haven't finished yet!

ltratt:
OK.

vext01:
Sorry :jack_o_lantern:

snim2:
OK, the code now tries to detect whether the inset will collide with data. In this testcase, both H2 and Tradebeans do collide: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/634577/test_insets.pdf)

ltratt:
I can't see why tradebeans collides with anything (though H2 definitely does).

[Repeating the premature comment] On p1 the insets x axis only goes up to 45, but actually seems to plot 43 or 44 iterations?

snim2:
Yes there seem to be a few of those, but it isn't obvious to me why it's happening....

ltratt:
I think we need to fix those two things (tradebeans + short x axes), but otherwise this is looking very nice.

snim2:
Also I'm not sure we will get the right x-axis on the inset if we use `--xlimits` on the command line.

ltratt:
I'm happy for the manual option to be killed, at least for the time being.

snim2:
I mean, if you only want to plot iterations A-B on the main plot, you will get a silly inset.

snim2:
OK, next test case: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/634711/test_insets.pdf)

ltratt:
The x axis in the insets is now clearly wrong even on p1: 8?!

snim2:
OK, try again: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/634848/test_insets.pdf)

snim2:
Test case: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/634889/test_insets.pdf)

ltratt:
This is looking close to done. One final thing: the time-series data starts on the LHS of the x axis, but looks like it finishes at entry 49 (not 50) on the RHS. Any thoughts?

snim2:
Try this: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/635255/test_insets.pdf)

ltratt:
Perfect! Let's squash, then I can merge.

snim2:
Squashed!

ltratt (approved):

PR #278, opened 2016-12-06
Inset and wallclock times have identical ytick format.

snim2:
This PR makes the ytick formats to insets identical to those on the wallclock time axis.

This means that if the yticks are `1.2 x 10^-6` or similar, the inset will be formatted in the same way.

Test case: [test_insets.pdf](https://github.com/softdevteam/warmup_experiment/files/635278/test_insets.pdf)


ltratt:
I like it!

ltratt (approved):

PR #283, opened 2016-12-09
Fix inset placement

snim2:
This is a bug-fix PR, new-feature PRs will come later. Changes are:

  * Insets are drawn by default, `--no-insets` turns them off
  * Fixes a bug where insets were resized (not just moved) if they could not be placed in the default position.
  * The inset placement algorithm has been improved to look much harder to find a position where the inset can be drawn. This means that more charts can have insets than before.

Test case (particularly page 1 and 3): [test_30_new_placements.pdf](https://github.com/softdevteam/warmup_experiment/files/643435/test_30_new_placements.pdf)

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -1084,14 +1084,16 @@ def create_cli_parser():
                              "'start,end'. Samples start from 0. e.g. "
                              "'-x 100,130' will show samples in the range "
                              "100 to 130.")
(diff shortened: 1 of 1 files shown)


ltratt:
Looks good to me! Do you want to squash?

snim2:
I think it's probably fine as-is.

ltratt (approved):

PR #284, opened 2016-12-13
Update VMs.

vext01:
Let's start looking at this monster. Don't merge just yet.

I've tested using the `warmup.krun` config, with a small number of iterations and one exec.

For OpenBSD, I tested on -current (which should be close to 6.0) which has W^X enforced. For the VMs that needed it, I added the link flag. V8 and LuaJIT are W^X compatible out the box.

Note that we also need another JDK built, as graal needs a very specific version.

At this point, we can start thinking of whether to add the new JVM that was posted on ports@. My bulk build hasn't finished yet, but no errors so far. It will probably be OK. I think we should speculatively import this update now to save time. Thoughts?


vext01:
Note you have to click to see the build.sh change.

build.sh
```diff
@@ -237,64 +242,82 @@ build_pypy() {
...
+    cp ${pypy_make_dir}/pypy-c ${wrkdir}/pypy/pypy/goal/pypy-c || exit $?
+    rm -rf ${usession}
 }
 
-V8_V=5.1.281.65
-DEPOT_TOOLS_V=26f3e4eecb09d2608b1ac304de9d5d3c68de67ce
+# Look here to know which V8 branch is currently stable:
+# https://omahaproxy.appspot.com/
+#
+# Note that there is often a newer tag in git than what is shown on that page.
```
ltratt: "than what is shown on" -> "than is shown on"

build.sh
```diff
@@ -379,47 +405,102 @@ build_jdk() {
...
-    chmod -R 755 ${wrkdir}/openjdk/build || exit $?
+    # JDK installs some jar files unreadable to "other" users meaning that the
+    # benchmark user can't access them. This becomes a problem later for graal,
+    # which takes a copy of this JDK's jar files.
+    chmod -R 755 ${wrkdir}/openjdk/build/*-release/jdk/lib || exit $?
+}
+
+
+# This is a bootstrap JDK used only for Graal, which requiries a very specific
+# version of the JDK. Too old, or too new will not work.
```
ltratt: Can delete the "too old" sentence.

build.sh
```diff
@@ -379,47 +405,102 @@ build_jdk() {
...
+BOOT_JDK_TAR=${BOOT_JDK_BASE}_fullsource.tgz
+build_bootstrap_jdk() {
+    echo "\n===> Download and build graal bootstrap JDK8\n"
+    if [ -f ${BOOT_JAVA_HOME}/bin/javac ]; then return; fi
+
+    cd ${wrkdir} || exit $?
+    # We fetch a hand-rolled tarball, as the JDK repo build downloads things
+    # and I am not sure that they are fixed versions. The tarball was rolled on
+    # 20161102.
+    if [ ! -f ${wrkdir}/${BOOT_JDK_TAR} ]; then
```
ltratt: I'm not sure that date is correct? It says November, but I suspect December is correct? Also, can we use ISO format YYYY-MM-DD pretty please? :)

build.sh
```diff
@@ -379,47 +405,102 @@ build_jdk() {
...
+        wget https://archive.org/download/softdev_warmup_experiment_artefacts/distfiles/${BOOT_JDK_TAR}
+    fi
+    if [ ! -d ${BOOT_JDK_BASE}_fullsource ]; then
```
ltratt: Why not just have the `tar zxf` immediately after the `wget`?

build.sh
```diff
@@ -379,47 +405,102 @@ build_jdk() {
...
-MX="env JAVA_HOME=${OUR_JAVA_HOME} python2.7 ${wrkdir}/mx/mx.py"
+JVMCI_VERSION=jvmci-0.23
+MX_VERSION=d9c7efa53f60e4ca493da08438af01f8ca985985
+GRAAL_VERSION=graal-vm-0.18
 build_graal() {
     echo "\n===> Download and build graal\n"
 
+    if [ -f ${wrkdir}/graal-jvmci-8/jdk1.8*/product/bin/javac ]; then return; fi
+
+    # A tool called mx is used to clone the code
```
ltratt: This comment is superfluous.

ltratt:
I have some not-very-deep comments, none difficult to address.

Has this been tested on both Linux and OpenBSD?

build.sh
vext01: I think you are probably right about nov/dec. The file was archived on the 5th of dec.

I prefer the short format myself, but I don't mind that much (although I thought you didn't like hyphens in filenames?)

vext01: Ah, i see the date doesn't appear in the filename, just the comment.

build.sh
vext01: So you can blast the source dir and have it untarred afresh without re-fetching the tarball. I do this a lot.

build.sh
vext01: I agree.

vext01:
Yes, this was test with the warmup experiment with small parameters on both systems.

vext01:
Your comments, addressed.

ltratt:
Thanks. Let's squash then I can merge.

ltratt (approved):

vext01:
Splat

PR #290, opened 2016-12-22
Consistent table entries marked with Pexec counts.

snim2:
This PR makes a small change to the information we place in the tables. Previously we said:

  * For each benchmark:
    * Sort the categories in order of most occurrences 
    * If the most frequently occurring category came us as many times as the total number of process executions (10 in our experiment) print it on its own
    * If not, the benchmark is inconsistent, and a list of categories and their numbers of occurrences should be printed.

This missed an important case --  a consistent benchmark where one or more process execution failed with an error. I don't think we ever saw such a case, but if we had the reader would never have known. Potentially we could have lacked a lot of data and misled the reader into thinking we had no failed benchmarks at all.

This PR changes the tables slightly. Firstly we check whether we have only one category in the process executions (rather than checking the number of process executions). Secondly, we now print the number of process executions next to the sparkline, for consistent benchmarks.

### Bencher 3 table test

b3table

### DaCapo / Octane table test

other_tables

### DIFF
--- a/bin/table_classification_summaries_main
+++ b/bin/table_classification_summaries_main
@@ -88,9 +88,12 @@ def main(data_dcts, window_size, latex_file, with_preamble=False):
                 else:  # Flat execution, no changepoints.
                     time_to_steadys.append(0.0)
             # Average all information.
-            category, occurences = Counter(categories).most_common()[0]
-            if occurences == n_pexecs:
-                reported_category = STYLE_SYMBOLS[category]
+            if len(set(categories)) == 1 and len(categories) == n_pexecs:
+                reported_category = STYLE_SYMBOLS[categories[0]]
+            elif len(set(categories)) == 1:
+                # Consistent benchmark, but some process execs failed with error.
+                reported_category = ' %s\\scriptsize{($%d$)}' % \
+                                    (STYLE_SYMBOLS[categories[0]], len(categories))
             else:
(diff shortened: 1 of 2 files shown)


ltratt:
Hmm, I'm not totally sure about this to be honest. For the main paper, we won't have any errors, because we'd rather rerun the experiment than have incomplete data. So I tend to think the extra numbers just make the table harder to read. However, for other experiments that people might use the stats for, and which may not be as carefully run, I can see that reporting this *might* be useful.

However, I wonder if we should set the rule "if all *n* runs are successful, don't print out *n*; but if less than *n* are successful, then print out that number". That way we'll cover both cases (not muddying the picture for the standard case, but making the error case clear(ish) to people).

snim2:
I have made that change, I don't really have a test case for it, but I have checked that our tables come out as expected.

If you are happy with the logic, this needs squashing.

ltratt:
So just to check: for our existing data, this PR is a no-op? But (fingers crossed) if future data has errors in it then it should show up?

snim2:
Yes, our current tables don't change. Here's the PDF (you will have to ignore the Empty entry): 
[tables_numbers.pdf](https://github.com/softdevteam/warmup_experiment/files/670708/tables_numbers.pdf)

Ready to squash?

ltratt:
Yep, let's squash.

snim2:
Squashed

ltratt (approved):

PR #292, opened 2017-01-03
Fix rectangle collision algorithm.

### LINKED ISSUE #291: Improve inset placements
Our recent testing uncovered a case where insets are drawn over data (page one only):

[inset_test_case.pdf](https://github.com/softdevteam/warmup_experiment/files/670636/inset_test_case.pdf)

snim2:
Fixes #291 

### Small test case

  * Before fix: [inset_test_case_before.pdf](https://github.com/softdevteam/warmup_experiment/files/682146/inset_test_case_before.pdf)

  * After fix: [inset_test_case_after.pdf](https://github.com/softdevteam/warmup_experiment/files/682147/inset_test_case_after.pdf)

### Full run

  * Bencher 3 v0.7 data: [test_inset_b3.pdf](https://github.com/softdevteam/warmup_experiment/files/682204/test_inset_b3.pdf) (ignore the black dots for changepoints, I forgot how the CLI switches work)

### DIFF
--- a/warmup/plotting.py
+++ b/warmup/plotting.py
@@ -58,10 +58,19 @@ def collide_rect((left, bottom, width, height), fig, axis, data):
     # Next convert the bottom and top of the rect to data coordinates (seconds).
     _, inset_top = axis_data_transform(axis, 0, bottom + height, inverse=False)
     _, inset_bottom = axis_data_transform(axis, 0, bottom, inverse=False)
-    if ((bottom > 0.5 and maximum_y > inset_bottom) or  # inset at top of chart
-           (bottom < 0.5 and minimum_y < inset_top)):   # inset at bottom
-        return True
-    return False
+    if bottom > 0.5:  # inset at top of chart
+        dist = math.fabs(inset_bottom - maximum_y)
+        if maximum_y > inset_bottom:
+            return True, dist
+        else:
+            return False, dist
+    elif bottom < 0.5:  # inset at bottom
+        dist = math.fabs(inset_top - minimum_y)
+        if minimum_y < inset_top:
+            return True, dist
+        else:
+            return False, dist
+    assert False  # Unreachable.
 
 
 def add_inset_to_axis(fig, axis, rect):
--- a/warmup/plotting.py
+++ b/warmup/plotting.py
@@ -47,6 +47,11 @@ def collide_rect((left, bottom, width, height), fig, axis, data):
     x_right_float, _ = axis_data_transform(axis, left + width, 0, inverse=False)
     x_left = int(math.floor(x_left_float))
     x_right = int(math.ceil(x_right_float))
+    # Clamp x values.
+    if x_left < 0:
+        x_left = 0
+    if x_right >= len(data):
+        x_right = len(data) - 1
     # Next find the highest and lowest y-value in that segment of data.
     minimum_y = min(data[int(x_left):int(x_right)])
     maximum_y = max(data[int(x_left):int(x_right)])
--- a/warmup/plotting.py
+++ b/warmup/plotting.py
@@ -1,10 +1,70 @@
-# import numpy
+import math
+
 from matplotlib import pyplot
 from matplotlib.ticker import ScalarFormatter
 
 SPINE_LINESTYLE = 'solid'
 SPINE_LINEWIDTH = 1
 ZORDER_GRID = 1
+
+
+def axis_data_transform(axis, xin, yin, inverse=False):
+    """Translate axis and data coordinates.
+    If 'inverse' is True, data coordinates are translated to axis coordinates,
+    otherwise the transformation is reversed.
+
+    Code by Covich, from: http://stackoverflow.com/questions/29107800/
+    """
+    xlim = axis.get_xlim()
+    ylim = axis.get_ylim()
+    xdelta = xlim[1] - xlim[0]
+    ydelta = ylim[1] - ylim[0]
+    if not inverse:
+        xout =  xlim[0] + xin * xdelta
+        yout =  ylim[0] + yin * ydelta
+    else:
+        xdelta2 = xin - xlim[0]
+        ydelta2 = yin - ylim[0]
+        xout = xdelta2 / xdelta
+        yout = ydelta2 / ydelta
+    return xout, yout
+
+
+def axis_to_figure_transform(fig, axis, coord):
+    """Transform axis coordinates to figure coordinates.
+    Code by Ben Schmidt http://stackoverflow.com/questions/41462693/
+    """
+    return fig.transFigure.inverted().transform(axis.transAxes.transform(coord))
+
+
+def collide_rect((left, bottom, width, height), fig, axis, data):
+    """Determine whether a rectangle (in axis coordinates) collides with
+    any data (data coordinates, or seconds). We use the matplotlib transData
+    API to convert between display and data coordinates.
+    """
+    # Find the values on the x-axis of left and right edges of the rect.
+    x_left_float, _ = axis_data_transform(axis, left, 0, inverse=False)
+    x_right_float, _ = axis_data_transform(axis, left + width, 0, inverse=False)
+    x_left = int(math.floor(x_left_float))
+    x_right = int(math.ceil(x_right_float))
+    # Next find the highest and lowest y-value in that segment of data.
+    minimum_y = min(data[int(x_left):int(x_right)])
+    maximum_y = max(data[int(x_left):int(x_right)])
+    # Next convert the bottom and top of the rect to data coordinates (seconds).
+    _, inset_top = axis_data_transform(axis, 0, bottom + height, inverse=False)
+    _, inset_bottom = axis_data_transform(axis, 0, bottom, inverse=False)
+    if ((bottom > 0.5 and maximum_y > inset_bottom) or  # inset at top of chart
+           (bottom < 0.5 and minimum_y < inset_top)):   # inset at bottom
+        return True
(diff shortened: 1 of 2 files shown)


ltratt:
The example on p1 in the "after" case suggests we're not taking into account axis labels into the size? I don't know how easy that might be, but perhaps we can allow a fixed-side fudge factor of some sorts around the inset? [I vaguely remember we already have such a fudge factor...  if so, maybe we just need to increase it a bit?]

snim2:
Well, we make the inset a particular size, and add some padding, so certainly we can increase the padding a little. However, I would rather not mess with fudge factors in this PR, because we would need to test them against a whole data set, rather than this small test case.

ltratt:
We should test this PR against an entire data-run anyway (the current data we're collecting). That said, I don't mind doing the fudge-increasing in a separate PR.

snim2:
Yes, this is why the PR is marked "still testing" :)

snim2:
OK, full test case added

ltratt:
binarytrees/pypy pexec#8 seems to be clipping?

snim2:
It does, I think this is the same problem as the small test case where the ticklabels overlapped some data.

ltratt:
The actual inset box appears to be clipping too.

snim2:
Ugh :( I think we need to try again with something a bit more sophisticated

snim2:
The above commits remove our hacky code which tried to convert "axis coordinates" (`(0,0)` is bottom left, `(1,1)` is top-right) to and from data coordinates (in iterations / seconds). The code that replaces this uses the matplotlib transforms API to perform these calculations exactly. So, this is clearly a more sensible approach.

Unfortunately it doesn't solve the "overlaps with ticklabels" problem because ticklabels are outside the axis. So next I'll find out what API can give me the size of the ticklabels...

warmup/plotting.py
```diff
@@ -7,6 +8,51 @@
...
+    x_right_float, _ = axis_data_transform(axis, left + width, 0, inverse=False)
+    x_left = int(math.floor(x_left_float))
+    x_right = int(math.ceil(x_right_float))
+    # Next find the highest and lowest y-value in that segment of data.
+    minimum_y = min(data[int(x_left):int(x_right)])
+    maximum_y = max(data[int(x_left):int(x_right)])
+    # Next convert the bottom and top of the rect to data coordinates (seconds).
+    _, inset_top = axis_data_transform(axis, 0, bottom + height, inverse=False)
+    _, inset_bottom = axis_data_transform(axis, 0, bottom, inverse=False)
+    print '(min_y, max_y):', minimum_y, maximum_y, '(inset_top, inset_bottom)', inset_top, inset_bottom
```
ltratt: Debugging left in?

ltratt:
Sounds good to me!

snim2:
**Small test case:** [inset_test_case.pdf](https://github.com/softdevteam/warmup_experiment/files/696216/inset_test_case.pdf)

**Larger test case:** [test_inset_b3.pdf](https://github.com/softdevteam/warmup_experiment/files/696217/test_inset_b3.pdf)

ltratt:
OK, this is looking really good, and is mergeable as-is.

I have one question/suggestion: if it's too difficult, don't worry about it. At the moment, we start in the top-right, and if there's clipping, move things until we find the first place that doesn't clip. Sometimes the first place, however, isn't the best place. For example, if you look at binarytrees/pypy/#9, it would arguably be best if if the inset was close to the top-left rather than close to the top-right. Basically, I guess one would slightly tweak the algorithm to say "try all positions; at each position, if the 'score' (i.e. how far away from the data) is better than the previous best, take that as the new best; at the end put the inset at the best position." This may be trickier than I imagine, but I thought I'd run it by you.

snim2:
Let me have a go and see. This will increase the running time quite a bit though.

ltratt:
OK. If it's too slow, maybe it's a non-runner (or only accessible via a switch).

snim2:
Going by the current time (written into my BASH prompt) the previous version ran on bencher2 from 13:02ish--13:33, this version ran from 15:10--15:41.

**Small test case:** [inset_test_case.pdf](https://github.com/softdevteam/warmup_experiment/files/696563/inset_test_case.pdf)

**Large test case:** [test_inset_b3.pdf](https://github.com/softdevteam/warmup_experiment/files/696566/test_inset_b3.pdf)

ltratt:
It got 1 minute faster then ;) OK, so performance doesn't look like an issue!

snim2:
I expect I didn't press return before running it the first time, but the difference can't be very great.

ltratt:
OK, so this pretty much fixes the issue, which is good. I think we should try one minor tweak and then we're probably done: at the moment, very tiny differences in the distance to data can mean the inset dances around like a monkey on a hot tin roof. binarytrees/c is a very good example of this. Maybe we should add a small tolerance that says "unless position X is at least Y% better than position Z, stick with Z"? So the logic would be more like:

```python
FUDGE = 1.05 # 5%

best_dist = None
for left_offset, bottom in zip(...):
    inset_collides, dist = collide_rect(...)
    if best_dist is None or dist[1] >= best_dist[1] * FUDGE:
        best_dist = dist
```

[This assumes, as I think is the case, that we always try positions in a fixed order.]

Warning: this may be a bad idea!

snim2:
With tweaks.

**Small test case:** [inset_test_case.pdf](https://github.com/softdevteam/warmup_experiment/files/696902/inset_test_case.pdf)


**Large test case:** [test_inset_b3.pdf](https://github.com/softdevteam/warmup_experiment/files/696899/test_inset_b3.pdf)

ltratt:
This looks good. One question that I can't answer though. Consider binarytrees/pypy. Why are some of the insets (e.g. #1) on the RHS and close to data and some a long way to the left and far away from data? I'm more curious than anything!

snim2:
I think because sometimes the difference in distances is just less than the threshold (1%). I've added some rounding, and this is the result for the binary trees charts only: [bintrees.pdf](https://github.com/softdevteam/warmup_experiment/files/696993/bintrees.pdf)

Below is the debug output from binarytrees/PyPy. What you are seeing here is, for each pexec, a list of tuples containing the distance between the data and inset and the difference between that distance and the current "best" distance. On the next line the final "best" distance is printed. If there are fewer than 10 distances, this is because some potential rectangles overlapped with data: 

```
Plotting Linux$_\mathrm{4790K}$: Binary Trees (PyPy) on page 07 of 08.
10 plots arranged in 5 rows and 2 columns.
(0.0023963397119999286, 1.0) (0.005204339711999961, 0.0) (0.009282339711999876, 0.01) (0.012330339711999927, 0.01)
0.002396339712

(0.0017223397119999762, 1.0) (0.004354339711999944, 0.0) (0.008262339711999966, 0.01) (0.01133433971199993, 0.01)
0.001722339712

(0.008181339711999969, 1.01) (0.013310339711999908, 0.01) (0.015117339711999911, 0.01) (0.016942339711999876, 0.01) (0.019042339711999978, 0.01) (0.020569339711999923, 0.01)
0.008181339712

(0.004969339711999976, 1.0) (0.010829339711999952, 0.01) (0.012630339711999894, 0.01) (0.014324339711999978, 0.01) (0.016194339711999906, 0.01) (0.018089339711999886, 0.01)
0.004969339712

(0.005792339711999883, 1.01) (0.009447339711999958, 0.0) (0.011966339711999896, 0.01) (0.013334339711999932, 0.01) (0.01597833971199991, 0.01) (0.016857339711999875, 0.01)
0.005792339712

(0.0023683397119999006, 1.0) (0.0055233397119999195, 0.0) (0.008454339711999936, 0.01) (0.01250133971199996, 0.01)
0.002368339712

(0.0051993397119999285, 1.01) (0.00692733971199988, 0.0) (0.008834339711999983, 0.0) (0.010965339711999977, 0.01) (0.012431339711999945, 0.01)
0.005199339712

(0.003437339711999887, 1.0) (0.007204339711999963, 0.0) (0.0019062733440000645, -0.0)
0.003437339712

(0.0003503397119999363, 1.0) (0.0014843397119999047, 0.0) (0.004579339711999975, 0.0) (0.008232339711999881, 0.01) (0.01058533971199993, 0.01)
0.000350339712

(0.005537339711999878, 1.01) (0.010992339711999977, 0.01) (0.01120833971199997, 0.01) (0.014096339711999972, 0.01) (0.016547339711999953, 0.01) (0.0180743397119999, 0.01)
0.005537339712
```

ltratt:
OK, so I guess let's do this on the full run, double-check that it looks OK, then we can merge (after rebasing against master)?

snim2:
Yup, plotting away on b2 now...

ltratt:
Cool! We're almost there :)

snim2:
**Small test case:** [inset_test_case.pdf](https://github.com/softdevteam/warmup_experiment/files/697096/inset_test_case.pdf)

**Large test case:** [test_inset_b3.pdf](https://github.com/softdevteam/warmup_experiment/files/697098/test_inset_b3.pdf)

ltratt:
We have a winner! You need to do a rebase against master. Do you want also want to squash at the same time? Either way, let me know when you're done, and then I can merge.

snim2:
I'll squash first...hold on to your hat...

snim2:
@ltratt done!

ltratt:
One tweak: the last commit has a weird name "Place inset as from data as possible". Missing a word?

ltratt (approved):

ltratt:
Thanks!

Issue #288, opened 2016-12-20
Make clear the timings of segments


ltratt:
At the moment we often don't really know what the segments on plots are, because the y-axis is so zoomed out that we can't see what's going on. This means that we sometimes appear to have a flat line but with changepoints, which can be hard to understand and defend.

I have a couple of vague ideas for this. First, on the plot itself, we could (above each segment) print out its mean (and variance?). Second, we could plot an inset for the *final* changepoint in a plot (or, if it's a slowdown, maybe the changepoint that causes that?).

snim2:
Ah yes, I hadn't forgotten this, but it isn't altogether simple. The problem with annotating segment means in the plots is that it is possible to have very short segments, which would not leave enough space to show any annotated text. In most cases this is not true of the last segment (but it might be true in a *no steady state* benchmark). 

Adding some more data in the insets sounds more plausible, but that will make them larger, so there may be more charts where we can't show the inset at all.

ltratt:
If a segment's really small and inconsequential, I guess that not showing the times is irrelevant. Maybe we just need to show the "important" segment's timings?

ltratt:
OK, so I think there are probably two aspects to this issue:

  1) Finding the important changepoint/changepoints: there's no point showing the user information that isn't going to affect their understanding of a plot. For flat, there are no interesting changepoints (easy!). For warmup/slowdown, there might be one or two interesting changepoints: if the steady-state segment has a mean "much" different than all other segments, then just showing that is fine; if the mean of that segment is similar to another segment, then showing both of them is probably necessary to understanding them.

  2) Working out how to display this information to the user. We could do a second inset (but do we sometimes need a third? or, for "interesting" cases, maybe we drop the current "first" inset?). Or we can add timing information (means? means and variance?) directly onto the plot. Or ...?

snim2:
This was something we used for our down debugging and testing, and then removed from the repo. I think we can safely close this issue, as we did not need the requested feature "in production".

PR #294, opened 2017-01-04
Annotate segment means

snim2:
This is a simple first step towards resolving #288 

Whenever a warmup or slowdown process execution has > 1 changepoint, the last and penultimate segments are annotated with their means, drawn either side of the last changepoint. 

### Small test case

[issue_288_test.pdf](https://github.com/softdevteam/warmup_experiment/files/685946/issue_288_test.pdf)

### Larger test case (v0.7)

[issue_288_test_large.pdf](https://github.com/softdevteam/warmup_experiment/files/685947/issue_288_test_large.pdf)

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -425,7 +425,7 @@ class ProcessExecChart(object):
         self.minor_xticks = compute_grid_offsets(self.x_bounds[0], self.x_bounds[1], GRID_MINOR_X_DIVS)
         # Shared information for segment annotations.
         self.annotations_x_padding = (x_bounds[1] - x_bounds[0]) * 0.01
-        self.annotations_y_padding = (y_range[1] - y_range[0]) * 0.02
+        self.annotations_y_padding = (y_range[1] - y_range[0]) * 0.1
         # Calculate the number of plots needed.
         self.n_rows = 1
         if cycles_data:
@@ -581,7 +581,9 @@ class ProcessExecChart(object):
                               zorder=ZORDER_CHANGEPOINTS)
 
     def _plot_segment_annotations(self, axis):
-        """Annotate axis with segment means as text."""
+        """Annotate axis with segment means as text.
+        TODO: Check the user has passed in --with-changepoint-means!
+        """
         if self.changepoints:
             for cpt in xrange(len(self.changepoint_means)):
                 if cpt == 0:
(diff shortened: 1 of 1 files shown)


ltratt:
For this simple version, I think we might as well just plot the means everywhere (there's no use in restricting it to warmup/slowdown). Even with this subset, we can clearly see that there are some things worthy of our attention. For example: in the small test case, jython/graal/#3 if the "merge" threshold was 0.005s (not 0.001s) this would be a warmup. That, and a few examples, are suggestive that our original threshold of 0.001s is probably too conservative (which I suspected before, but couldn't really confirm either way).

I think we'll need a nicer way of visualising than this in the long run, but this is definitely very helpful as a first step! Let's first plot the means everywhere, look at the data, then we might try bumping up the merge threshold and replotting, and see where that gets us.

[BTW, I think you might have created this branch from that for PR #292. We can either wait until we merge 292, or you might want to rebase the relevant commit against master and force push against this branch.]

snim2:
I'm not so sure about plotting the means everywhere. Many plots have changepoints that occur closely together, which means that the mean will be obscured by the changepoint line and maybe other artists.

ltratt:
At the moment there are many examples where I want to see what's going on, but there are no numbers. I'd prefer too many than too little, at least while we're trying to understand our data better.

snim2:
Try this: [issue_288_test.pdf](https://github.com/softdevteam/warmup_experiment/files/686073/issue_288_test.pdf)

ltratt:
I'm confused: this doesn't annotate all segments even on page 1?

ltratt:
[For example, I *really* want to see the segment times on fop/graal.]

snim2:
Warmups / slowdowns with only one changepoint are deemed "uninteresting", the Fop/Graal charts are "no steady state", so not annotated at all.

ltratt:
At the moment we're debugging, so let's please annotate the no steady states so we can see what's going on.

ltratt:
[Even with the limited annotations, it's clear that we need to think more carefully about when to merge segments. We can get more clever later with the "how to present this to everyone else", but at the moment I think we really need to see *all* the segment means.]

snim2:
Next: [issue_288_test.pdf](https://github.com/softdevteam/warmup_experiment/files/687148/issue_288_test.pdf)

snim2:
About to squash this in order to bring it up to date with master...

snim2:
This branch now has the improved inset placements from master.

snim2:
Bringing this up to date with master again...

snim2:
This PR and branch was used for our debugging, and can probably be closed / deleted before we split this repo.

ltratt:
I agree -- closing.

PR #300, opened 2017-01-16
Adding a "zoomed in" plot to the stack

snim2:
This is an experiment in trying to explain the changepoints a bit better, if people like it there will need to be a little more work to make this merge-able.

The idea is to plot the measurement data twice: once with the unified y-range, and once with a unified y-range that cuts-off any "slow iterations" (i.e. as the data would look if the first N iterations were not there, or the iterations before the first changepoint). The idea is simply to give the user a clearer view of the fast iterations, for the whole run-sequence i.e. not just the few iterations we can fit in the inset.

Here is how the result looks:

  * Test one: [test_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/709363/test_zoomey.pdf)
  * Test two: [test_zoomey_dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/709365/test_zoomey_dacapo.pdf)

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -23,7 +23,7 @@ from warmup.krun_results import pretty_print_machine, read_krun_results_file
 from warmup.outliers import get_window
 from warmup.plotting import add_inset_to_axis, add_margin_to_axes, axis_to_figure_transform
 from warmup.plotting import collide_rect, compute_grid_offsets, format_yticks_scientific
-from warmup.plotting import get_unified_yrange, style_axis
+from warmup.plotting import get_unified_yrange, min_no_outliers, max_no_outliers, style_axis
 from warmup.vm_instruments import INSTRUMENTATION_PARSERS
 
 seaborn.set_palette('colorblind')
@@ -93,6 +93,9 @@ ZORDER_MARKERS = 20
 # Gap between right-hand y-axes.
 TWINX_OFFSET = 1.1
 
+# Padding for unified y-limits.
+Y_LIM_PADDING = 0.002
+
 # Use a fixed y-range for the aperf/mperf data.
 PERF_YRANGE = (0.0, 1.2)
 PERF_YTICK_FORMAT = '%.1f'
@@ -105,6 +108,11 @@ INSET_DIST_DELTA = 0.01
 INSET_MIN_ITERS = 50
 INSET_MAX_ITERS = 150
 INSET_TICK_FONTSIZE = 8
+
+# Zoomed in plot of measurement data.
+ZOOM_PROPORTION = 0.02
+ZOOM_EXTRA_Y_LIM_PADDING = 0.1
+ZOOM_OUTLIER_SIZE = 8
 
 # Default (PDF) font sizes
 TICK_FONTSIZE = 8
@@ -171,6 +179,7 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
     instr_pages = list()
     all_outliers, all_common, all_unique = list(), list(), list()
     all_changepoints, all_changepoint_means = list(), list()
+    all_classifications = list()
 
     # By default, each benchmark from each machine is placed on a separate
     # page. If the --one-page switch has been passed in, then we place
@@ -182,6 +191,7 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
         instr_page = list()
         page_common, page_unique, page_outlier = list(), list(), list()
         page_changepoints, page_changepoint_means = list(), list()
+        page_classifications = list()
         for key in sorted(data_dcts['data']):
             for machine in sorted(data_dcts['data'][key]):
                 for index, run_seq in enumerate(data_dcts['data'][key][machine]):
@@ -208,12 +218,15 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
                     if changepoints:
                         page_changepoints.append(data_dcts['changepoints'][key][machine][index])
                         page_changepoint_means.append(None)
+                        page_classifications.append(data_dcts['classifications'][key][machine][index])
                     if changepoint_means:
                         page_changepoints.append(data_dcts['changepoints'][key][machine][index])
                         page_changepoint_means.append(data_dcts['changepoint_means'][key][machine][index])
+                        page_classifications.append(data_dcts['classifications'][key][machine][index])
                     else:
                         page_changepoints.append(None)
                         page_changepoint_means.append(None)
+                        page_classifications.append(None)
         pages.append(page)
         cycles_pages.append(cycles_page)
         instr_pages.append(instr_page)
@@ -223,6 +236,7 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
         all_unique.append(page_unique)
         all_changepoints.append(page_changepoints)
         all_changepoint_means.append(page_changepoint_means)
+        all_classifications.append(page_classifications)
     else:  # Create multiple pages.
         for key in sorted(data_dcts['data']):
             for machine in sorted(data_dcts['data'][key]):
@@ -243,12 +257,15 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
                 if changepoints:
                     all_changepoints.append(data_dcts['changepoints'][key][machine])
                     all_changepoint_means.append(None)
+                    all_classifications.append(data_dcts['classifications'][key][machine])
                 elif changepoint_means:
                     all_changepoints.append(data_dcts['changepoints'][key][machine])
                     all_changepoint_means.append(data_dcts['changepoint_means'][key][machine])
+                    all_classifications.append(data_dcts['classifications'][key][machine])
                 else:
                     all_changepoints.append(None)
                     all_changepoint_means.append(None)
+                    all_classifications.append(None)
 
     # Draw each page and display (interactive mode) or save to disk.
     try:
@@ -280,14 +297,15 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
             unique = only_uncrashed(all_unique[index])
             changepoints = only_uncrashed(all_changepoints[index])
             changepoint_means = only_uncrashed(all_changepoint_means[index])
+            classifications = only_uncrashed(all_classifications[index])
 
             fig = draw_page(is_interactive, wct_page, cycles_page,
                                          instr_page, subplot_titles,
                                          window_size, xlimits, outliers,
                                          unique, common, changepoints,
-                                         changepoint_means, median, tukey,
-                                         inset, legend_off, core_cycles,
-                                         cycles_ylimits)
+                                         changepoint_means, classifications,
+                                         median, tukey, inset, legend_off,
+                                         core_cycles, cycles_ylimits)
             if fig is not None:
                 if not is_interactive:
                     pdf.savefig(fig, dpi=fig.dpi, orientation='landscape',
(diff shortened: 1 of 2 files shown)


ltratt:
This is an interesting idea and worthy of exploration. Can we maybe try it with a bigger dataset to see how effective it is in more cases?

vext01:
I think this makes the second example much clearer.

I was looking at how this works. You make a new pair of axes with a different y-scale:
```
y_min_50, y_max_50 = get_unified_yrange(executions, 50 , xlimits_stop, padding=0.02)
```

So at the moment the range used to make the second ylimit is fixed at `[50:]` is that intentional? Or do we need something more sophisticated. I suppose some vms will take > 50 iterations to get to faster iterations?

I guess this would not account for slow iterations later than iter #50.

Thoughts?

snim2:
I think we do need something more sophisticated, the 50 was just a quick hack to get something together that you could look at.

Right now I'm thinking that we will need to consider each pexec on the page separately:

  * If *all* pexecs on a page are "flat", then there is no point in having the extra axis at all
  * Otherwise, for "warmup" / "slowdown" pexes we will usually want the new axis to start from the first changepoint. This might not catch cases where fast iterations start two or three changepoints in
  * If some pexecs are "flat" or "no steady state", but not all, then for those we will need a fixed number of iterations and 50 is probably fine.

So we will need to use those rules to create the set of data to pass to `get_unified_yrange()`.

Does that sound sensible?

vext01:
Sounds sensible.

By the way, above you said that the y-axes for the zoomed plots would be unified. I wonder if we want those independent? I can see pros and cons of both approaches though.

snim2:
This is a good point. Non-unified would probably be a little easier to implement, so maybe I'll start with that and we can see how it looks. Once we have charts that look sensible, there will be a couple of CLI issues to think about.

snim2:
v0.7 Bencher 3: [bencher3_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/711201/bencher3_zoomey.pdf)

ltratt:
binarytrees/v8 is interesting -- are we using the segment mean as the start of the y axis?

snim2:
We are using the iteration after the first changepoint, and taking the min / max from there to the end of the run sequence. Personally, I think that isn't "zoomed in" enough for many of the pages, but I'm not sure what would be better.

vext01:
I noticed that some of the zoomed plots chop of the line close to the bottom of the y axis:

bot

snim2:
Ah, I think you mean there needs to be some padding at the top and bottom of the y-axis?

ltratt:
Yep (I also agree that the zooming isn't enough in most cases, and that it's hard to work out what to do).

vext01:
@snim2 yeah.

@ltratt Can you give an example?

snim2:
@vext01 The PNG you posted is a good example, most of the data should be around the middle of the y-axis, but in fact most is at the bottom

vext01:
I mean an example of where zooming isn't enough.

snim2:
Yup, that's an example of where zooming wasn't sufficient. Also, the last page (SpectralNorm / V8) is one where outliers have made the y-limits far too wide, so removing outliers is clearly one thing we need to do.

vext01:
Ah I see what you mean now.

Shouldn't the y-axis on the zoomed graph have been determined using only the data in the second segment? Looks like it should work, no?

snim2:
Ah, I see we should have `y_zoom_min = min(executions[index][first_n:])  - Y_LIM_PADDING` not `y_zoom_min = min(executions[index][first_n:])  + Y_LIM_PADDING`

snim2:
Next round of test cases:

  * Small test case: [test_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/712386/test_zoomey.pdf)
  * v0.7 da capo test case: [test_zoomey_dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/712389/test_zoomey_dacapo.pdf)
  * Full 0.7 bencher 3 run: [bencher3_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/712488/bencher3_zoomey.pdf)


If that looks sensible, we need to make some decisions about how this new plot should be invoked from the command line. 

  * Should `--wallclock-only` include this plot (matter for the paper and presetations)?
  * If not, will we sometimes want to draw the zoomed chart but *not* the core cycles data?

ltratt:
On the small test case, why do the 3rd and 4th page not have zoomey graphs? [There are other cases in the other files too.]

Perhaps a bigger issue is that sometime the "zoomed in" graphs are actually less "zoomed in" than the main graphs! That's not very useful, so I wonder if we can say something like "if our first attempt at zooming in would actually lead to a zoom out, halve the scale again"?

snim2:
In the small test case, pages 3 and 4 have all their process execs classified as flat, if *every* chart on a page if flat then the "unified" y-limits will already be zoomed in reasonably well, so there doesn't seem to be much point in adding the extra chart to the stack.

Can you give an example where the zoomed in charts are "less zoomed in" than the main plot? That shouldn't be possible...

ltratt:
Look at binarytrees/c for a zoomed out example: because the zoomed in graph is much smaller (~20%?) of the main graph, you can get this odd effect.

I sort-of understand the flat thing... but it's *really* confusing to sometimes see them and sometimes not. I'd suggest, for simplicity's sake we always put them in.

snim2:
I'm not sure I follow you about the zoomed in charts, the y-limits of the zoomed in chart are always smaller than the main plots. The binary-trees/C example is a no-steady state, and really it's the warmup runs we are more interested in, because that is where the "fast" iterations will contain detail you can't see on the main chart. If you look at the last few pages of the bencher 3 example there is a lot of detail there that wouldn't otherwise be visible.

ltratt:
In many cases the zooming works.

But by "zoomed out" I mean I get *more* info from the main graph than the zoomed in one. Crudely put, if any of the peaks in the "zoomed in" graph are smaller (in pixels) than the "zoomed out" graph, then really the "zoomed in" graph isn't zoomed in. On binarytrees/c, the "bouncing ball" is more pronounced in the main graph than the zoomed in one: therefore the latter isn't adding any useful information at all.

snim2:
That's true, but that's only because there aren't any non-outlier iterations that are much larger than the majority of the data. We could say "don't draw the zoomed in chart for no-steady state", but then we will miss some cases where it is useful.

I'm not sure there is a way of defining the y-limits so that you would get more information on a flat process exec or the binary trees / C example, in those cases the zoomey graph just isn't really necessary.

ltratt:
Well, in this particular case, we could try and zoom in much tighter on the segment means? That would be moderately useful (or, at least, better than what we currently have). I don't know if that will always do the trick though.

snim2:
The y-limits of the zoomed in graph show all the data (after the window size) that aren't outliers, so if we do zoom in any more, then there will be data that doesn't fit on the chart at all. If you think that's worth trying then I'll have a go tomorrow, but I would think that in most cases we will lose detail.

ltratt:
Yes, we will definitely lose detail (that happens in some graphs already anyway), but I think it's worth trying tomorrow. Thanks!

snim2:
In this set of tests, the y-max/min is defined by the 75% and 25% percentile of the data (minus outliers):

  * Small test case: [test_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/713537/test_zoomey.pdf)
  * v0.7 da capo test case: [test_zoomey_dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/713540/test_zoomey_dacapo.pdf)
  * Full 0.7 bencher 3 run: [bencher3_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/713542/bencher3_zoomey.pdf)

ltratt:
I wonder if we can have a mix of this new, and the old, method? Basically, when the old zoomey method worked, it worked well; but when it wasn't useful (e.g. binarytrees/c), maybe we can do something like the new method?

vext01:
I think we have zoomed in too far. Some of these plots are almost a black rectangle. Should we try 90/10 percentile instead?

As for CLI user interface. I suppose we could begin by including these plots unless `--no-zoom-plot` is passed?

snim2:
OK, so does `--wallclock-only` mean both the wallclock plots or just the main plot?

ltratt:
Let's not worry about `--wallclock-only` just yet: let's see if we can get the zoomed in plots to the point whether no-one in their right mind would want to turn them off.

vext01:
I dont want these plots in my slides, so I do want the option to omit them.

I suppose since the zoomed plot is a wallclock plot, `--wallclock-only` would include them.

snim2:
This batch of test cases shows the change @vext01 asked for (percentiles changed to 90/10), not the change @ltratt asked for (different y-limit algorithms for different charts), which will follow later:

  * Small test case: [test_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/713800/test_zoomey.pdf)
  * Da Capo test case: [test_zoomey_dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/713802/test_zoomey_dacapo.pdf)
  * Bencher 3 test case: [bencher3_zoomey.pdf](https://github.com/softdevteam/warmup_experiment/files/713804/bencher3_zoomey.pdf)

vext01:
This looks good to me. And I think it certainly does help to analyse the plots.

E.g.:

zoom

Using the zoomed out plot i never would have noticed these missed segments.

ltratt:
Have we ever worked out why we sometimes get y axes like "8.06×10^-5+4. 6503×10^-1" (test_zoomey.pdf p4)?

ltratt:
[Maybe I'm dumb, but I can't translate those axes into seconds easily.]

snim2:
We use the same formatting algorithm for the "main" chart and the "zoomey" chart, which uses (LaTeX-ified) scientific formatting, to avoid having the exponent above the top left corner of the axes. I'm not sure why there are two bits added together though.

ltratt:
Personally, I would *much* prefer all times to be plain x.yz (imho, 0.0000001 is better than 10^-x).

snim2:
In which case, we would have to decide how many decimal places is reasonable to print, otherwise we will be back to having "0.0" or similar on every y-tick. I'm not sure how many we would need to differentiate between the ticks, in the core cycles we sometimes get `7. 944x10^10` etc.

ltratt:
For core cycles, I don't think the scale matters much at all (I could even live with having no numbers on the y axis at all). I would have thought 4DP (maybe 5DP at the outside) would be fine for the others. If we're zooming in beyond that point, it's probably meaningless. [Maybe we should have a "break" on the zooming to make sure it doesn't go silly-far?]

ltratt:
OK, I agree: the zooming is now useful in pretty much every context, and I'm happy to merge things.

I don't know if we want to fix the axes thing in this PR or another.

And another vaguely related thing: I think the inset should have changepoints on it to match the main and zoomy plots.

snim2:
OK, in that case, I will make the changes to the CLI options, and raise the other two as issues for the next PRs.

snim2:
The CLI changes have been made, so I'll squash this and bring it up to date with master to make the final review simpler.

ltratt:
I'm not sure we need to squash this before review.

snim2:
OK, I'll wait till after. I don't have any further commits to make unless they are requested with the review.

ltratt:
Unless @vext01 objects, I think you should simply squash this as you see fit, and then I should merge it. [Best wait for Edd to confirm that he's OK with it though!]

vext01:
New CLI arg looks fine to me.

snim2:
OK to squash?

vext01:
sure

snim2:
Git hub is not telling me whether this is merge-able. Hmm...

ltratt (approved):

ltratt:
It is mergeable. OK for me to do so?

snim2:
If you can, yes, but I would have thought I need to bring this up to date with master?

ltratt:
I guess you rebased against master, so you implicitly did so? Anyway, GitHub can do a merge even if something hasn't been rebased, provided that there are no conflicts.

snim2:
Thanks. I'm amazed there are no conflicts!

PR #305, opened 2017-01-18
Dacapo better timer

ltratt:
This commit a) changes DaCapo to use a better-than-millisecond timer b) unbreaks Graal when running DaCapo.

Here's a (very small) subset of the output, so that we can see that it's working:

```
0,avrora,3.00826257,1.750913261,1.667840536,1.589991021,1.587320767,1.584059504,
1.569380714,1.565149822,1.559371238,1.582767391,1.580639695,1.583717649,1.572852
318,1.576273995,1.585546299,1.556570193,1.583217447,1.562401111,1.584982765,1.58
9988128,1.566040396,1.600723317,1.560411806,1.571652799,1.582280081,1.564775326,1.566714208,1.570006142,1.578958131,1.573332607
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -14,6 +14,7 @@ check_for () {
     fi
 }
 
+check_for ant
 check_for cc
 check_for g++
 check_for bunzip2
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -639,12 +640,30 @@ build_spidermonkey() {
...
 
-fetch_dacapo_jar() {
-    echo "\n===> Download DaCapo .jar file\n"
+build_dacapo() {
+    echo "\n===> Build DaCapo\n"
 
     if [ -f "${HERE}/extbench/dacapo-9.12-bach.jar" ]; then return; fi
 
-    wget "http://downloads.sourceforge.net/project/dacapobench/9.12-bach/dacapo-9.12-bach.jar?r=https%3A%2F%2Fsourceforge.net%2Fprojects%2Fdacapobench%2Ffiles%2F&ts=1474888492&use_mirror=freefr" -O ${HERE}/extbench/dacapo-9.12-bach.jar  || exit $?
+    cd ${wrkdir}
```
vext01: Can we add a brief comment saying why we build it and then fetch a binary distribution. You could even copy/paste from the commit message which explains nicely.

vext01 (changes requested):
Other than that one comment, this seems OK to me.

build.sh
ltratt: We haven't done that elsewhere in the file TBH...

vext01: I've added verbose comments when either: a) there's an important detail that will be forgotten when updating the VM, or b) something is non-obvious. I think this falls into the latter? No?

ltratt:
Point taken. See if that commit helps.

vext01:
Perfect.

Would you like to squash?

ltratt:
Squashed.

vext01:
Oh, do we need to do the same arguments change for graal/octane btw?

ltratt:
Octane doesn't use Graal. [I am going to double check Octane anyway, but I think it less likely that anything needs fixing, and if it does, it'll be in a separate PR.]

vext01 (approved):

PR #311, opened 2017-01-18
Improve y-tick labels

### LINKED ISSUE #306: Scientific ticklabels are too complicated
The y-tick formatter applies scientific formatting to y-axis of a given set of axes, it changes the ticklabels from 'offset' notation (where a number is added / subtracted to each ticklabel) to the more intuitive scientific notation (where a number multiplies each ticklabel). It then removes the multiplier from the top-left hand corner of the plot, and add it to the ticklabel instead.

However, we sometimes get very complex ticklabels like this:

`0. 046 x 10^4 +9. 299x10^-1`

which are too complicated.

snim2:
Fixes #306 

### Before / after tests

See page 3:

  * Before: [test_ticks_before.pdf](https://github.com/softdevteam/warmup_experiment/files/714572/test_ticks_before.pdf)
  * After: [test_ticks.pdf](https://github.com/softdevteam/warmup_experiment/files/714573/test_ticks.pdf)

See page 52:

  * Before: [bencher3_test_ticks_before.pdf](https://github.com/softdevteam/warmup_experiment/files/714605/bencher3_test_ticks_before.pdf)
  * After: [bencher3_test_ticks_after.pdf](https://github.com/softdevteam/warmup_experiment/files/714735/bencher3_test_ticks_after.pdf)

### Other test cases

  * Da capo: [test_ticks_dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/714571/test_ticks_dacapo.pdf)

### DIFF
--- a/warmup/plotting.py
+++ b/warmup/plotting.py
@@ -1,7 +1,7 @@
 import math
 
 from matplotlib import pyplot
-from matplotlib.ticker import ScalarFormatter
+from matplotlib.ticker import ScalarFormatter, FormatStrFormatter
 
 SPINE_LINESTYLE = 'solid'
 SPINE_LINEWIDTH = 1
@@ -83,17 +83,21 @@ def format_yticks_scientific(axis):
     multiplies each ticklabel). Remove the multiplier from the top-left hand
     corner of the plot, and add it to the ticklabel instead.
     """
-    formatter = ScalarFormatter(useMathText=True)
-    formatter.set_scientific(True)
-    formatter.set_powerlimits((-2, 2))
-    axis.yaxis.set_major_formatter(formatter)
-    y_axis = axis.yaxis
+    y_min, y_max= axis.get_ylim()
+    if y_max > 100000 or (y_min < 0.00001 and y_min > 0.0):
+        formatter = ScalarFormatter(useMathText=True, useOffset=False)
+        formatter.set_scientific(True)
+        formatter.set_powerlimits((-6, 6))
+        axis.yaxis.set_major_formatter(formatter)
+    else:
+        formatter = FormatStrFormatter('%.5f')
+        axis.yaxis.set_major_formatter(formatter)
     pyplot.draw()
-    offset = y_axis.get_offset_text().get_text()
+    offset = axis.yaxis.get_offset_text().get_text()
     if len(offset) > 0:
         labels = [label.get_text() + offset for label in axis.get_yticklabels()]
         axis.set_yticklabels(labels)
-        y_axis.offsetText.set_visible(False)
+        axis.yaxis.offsetText.set_visible(False)
 
 
 def get_unified_yrange(executions, xlimits_start, xlimits_stop, padding=0.02):


ltratt:
Two obvious questions. First, what does "most" mean ;) Second, can we fix the number of DP? On the "small" example, the DP for the main plots is 5, 4, 7 (!), 6 for pages 1, 2, 3, 4 respectively. I would like to think we could get away with 4DP, but 5DP is probably the maximum we can show without people laughing at us. [Some of the timers / float representations (e.g. Octane/V8) are only accurate to 6DP anyway it seems. I don't know about other VMs though.]

snim2:
Re "many": I have turned "offsets" off. Offsets are the text that read 10^... that used to appear on the top-left of the axes. I would have thought that turning off offsets would turn them off, in fact it turns them off *sometimes*. So I guess that turning them on meant "please have an offset even if it is completely unnecessary to do so".

ltratt:
Erk. OK, I guess we can see if that problem does actually happen in practise. Maybe we'll get lucky!

If we can sort out the DP thing, then I think we'll be good to merge.

snim2:
There isn't a way to do what you want. This next round of test cases uses the following:

  * If the ticklabels are reasonably small (i.e. they are likely to fit in 5sf) then use a simple string formatter
  * Otherwise use the more complex formatter with the offset

Potentially we might still get a case where the formatter gives us a complex offset, but this change should make that very unlikely.

  * Small test: [test_ticks_after.pdf](https://github.com/softdevteam/warmup_experiment/files/716303/test_ticks_after.pdf)
  * Da capo test: [test_ticks_dacapo.pdf](https://github.com/softdevteam/warmup_experiment/files/716304/test_ticks_dacapo.pdf)
  * Bencher 3 test: [bencher3_test_ticks_after.pdf](https://github.com/softdevteam/warmup_experiment/files/716306/bencher3_test_ticks_after.pdf)

ltratt:
If I look at the small & dacapo files, everything is 5DP except xalan and fop which are 3DP. Any ideas why?

snim2:
Try this: [xalan.pdf](https://github.com/softdevteam/warmup_experiment/files/716446/xalan.pdf)

ltratt:
Aha, perfect! Do we need to check it on all the others, or are we fairly sure that everything that reasonably can be 5DP will be 5DP?

snim2:
The problem was caused by a condition: `if y_max > 100000 or y_min < 0.00001:` was being triggered to use the scientific formatting because `y_min` was `0.0`. The condition is now:

```
    if y_max > 100000 or (y_min < 0.00001 and y_min > 0.0):
```

so other cases should not be affected.

ltratt:
Is that a "yes"? ;)

snim2:
Yes, the other pages should be unaffected.

ltratt:
Excellent! OK, squash as you see fit, then I can merge.

snim2:
Squashed.

ltratt:
Thanks!

ltratt (approved):

PR #313, opened 2017-01-19
The user may choose not to merge adjacent segments. 

snim2:
This PR allows the user to choose not to merge adjacent segments with similar means. This is something I think we might want to use in our comparison between different deltas.

### DIFF
--- a/bin/mark_changepoints_in_json
+++ b/bin/mark_changepoints_in_json
@@ -224,7 +224,8 @@ def main(in_files, steady_state, delta):
                 else:
                     outliers = list()
                 segments = get_segments(cpt, delta, p_exec, steady_state, outliers)
-                segments.merge_overlapping_segments()
+                if delta > 0.0:
+                    segments.merge_overlapping_segments()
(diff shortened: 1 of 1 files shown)


vext01:
Looks good to me :)

ltratt:
I may be overly-ignorant of floats, but can we just say `< 0` and `> 0` at the appropriate places? Or is `0.0` definitely needed?

snim2:
I always use a float to ensure that the reader knows the type of the variable, not for any clever reason about the semantics of floats. In this case I think we can assume that DELTA will not be sub-normal or require a fuzzy comparison.

ltratt:
OK, that makes sense in the code. In the doc comment, however, I think we would be better saying just "less than 0" (.0 is a bit confusing IMHO).

snim2:
Good to squash?

ltratt:
Squash away!

snim2:
Done!

ltratt (approved):

PR #316, opened 2017-01-20
Ensure the classification key is scaffolded in data dict.

### LINKED ISSUE #303: Crash in plot script using -b
```
$ python2.7 ../bin/plot_krun_results -b bencher3:fasta:Graal:default-java:0 -o out/10_classic1.pdf warmup_results_0_7_linux1_i7_4790k.json.bz2 warmup_results_0_7_openbsd1_i7_4790.json.bz2
/home/vext01/.local/lib/python2.7/site-packages/matplotlib/__init__.py:878: UserWarning: axes.color_cycle is deprecated and replaced with axes.prop_cycle; please use the latter.
  warnings.warn(self.msg_depr % (key, alt_key))
Saving results to: out/10_classic1.pdf
Using matplotlib version 1.5.3 with backend agg
Charting with sliding window size: 200, xlimits: None
Loading: warmup_results_0_7_linux1_i7_4790k.json.bz2
No VM instrumentation data is available.
Adding run sequence to  fasta:Graal:default-java bencher3
Loading: warmup_results_0_7_openbsd1_i7_4790.json.bz2
No VM instrumentation data is available.
  File "../bin/plot_krun_results", line 1260, in <module>
    cycles_ylimits=cycles_ylimits)
  File "../bin/plot_krun_results", line 231, in main
    instr_pages.append(data_dcts['instr_data'][key][machine])
KeyError: 'fasta:Graal:default-java'
```

If you remove the file `warmup_results_0_7_openbsd1_i7_4790.json.bz2` from the invocation, then no crash. I think it's because graal doesn't run on OpenBSD and thus the key is nonexistent for that file.

vext01:
Fixes #303.

Looking at the code, I am not really sure why we aren't being bitten by this more frequently with -b.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -1066,6 +1066,7 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
                     data_dictionary['all_outliers'][key] = dict()
                     data_dictionary['common_outliers'][key] = dict()
                     data_dictionary['unique_outliers'][key] = dict()
+                    data_dictionary['classifications'][key] = dict()
 
                 if machine not in data_dictionary['data'][key]:
                     data_dictionary['data'][key][machine] = list()


snim2:
I think we have usually used `-b` without the changepoints, so that's probably why we missed it.

Are you ready to have this merged?

vext01:
The invocation causing the crash doesn't use changepoints. Which is the source of my confusion.

snim2:
Ah, but we append `None`s to the list. Hmm. Still, the original report was about instrumentation data, I think

vext01:
Oh yeah. Hah. So I guess this is another fix. Regardless, with this change, the invocation in #303 no longer crashes.

vext01:
Shall we merge?

snim2 (approved):

Issue #302, opened 2017-01-17
Tweak needed in rect collision algorithm


snim2:
The current rect collision algorithm (used for inset placement) looks to see whether any data which overlaps the rect on the x-axis has its y-value between the top and bottom of the rect.

This fails to catch cases where a measurement passes straight through the rect, for example:

![rect_collision_bug](https://cloud.githubusercontent.com/assets/97674/22018346/4ca86d64-dca6-11e6-8f81-1c86ceb1bf6d.png)

snim2:
Also there are cases where the "tight" bounding box from the figure renderer clips the data. This is obvious if the bounding box is painted in. The rectangle collision algorithm needs to test this "tight" bounding box, not the smaller box that the inset is later resized to.

PR #320, opened 2017-01-26
Aesthetic changes

snim2:
This PR makes outlier markers a little bit smaller and adds a white background to insets, so their ticklabels can be seen more clearly.

[Test case](https://dl.dropbox.com/s/m786npkwqreb3j6/aesthetic_changes_0.pdf?dl=0)


ltratt:
I'm more-or-less fine with this, but the big white box means we're now clipping again: see e.g. binarytrees/pypy/#8. If we can fix that, then I'm fine with merging this.

snim2:
I'd prefer to leave that to Issue #302 if you don't mind

ltratt:
It's not the same bug (although obviously very similar!) -- the clipping is hitting data that the rectangle wholly overwrites. You can add it as a secondary part of 302 if you want though.

snim2:
Yup, although the bugs are different, the fix will be that the rect collision code needs another re-write.

ltratt:
OK.

ltratt (approved):

PR #324, opened 2017-02-13
First sensible bash at new classifier

snim2:
This will probably need a few iterations (and some tidying up when the code has settled).

**Without this change:**

  * Table: [without_change_no_merges.pdf](https://github.com/softdevteam/warmup_experiment/files/770958/without_change_no_merges.pdf)
  * Plot: [plots_no_change.pdf](https://github.com/softdevteam/warmup_experiment/files/770959/plots_no_change.pdf)

**With this change:**

  * Table: [with_change_no_merges.pdf](https://github.com/softdevteam/warmup_experiment/files/770961/with_change_no_merges.pdf)
  * Plot: [plots_with_change.pdf](https://github.com/softdevteam/warmup_experiment/files/770962/plots_with_change.pdf)

### DIFF
--- a/bin/mark_changepoints_in_json
+++ b/bin/mark_changepoints_in_json
@@ -65,13 +65,15 @@ class Segments(object):
     """A list of Segments for a whole run sequence.
     """
 
-    def __init__(self, delta, length, steady_state, cpts, means, variances, data, outliers):
+    def __init__(self, delta, half_bound, steady_state, length, cpts, means,
+                 variances, data, outliers):
         self.delta = delta
+        self.half_bound = half_bound
+        self.steady_state = steady_state
         self.length = length  # Length of original data with outliers.
         assert self.length == len(data)
         self.data = data
         self.outliers = outliers
-        self.steady_state = steady_state
         self.segments = list()
         assert len(means) == len(variances) == len(cpts)
         if len(means) == 1:  # No changepoints.
@@ -174,27 +176,29 @@ class Segments(object):
                 self._merge_segment(0)
 
     def get_classification(self):
-        """Return the classification for this run sequence.
-        """
-        classification = ''
-        # Flat -> no changepoints.
-        if len(self.segments) == 1:
-            classification = 'flat'
-        # No steady state -> there are changes after we expect steady state.
-        elif self.segments[-1].start > (self.length - self.steady_state):
-            classification = 'no steady state'
-        # Warmup -> first segment mean is highest and last is the lowest.
-        elif (self.segments[-1].start <= (self.length - self.steady_state) and
-              self.segments[-1].mean == min(self.segments, key=lambda s: s.mean).mean):
+        """Return a classification for this run sequence."""
+
+        upper_bound = 1.0 + self.half_bound
+        lower_bound = 1.0 - self.half_bound
+        classification = 'flat'
+        last_segment = self.segments[-1]  # Hopefully the steady state segment.
+        for index in xrange(len(self.segments) - 2, -1, -1):
+            current_segment = self.segments[index]
+            if (current_segment.mean > last_segment.mean * lower_bound and
+                current_segment.mean < last_segment.mean * upper_bound):
+                continue
+            elif current_segment.end > (self.length - self.steady_state):
+                classification = 'no steady state'
+                break
+            elif current_segment.mean < last_segment.mean * lower_bound:
+                classification = 'slowdown'
+                break
+            assert current_segment.mean > last_segment.mean * upper_bound
             classification = 'warmup'
-        # Slowdown (b) -> no warmup and no changepoints after we expect a
-        #                 steady state.
-        else:
-            classification = 'slowdown'
         return classification
 
 
-def main(in_files, steady_state, delta):
+def main(in_files, delta, half_bound, steady_state):
     cpt = rpy2.interactive.packages.importr('changepoint')
     r_version = '.'.join(R_VERSION_BUILD[:2])
     print 'Using R version %s and changepoint library %s' % (r_version, cpt.__version__)
@@ -223,7 +227,8 @@ def main(in_files, steady_state, delta):
                     outliers = krun_data[filename]['all_outliers'][bench][index]
                 else:
                     outliers = list()
-                segments = get_segments(cpt, delta, p_exec, steady_state, outliers)
+                segments = get_segments(cpt, delta, half_bound, steady_state,
+                                        p_exec, outliers)
                 if delta > 0.0:
                     segments.merge_overlapping_segments()
                 changepoints[bench].append(segments.changepoints)
@@ -242,7 +247,7 @@ def main(in_files, steady_state, delta):
         write_krun_results_file(krun_data[filename], new_filename)
 
 
-def get_segments(cpt, delta, data, steady_state, outliers):
+def get_segments(cpt, delta, half_bound, steady_state, data, outliers):
     p_exec = data[:]  # data will be passed to Segments unchanged.
     length = len(p_exec)  # Will change when we remove outliers.
     indices = sorted(outliers, reverse=True)
@@ -265,7 +270,7 @@ def get_segments(cpt, delta, data, steady_state, outliers):
         means.append(float(mean))
     for var_ in changepoints.slots['param.est'][changepoints.slots['param.est'].names.index('variance')]:
         variances.append(float(var_))
-    return Segments(delta, length, steady_state, c_points, means, variances, data, outliers)
+    return Segments(delta, half_bound, steady_state, length, c_points, means, variances, data, outliers)
 
 
 def create_output_filename(in_file_name):
@@ -299,7 +304,7 @@ the output of this script will be a new file named:
 
 Example usage:
     $ python %s results1.json.bz2
-    $ python %s  --steady 500 --delta 0.001 results1.json.bz2 results2.json.bz2\n""" % (script, script))
+    $ python %s  --steady 500 --delta 0.001 --bound 0.001 results1.json.bz2 results2.json.bz2\n""" % (script, script))
     parser = argparse.ArgumentParser(description)
     parser.add_argument('json_files', nargs='+', action='append', default=[],
                         type=str, help='One or more Krun result files.')
@@ -313,6 +318,10 @@ Example usage:
                               'merged. If D <= 0 then no attempt will be made to '
                               'merge adjacent segments (this is likely to be '
                               'useful only in testing).'))
+    parser.add_argument('--bound', '-b', action='store', dest='half_bound',
+                        default=0.001, type=float, metavar='B',
+                        help=('Adjacent segments must differ by more than B%% in '
+                              'order to be considered a warmup or slowdown.'))
     return parser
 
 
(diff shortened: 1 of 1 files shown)


ltratt:
In `plots_with_changes` avrora/graal#4 is classified a slowdown, but visually it's a pretty clear warmup?

snim2:
Hmm. The classifier says "check whether the intermediate segments are >0.01% slower than the last segment", and in that case the middle of the three segments is slower. So I guess we need something more sophisticated.

ltratt:
If they're the same (within the margin error), then this should just be a warmup. Looking at the code, I have a feeling it's currently more complex than it needs to be -- but I might be wrong! Let me make a comment or two there...

bin/mark_changepoints_in_json
```diff
@@ -173,24 +173,42 @@ class Segments(object):
...
     def get_classification(self):
         """Return the classification for this run sequence.
         """
+        PC_BOUND = 0.01  # TODO: Factor out as a CLI option before merging with master.
         classification = ''
         # Flat -> no changepoints.
         if len(self.segments) == 1:
             classification = 'flat'
+        # Warmup (a) -> only two segments and second is faster.
+        elif len(self.segments) == 2 and self.segments[1].mean < self.segments[0].mean:
```
ltratt: I doubt we want to special case this (or the clause immediately below it).

bin/mark_changepoints_in_json
```diff
@@ -173,24 +173,42 @@ class Segments(object):
             if math.fabs(self.segments[0].mean - self.segments[1].mean) < self.delta:
                 self._merge_segment(0)
 
+    def _is_segment_a_slowdown(self, index, pc_bound=0.1):
+        """Test whether mean of segment index is slower than the last segment
+        mean by >= pc_bound percent.
+        """
+        if self.segments[index].mean <= self.segments[-1].mean:
```
ltratt: This doesn't look right -- it means that if `segments[index].mean` is 0.100001 and `segments[-1].mean` is 0.100002, then this will return `False` immediately without doing the `pc_bound`% check?

bin/mark_changepoints_in_json
snim2: We probably don't *need* these two clauses, as such, but the intention was to deal with a couple of common cases (obviously the warmup case is much more common than the 2-segment slowdown) and avoid calculating the list on line 205 when we don't need it.

ltratt:
My vague guess as to an algorithm is something like:
```python
class_ = "flat"
for i in range(len(segments) - 2, 0, -1):
    if segments[i].mean * 1.01 < segments[-1].mean:
        class_ = "slowdown"
        break
    elif segments[i].mean * 1.01 > segments[-1].mean:
        class_ = "warmup"
```
This doesn't cope with no steady state; it's not 100% clear to me what we should do if the last segment is less than 500 iterations, but its neighbours are within 1% of its mean.

bin/mark_changepoints_in_json
ltratt: I think this is where I use the "premature optimisation" quote ;) Let's keep the code simple, and optimise only if we really need to.

ltratt:
[One thing neither the algorithm I vaguely suggested, nor this PR, do is consider the variance -- should we? Dunno.]

bin/mark_changepoints_in_json
```diff
@@ -173,24 +173,35 @@ class Segments(object):
...
+                       self.segments[-1].mean * 100.0)
+        return pc >= pc_bound
+
     def get_classification(self):
         """Return the classification for this run sequence.
         """
+        PC_BOUND = 0.01  # TODO: Factor out as a CLI option before merging with master.
         classification = ''
         # Flat -> no changepoints.
         if len(self.segments) == 1:
```
ltratt: I don't think this is right any more -- you can still be flat even if there are multiple change points, provided they're all within X% of each other.

snim2: Ah, I see what you mean. This was why it was a bit simpler, code-wise to merge. I need to think a bit about how this should look now.

ltratt: Maybe we can start with a variant of the algorithm I outlined? The thing that it doesn't do -- and I don't even totally know *what* it should do -- is deal with no steady state. But that's something we can probably worry about as a second step.

snim2: Yup, coming up.

ltratt: Cool, thanks!

snim2:
Hmm, I think you are right

snim2:
Tests for second go:

  * Table: [with_change_no_merges.pdf](https://github.com/softdevteam/warmup_experiment/files/771333/with_change_no_merges.pdf)
  * Plots: [plots_with_change.pdf](https://github.com/softdevteam/warmup_experiment/files/771340/plots_with_change.pdf)

snim2:
OK, I was surprised by how consistent the table was for this run, but the charts look sensible.  Tradebeans pexec 1 seem a bit surprising for a warmup, but we could tweak the value of `PC` if we agree it should be a no steady state (as it was before).

ltratt:
OK, so this looks like it's worth running against all the 0.7 data, and start eyeballing some of the previously unsatisfactory cases. One obvious question is whether 1% is the right fudge factor or not.

bin/mark_changepoints_in_json
```diff
@@ -174,23 +174,25 @@ class Segments(object):
...
+            return classification
+        # > 1 segments: iterate from the penultimate to the first segment.
+        for index in xrange(len(self.segments) - 2, -1, -1):
+            if self.segments[index].mean * PC < self.segments[-1].mean:
+                if self.segments[index].end > (self.length - self.steady_state):
+                    classification = 'no steady state'
+                else:
+                    classification = 'slowdown'
+                break
+            elif self.segments[index].mean * PC >= self.segments[-1].mean:
```
vext01: This line can be just `else:` I think.

bin/mark_changepoints_in_json
```diff
@@ -174,23 +174,25 @@ class Segments(object):
...
-            classification = 'warmup'
-        # Slowdown (b) -> no warmup and no changepoints after we expect a
-        #                 steady state.
-        else:
-            classification = 'slowdown'
+            return classification
+        # > 1 segments: iterate from the penultimate to the first segment.
+        for index in xrange(len(self.segments) - 2, -1, -1):
+            if self.segments[index].mean * PC < self.segments[-1].mean:
+                if self.segments[index].end > (self.length - self.steady_state):
```
vext01: If we lift this line out of the outer conditional, we wouldn't have to duplicate it in the else branch, right?

bin/mark_changepoints_in_json
```diff
@@ -174,23 +174,25 @@ class Segments(object):
...
 
     def get_classification(self):
-        """Return the classification for this run sequence.
-        """
-        classification = ''
-        # Flat -> no changepoints.
+        """Return the classification for this run sequence."""
+        PC = 1.001  # TODO: Factor out as a CLI option before merging with master.
+        classification = 'flat'
         if len(self.segments) == 1:
```
snim2: Note: this `if` can probably be removed.

bin/mark_changepoints_in_json
vext01: in other words, if the last segment is too short, we have "no steady state" regardless of the mean, so we can bail out earlier.

ltratt:
I guess we need to address `half_bound = 0.001  # TODO: Make a CLI option before merging with msater.` before we think about merging.

snim2:
This is the last thing on my list, so if there is anything else, please shout!

snim2:
@ltratt last commit is in. Do you want to review now or shall I squash first ?

ltratt:
Looks good to me. If @vext01 is OK with it, then squash and I'll merge.

bin/mark_changepoints_in_json
```diff
@@ -65,13 +65,15 @@ class Segments(object):
     """A list of Segments for a whole run sequence.
     """
 
-    def __init__(self, delta, length, steady_state, cpts, means, variances, data, outliers):
+    def __init__(self, delta, half_bound, steady_state, length, cpts, means,
+                 variances, data, outliers):
         self.delta = delta
+        self.half_bound = half_bound
```
vext01: `half_bound` is (e.g.) 0.01 for 1%, right? Why is it called `half_bound`. This would imply 50%.

bin/mark_changepoints_in_json
```diff
@@ -174,27 +176,28 @@ class Segments(object):
...
-        # Flat -> no changepoints.
-        if len(self.segments) == 1:
-            classification = 'flat'
-        # No steady state -> there are changes after we expect steady state.
-        elif self.segments[-1].start > (self.length - self.steady_state):
-            classification = 'no steady state'
-        # Warmup -> first segment mean is highest and last is the lowest.
-        elif (self.segments[-1].start <= (self.length - self.steady_state) and
-              self.segments[-1].mean == min(self.segments, key=lambda s: s.mean).mean):
+        """Return a classification for this run sequence."""
```
vext01, line 87: Blank line after docstring?

vext01 (dismissed):
Looks good bar two small comments.

bin/mark_changepoints_in_json
ltratt: I don't think that's necessary is it?

bin/mark_changepoints_in_json
ltratt: Good point, I tend to agree.

snim2: Later we subtract it from `1.0` to make the lower bound, and add it to `1.0` to make the upper bound, so it is a half-bound.

ltratt: Sort of, but it is a little confusing. I don't know what a better name would be though :/

vext01: How about `last_seg_tolerance` or something like that?

bin/mark_changepoints_in_json
vext01: It's just convention in Python.

ltratt: Sorry, I misunderstood you. I stupidly thought you meant a blank line within the docstring... which was a silly thing for me to think. Mea culpa!

bin/mark_changepoints_in_json
ltratt: Since we don't have an idea for a better name, let's keep it as-is.

ltratt:
OK, I think this is ready for squashing.

snim2:
Squashed, but before we merge, do we want to remove the segment merging code altogether?

vext01:
Up to you, but make sure that part is a separate commit.

ltratt:
Good point. I think it would be good to remove that code, assuming that it's now dead, and to do so in this PR (in a separate commit, as Edd says).

snim2:
Coming up

snim2:
Done

ltratt (approved):

ltratt:
So I think I can merge this now?

snim2:
If @vext01 is happy, yet

vext01:
ok

ltratt:
I love PRs that improve functionality yet delete more code than they add!

vext01:
amen!

PR #326, opened 2017-02-16
Copy results off whilst DaCapo and Octane are running.

ltratt:
In order for this to occur, one has to set the `SSH_DO_COPY` variable to something e.g.:

```
  SSH_DO_COPY=1 make bench-octane
```

this way "other" people can just do "make bench-octane" and they won't get all the stuff with our ssh copying confusing them.

### DIFF
--- a/extbench/rundacapo.py
+++ b/extbench/rundacapo.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python
 
-import csv, os, sys, time
+import csv, os, socket, sys, time
 from decimal import Decimal
 from krun.platform import detect_platform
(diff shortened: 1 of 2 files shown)


vext01:
Looks good, my only question, what do we want to do if a copy fails? Currently we ignore the return value of `os.system`, so the error would be ignored. Probably OK as is?

ltratt:
I don't see there's anything sensible we can do without making things much, much more complex (emailing etc.). The copy is optional, so if one fails here or there, it's not critical IMHO.

vext01 (approved):

PR #329, opened 2017-02-16
Ensure the external suites use the same environment as Krun.

vext01:
This small change ensures that the same checks that krun makes (cpu governor, scaler, alsr settings, ...) are used for Octane and Dacapo. This is in response to the fact that it is hard to have a consistent setup for octane/dacapo (since they are not Krunified) and it is easy to forget to set the max clock speed.

Tested on OpenBSD by checking apm status at various points in the runner script.

Needs test on Linux, but review can go ahead.

### DIFF
--- a/extbench/rundacapo.py
+++ b/extbench/rundacapo.py
@@ -3,6 +3,7 @@
 import csv, os, socket, sys, time
 from decimal import Decimal
 from krun.platform import detect_platform
+from krun.config import Config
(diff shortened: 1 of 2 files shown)


vext01:
Works OK on linux too.

ltratt:
Good stuff!

ltratt (approved):

PR #331, opened 2017-02-21
Front end script and readme

snim2:
Adds a new front-end script, refactoring old code along the way. New script:

```
usage: warmup [-h] [--debug DEBUG_LEVEL] --language LANGUAGE --vm VM --uname
              UNAME [--output-plots PDF_FILENAME]
              [--output-html HTML_FILENAME] [--output-latex LATEX_FILENAME]
              [--output-json JSON_FILENAME]
              csv_files [csv_files ...]

Analyse CSV results file and produce JSON, HTML (table) or PDF plots. The CSV input file(s) should be in this format:

	process num, bench_name, 0, 1, 2, ...
	0, spectral norm, 0.2, 0.1, 0.4, ...
	1, spectral norm, 0.3, 0.15, 0.2, ...

Example usage:

	$ python warmup results.csv --output-json summary.json -l javascript -v V8 -u "Linux snim2-r700 4.4.0-38-generic #57-Ubuntu SMP Tue Sep 6 15:42:33 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux"

positional arguments:
  csv_files             One or more CSV result files.

optional arguments:
  -h, --help            show this help message and exit
  --debug DEBUG_LEVEL, -d DEBUG_LEVEL
                        Debug level used by logger. Must be one of: DEBUG, INFO, WARN, DEBUG, CRITICAL, ERROR
  --language LANGUAGE, -l LANGUAGE
                        Language under test (default: javascript).
  --vm VM, -v VM        Virtual machine under test (default: V8).
  --uname UNAME, -u UNAME
                        Full output of `uname -a` from benchmarking machine.

Output formats:
  --output-plots PDF_FILENAME
                        Output a PDF file containing plots.
  --output-html HTML_FILENAME
                        Output an HTML file containing tables.
  --output-latex LATEX_FILENAME
                        Output a LaTeX file containing tables.
  --output-json JSON_FILENAME
                        Output a JSON file containing summary results.
```


ltratt (approved):

ltratt:
OK, so I know this is my fault as I originally suggested calling the command `warmup`, but in retrospect I wonder if `warmup_stats` would be a better name?

snim2:
OK, let's leave renaming the file to last though, otherwise it will be hard to rebase any small changes from the comments...

ltratt:
git will track the rename more-or-less appropriately, so it's probably not an issue to do it now, but I don't mind leaving it until later.

ltratt:
I wonder if we can do something about the `-u` feature. In retrospect, we should have hacked this to just be something like `bencher3` or `bencher7`. I fear that people are often going to get data from a machine and not know the uname...

snim2:
Presumably you mean "without changing the upstream code in Krun", in which case all the parsing code in this repo needs to cope with "proper Krun format" and "Krun format with an arbitrary string for a machine name".

README.md
```diff
@@ -34,6 +33,43 @@ Special notes:
...
+
+The Python script `bin/warmup` should be used as a front-end to all other scripts.
+The script can be used to generate JSON containing summary statistics for
+the input data, PDF plots or LaTeX tables.
+
+The script also needs the names of the language and VM under test, and the
+output of `uname -a` on the machine the benchmarks were run on. Example usage:
+
+```
+./bin/warmup  --output-plots plots.pdf --output-json summary.json -l javascript -v V8 -u "Linux bencher2 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u2 (2016-10-19) x86_64 GNU/Linux" results.csv
```
ltratt: We can simplify this a bit by saying `` -u `uname -a` ``? It might make the command look a little less intimidating!

ltratt:
I don't know. Maybe it's just not practical.

snim2:
Well, it's possible, but it's a reasonably large change. Maybe one for another PR?

ltratt:
OK.

ltratt:
Let's leave it for when we have some rainy afternoons without anything more important to do (ha!).

bin/warmup
```diff
@@ -0,0 +1,275 @@
...
+
+
+def create_arg_parser():
+    script = os.path.basename(__file__)
+    description = ('Analyse CSV results file and produce JSON, HTML (table) or '
+                   'PDF plots. The CSV input file(s) should be in this format:\n\n'
+                   '\tprocess num, bench_name, 0, 1, 2, ...\n'
+                   '\t0, spectral norm, 0.2, 0.1, 0.4, ...\n'
+                   '\t1, spectral norm, 0.3, 0.15, 0.2, ...\n'
+                   '\n\nExample usage:\n\n\t$ python %s results.csv --output-json summary.json '
```
ltratt: It would probably be more conventional to pull all the args first, then `results.csv` at the end of the command? We can also possibly reuse the `uname` trick mentioned elsewhere.

bin/warmup
```diff
@@ -0,0 +1,275 @@
...
+        if pdflatex_path is None:
+            fatal('You do not appear to have pdflatex installed, and so cannot '
+                  'compile LaTeX files. Please install LaTeX (e.g. tetex-live) and try again.')
+    if need_plots:
+        try:
+            import matplotlib
+        except ImportError:
+            fatal('Please install the Python matplotlib library to generate plots.')
+        try:
+            matplotlib.use('Agg')
```
ltratt: Does this need to be in the `ImportError` try block? [I don't know, it just looks a little unusual.]

README.md
```diff
@@ -34,6 +33,43 @@ Special notes:
...
+  * bzip2 / buzip2
+  * Python modules: numpy, rpy2
+  * R (e.g. `sudo apt-get install r-base`)
+  * Python modules required for plotting: matplotlib, seaborn
+  * Required for generating LaTeX tables: pdflatex (e.g. `sudo apt-get install tetex-live`)
+
+### CSV format
+
+Scripts here expect to work with benchmark data in CSV format. The first row
+should contain a header with a process execution id, benchmark name and sequence
```
ltratt: should -> must

ltratt: I think we need to be clear that the number of in-process iterations numbers must be *exactly* the number of in-process iterations for each process execution (or bad things happen).

bin/warmup
snim2: Yes, otherwise matplotlib tries to use the Tkinter backend, and Tk no longer comes with Python (or something) and seaborn cannot be imported.

bin/warmup
snim2: It would, but if we do that we cannot read in the existing CSV files from v0.7 etc.

ltratt: I'm not sure I understand. Why would that be a problem?

snim2: We already have CSV files in the format above, so if we change all the parsers to read in a different format, we would also need to change the CSV files we already have, and upload them again to archive.

ltratt: I'm not suggesting changing the format: I'm suggesting being clearer about it.

snim2: Ah, sorry!

README.md
```diff
@@ -34,6 +33,45 @@ Special notes:
  * Since we cannot distribute Java Richards, a patch is required to derive the
    tracing version (`patches/trace_java_richards.diff`)
 
+## Using the scripts here to analyse an existing benchmark
+
+### Requirements
+
+  * Python 2.7 (the code here is not Python 3.x ready)
+  * bzip2 / buzip2
```
ltratt: Is this supposed to be `bunzip2` (with an N)?

README.md
```diff
@@ -34,6 +33,45 @@ Special notes:
  * Since we cannot distribute Java Richards, a patch is required to derive the
    tracing version (`patches/trace_java_richards.diff`)
 
+## Using the scripts here to analyse an existing benchmark
+
+### Requirements
```
ltratt: I think we need to sub-split this into "mandatory requirements" and "optional" requirements. At the moment, I don't think anyone will understand why there's "python modulzes" and "python modules required for plotting" as separate items.

README.md
```diff
@@ -34,6 +33,45 @@ Special notes:
...
+
+  * Python 2.7 (the code here is not Python 3.x ready)
+  * bzip2 / buzip2
+  * Python modules: numpy, rpy2
+  * R (e.g. `sudo apt-get install r-base`)
```
ltratt: There's no point giving package names for an arbitrary Linux distribution here: they all call them different things. We can have a separate section or file called something like `README.debian` for that sort of stuff.

README.md
```diff
@@ -34,6 +33,45 @@ Special notes:
...
+  * Python modules required for plotting: matplotlib, seaborn
+  * Required for generating LaTeX tables: pdflatex (e.g. `sudo apt-get install tetex-live`)
+
+### CSV format
+
+Scripts here expect to work with benchmark data in CSV format. The first row
```
ltratt: Let's change the first sentence to something like "The scripts here take CSV files as input. The format is as follows. The first row must ...".

README.md
```diff
@@ -34,6 +33,45 @@ Special notes:
...
+should contain a header with a process execution id, benchmark name and sequence
+of iteration numbers. Subsequent rows should contain an id, benchmark name and
+times (in seconds) for each benchmark iteration. Each process execution must
+execute the same number of iterations as described in the header. For example:
+
```
ltratt: Too many blank lines?

README.md
```diff
@@ -34,6 +33,47 @@ Special notes:
...
+### Mandatory requirements
+
+  * Python 2.7 (the code here is not Python 3.x ready)
+  * bzip2 / bunzip2
+  * Python modules: numpy, rpy2
+  * R
+
+### Optional requirements
+
+  * Python modules required for plotting: matplotlib, seaborn
```
vext01: Did we not remove the need for seaborn at some point?

README.md
```diff
@@ -34,6 +33,47 @@ Special notes:
...
+  * Required for generating LaTeX tables: pdflatex
+
+### CSV format
+
+The scripts here take CSV files as input. The format is as follows. The first row
+must contain a header with a process execution id, benchmark name and sequence
```
vext01, line 45: `The first row is the CSV header line with the headers: "process num", "bench_name", and then one numeric index header for each in-process iteration.`

Why `process num` but `bench_name`? One is underscored.

Should we use "proc_exec_num" to be more consistent with terminology in the paper?

README.md
```diff
@@ -34,6 +33,47 @@ Special notes:
...
+of iteration numbers. Subsequent rows should contain an id, benchmark name and
```
vext01: `Subsequent rows are data rows, one per process execution. The in-process iteration index columns should contain the time in seconds for the corresponding in-process iteration` is probably enough?

README.md
```diff
@@ -34,6 +33,57 @@ Special notes:
...
+
+### Setting up R
+
+To run the scripts here, it is necessary to install some R packages. To do this,
+open R on the command line, and run the following commands:
+
+```R
+> install.packages("devtools")
+> devtools::install_github("rkillick/changepoint")
+```
```
vext01: Since this will write into the user's $HOME, can we tell them where we install? User may want to uninstall this stuff later.

bin/csv_to_krun_json
```diff
@@ -63,35 +32,21 @@ def create_cli_parser():
...
-                        dest='language',
-                        action='store',
-                        default='java-default',
-                        type=str,
+    parser = argparse.ArgumentParser(description,
+                                     formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument('csv_files', nargs='+', action='append', default=[],
+                        type=str, help='One or more CSV result files.')
+    parser.add_argument('--language', '-l', dest='language', action='store',
+                        default='java-default', type=str,
```
vext01: Default variants are usually `default-*`, so do we mean `default-java` here?

vext01: Do we really want defaults for language and vm? This could lead to confusion.

bin/mark_outliers_in_json
```diff
@@ -70,8 +70,7 @@ def create_output_filename(in_file_name, window_size):
 
 
 def create_cli_parser():
-    """Create a parser to deal with command line switches.
-    """
+    """Create a parser to deal with command line switches."""
```
vext01, line 73: \n after docstring.

bin/table_classification_summaries_main
```diff
@@ -45,119 +38,14 @@ TABLE_HEADINGS2 = '&&\\multicolumn{1}{c}{Class.} &\\multicolumn{1}{c}{iter (\#)}
...
 
-def write_results_as_latex(machine, all_benchs, summary, steady_state, tex_file,
+def write_latex_table(machine, all_benchs, summary, steady_state, tex_file,
                            with_preamble=False):
-    """Write a tex table to disk"""
+    """Write a tex table to disk.
+    This script has its own version of the write_latex_table function, in which
+    each VM is a separate row, and benchmarks are grouped together (not the other
+    way around).
+    """
```
vext01, line 48: \n

bin/warmup
```diff
@@ -0,0 +1,274 @@
...
+    def convert_to_krun_json(self):
+        header, self.krun_filename = csv_to_krun_json([self.csv_filename],
+                                             self.language, self.vm, self.uname)
+        info('Writing out: %s' % self.krun_filename)
+        try:
+            self.iterations = int(header[-1]) + 1  # Iteration numbers start at 0.
+        except ValueError:
+            fatal('CSV file has malformed header. Run this script with --help for more details.')
+
+    def _get_output_filename(self, output):
```
vext01: This is parsing the output of a script to find the output filename? This isn't ideal, but I could live with it.

warmup/summary_statistics.py
```diff
@@ -0,0 +1,300 @@
...
+                    fp.write('\\\\ \n')
+                bench_idx += 1
+            if split_row_idx < vms_per_split - 1:
+                fp.write('\midrule\n')
+            split_row_idx += 1
+        fp.write(end_table())
+        if with_preamble:
+            fp.write('\\end{table*}\n')
+            fp.write(end_document())
+    return
```
vext01: return needed?

vext01 (dismissed):
Some (somewhat superficial) comments.

vext01:
I'll test this on a VM now.

bin/csv_to_krun_json
ltratt: Good point -- we probably don't want such defaults.

README.md
ltratt: Is it possible to have it install into the current directory (i.e. the warmup experiment repo) rather than dumping stuff elsewhere?

vext01: I don't know if R has something like virtualenv.

README.md
snim2: We haven't merged the relevant PR yet, but that PR will need to remove seaborn from the README.

README.md
snim2: I don't know much about R packages, I'll have a search around.

bin/csv_to_krun_json
snim2: I added them in to be clear about where we expect lowercase and title case, but we could write that in words.

ltratt: I don't think people will necessarily pick up the lowercase/titlecase thing, so it's probably best to spell it out and remove the defaults.

README.md
```diff
@@ -34,6 +33,58 @@ Special notes:
...
+
+  * Python 2.7 (the code here is not Python 3.x ready)
+  * bzip2 / bunzip2
+  * Python modules: numpy, rpy2
+  * R
+
+### Optional requirements
+
+  * Python modules required for plotting: matplotlib, seaborn
+  * Required for generating LaTeX tables: pdflatex
```
vext01: Should we say "a latex distribution containing pdflatex and the x, y, z packages. TeX Live works fine."

snim2: Good idea.

README.md
```diff
@@ -34,6 +33,70 @@ Special notes:
...
+
+To run the scripts here, it is necessary to install some R packages. By default,
+R will install these packages in `$HOME/R`. If you do not want R to use your
+home directory, then set the environment variable `$R_LIBS_USER`, e.g. (in BASH):
+
+```bash
+$ git clone https://github.com/softdevteam/warmup_experiment.git
+$ cd warmup_experiment
+$ mkdir R
+$ export R_LIBS_USER=`pwd/R`
```
vext01: \`pwd\`/R

bin/warmup
snim2: Two scripts do this (`mark_outliers...` and `mark_changepoints`). Strictly speaking, we could just compute the filename in `warmup`, but I was worried that if we later change the scripts the `warmup` will break. 

I'm open to suggestions here, the file naming isn't great at the moment. We currently write the window size into the filename but not the other options, and we do now store all the important command line options in Krun JSON format anyhow, so we are not very consistent about naming.

README.md
```diff
@@ -34,6 +33,73 @@ Special notes:
...
+$ mkdir R
+$ export R_LIBS_USER=`pwd`/R
+$ echo "export R_LIBS_USER=`pwd`/R" >> ~/.bashrc
+```
+
+To install the necessary packages, open R on the command line, and run the
+following commands:
+
+```R
+> install.packages("devtools")
```
vext01: `At this point R may ask you to choose a mirror. Choose one, wait for installation to complete then run:`

vext01: Additional dependencies needed at this point (deb packages in brackets):
 * openssl and headers (libssl-dev)
 * pkg-config (pkg-config)
 * curl and headers (libcurl4-openssl-dev)
 * Although something else pulled in a C compiler and make, we should probably mention it anyway. (build-essential, make)

README.md
```diff
@@ -34,6 +33,73 @@ Special notes:
...
+> devtools::install_github("rkillick/changepoint")
```
vext01:
```
> devtools::install_github("rkillick/changepoint")
Downloading GitHub repo rkillick/changepoint@master
from URL https://api.github.com/repos/rkillick/changepoint/zipball/master
Installing changepoint
Error in if (capabilities("libcurl")) { : argument is of length zero
```

vext01: Which is odd, as R installed curl bindings earlier.

vext01:
```
>  install.packages("curl")
...
gcc -std=gnu99 -shared -Wl,-z,relro -o curl.so callbacks.o curl.o download.o e
scape.o fetch.o form.o getdate.o handle.o ieproxy.o init.o interrupt.o multi.o
 nslookup.o reflist.o utils.o version.o -lcurl -L/usr/lib/R/lib -lR
installing to /home/edd/warmup_experiment/R/curl/libs
** R
** data
*** moving datasets to lazyload DB
** inst
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded
* DONE (curl)

The downloaded source packages are in
        ‘/tmp/RtmpmTHDAi/downloaded_packages’
> devtools::install_github("rkillick/changepoint")
Downloading GitHub repo rkillick/changepoint@master
from URL https://api.github.com/repos/rkillick/changepoint/zipball/master
Installing changepoint
Error in if (capabilities("libcurl")) { : argument is of length zero
```

ltratt:
Let's not special case Debian in the readme. There are lots of Linuxes -- and lots of Unixes! -- that aren't Debian. Either Debian should be a sub-heading in the readme, or we can have a README.debian file.

ltratt:
One thing we could usefully do is have a Debian subsection in the README that has an `apt-get` command for all the relevant packages. So it would say something like:

```
apt-get install bzip2 python-numpy python-rpy2
```

and so on. That way a Debian user can just cut and paste that into their terminal. [I have no preference as to whether we add in all the optional requirements or not.] We can also, if someone works out what they are, do the same job for non-Debian OSs at a later point.

README.md
vext01: The workaround for this R bug is: `options(download.file.method = "wget")`. Let's mention it as there's about a 99% chance users will be using debian (which has the buggy R version).

vext01:
Agree with a debian section. It's most likely what our users will be using.

README.md
snim2: Just to be clear, that command needs to be entered into the R repl?

vext01: Yeah

vext01: Digging deeper: "capabilities('libcurl') was added in R 3.2, so for R 3.1, it being of length=0 is expected behavior."

vext01: (ugh)

snim2:
Asterisks are weird in apt-get, but they do work as above

ltratt:
The `*` is expanded *before* apt-get is called. Your shell *might* do something clever with it, possibly because of a config option you've got on. However, one must not rely on this behaviour, as the command will definitely behave differently on other shells.

vext01:
I would have expected `texlive-*` to expand to nothing unless you had files in `pwd` called `texlive...`

ltratt:
It totally depends on the shell and shell config. However, we should assume that everyone's got a completely basic shell like `ksh` which has no such cleverness.

vext01:
Weirdly sarah is right about bash and apt. But we should use a simpler command that works for other shells. Also, all of texlive might be overkill, i'm hoping texlive-extra will be enough.

ltratt:
For example, on OpenBSD my zsh config does clever package stuff if the command starts `pkg_info`/`pkg_add` etc. But the important thing is that the expansion happens *before* apt-get is called, and we can't rely on other shells / other shell configs doing such an expansion.

vext01:
I agreed with you laurie...

snim2:
Well, I have looked through the package lists for the texlive packages and I think these two packages have the latex things we need.

vext01:
Cool!  Also, I saw code mentioning PyPy will make stuff faster. Add as an optional dependency?

snim2:
Is there any way we can test the `sudo apt-get install` command?

vext01:
Yes, but it wouldn't be fair on travis to make them download and install texlive on each push.

vext01:
Next R issue:

```
$ ../../bin/warmup --output-html graal.html --output-latex graal.tex --output-json graal.json --language java --vm graal --uname "`uname -a`" --output-plots graal.pdf dacapo_graal_results_0_7_linux1_i7_4790k.raw
/usr/lib/python2.7/dist-packages/rpy2/robjects/packages.py:216: UserWarning: Conflict when converting R symbol in the package "tools" to a Python symbol (package.dependencies -> package_dependencies while there is already package_dependencies)
  warn(msg)
  File "/home/edd/warmup_experiment/bin/mark_changepoints_in_json", line 240, in <module>
    main(options.json_files[0], options.half_bound, options.delta, options.steady_state)
  File "/home/edd/warmup_experiment/bin/mark_changepoints_in_json", line 114, in main
    assert r_version >= '3.3.1', 'Please update R from CRAN.'
AssertionError: Please update R from CRAN.
  File "../../bin/warmup", line 274, in <module>
    main(options)
  File "../../bin/warmup", line 231, in main
    benchmark.mark_changepoints()
  File "../../bin/warmup", line 201, in mark_changepoints
    output = subprocess.check_output(' '.join(cli), shell=True)
  File "/usr/lib/python2.7/subprocess.py", line 573, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '/usr/bin/python2.7 /home/edd/warmup_experiment/bin/mark_changepoints_in_json -s 500 dacapo_graal_results_0_7_linux1_i7_4790k_outliers_w200.json.bz2' returned non-zero exit status 1
```

snim2:
Ah, that's saying you need a later version of R, so I guess I must have hit this issue before. I'll remove that line and push, and we can see if this all works with R 3.3.1. If not, we need to add manual instructions for installing a later R :(

vext01:
This time i get:

```
/usr/lib/python2.7/dist-packages/rpy2/robjects/packages.py:216: UserWarning: Conflict when converting R symbol in the package "tools" to a Python symbol (package.dependencies -> package_dependencies while there is already package_dependencies)
  warn(msg)
  File "/home/edd/warmup_experiment/bin/mark_changepoints_in_json", line 240, in <module>
    main(options.json_files[0], options.half_bound, options.delta, options.steady_state)
  File "/home/edd/warmup_experiment/bin/mark_changepoints_in_json", line 138, in main
    segments = get_segments(cpt, half_bound, delta, steady_state, p_exec, outliers)
  File "/home/edd/warmup_experiment/bin/mark_changepoints_in_json", line 166, in get_segments
    c_points = [int(cpoint - 1) for cpoint in changepoints.slots['cpts']]
AttributeError: 'RS4' object has no attribute 'slots'
  File "../../bin/warmup", line 274, in <module>
    main(options)
  File "../../bin/warmup", line 231, in main
    benchmark.mark_changepoints()
  File "../../bin/warmup", line 201, in mark_changepoints
    output = subprocess.check_output(' '.join(cli), shell=True)
  File "/usr/lib/python2.7/subprocess.py", line 573, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '/usr/bin/python2.7 /home/edd/warmup_experiment/bin/mark_changepoints_in_json -s 500 dacapo_graal_results_0_7_linux1_i7_4790k_outliers_w200.json.bz2' returned non-zero exit status 1
```

So I guess we do need a new R, but as I recall you can do that with cran itself?

vext01:
Tried `installr`:

```
package ‘installr’ is not available (for R version 3.1.1) 
```

snim2:
I think on Debian we need to add a new apt source

vext01:
If R isn't hard to build, I think I may prefer this... Let me have a quick go...

vext01:
The build is running. There are also binary distributions on the R page, we might get away with just saying "you need R>=3.2.2. Refer to R docs for getting a new R".

Will let you know how I get on under the new R.

vext01:
```
edd@rtest:~/R-3.3.2$ ~/R-inst/bin/R

R version 3.3.2 (2016-10-31) -- "Sincere Pumpkin Patch"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)
```

Will now nuke ~/R and try from scratch

ltratt:
@vext01 We could have a "build_stats.sh" file which does things like build R, install Rebecca's library, and so on?

snim2:
You need to run the R REPL to install the changepoint package BTW, not sure how we do that in a shell script.

ltratt:
I'd try the ol' classic `echo what you want to type into the REPL | R` -- it may not work, but it'll probably be OK.

vext01:
@ltratt maybe, but a separate PR. Let's solve this part first.

ltratt:
OK.

vext01:
Note also that the R package manager is interactive and asks things like which mirror to use. I also notice that the newer R asks different questions to the last version we tried.

snim2:
I think I have found a way to turn off the interactive choose-a-CRAN-mirror, but let's leave that to the next PR

vext01:
nice.

vext01:
Right, finally our script it doing something.

To build from R from source I needed the Xorg headers (Xorg-dev), and to configure with:
```
./configure --prefix=/home/edd/R-inst --enable-R-shlib
```

Then `make && make install` work fine.

Because R is in a non standard path, we then have to add `~/R-inst` to `PATH` and add `~/R-inst/lib/R/lib` to `LD_LIBRARY_PATH`.

Finally, we can't use rpy2 from apt any more, as it links the old R. I installed one that uses our new R with:
```
pip install --user rpy2
```

snim2:
OK, so you want me to add all this to the README?

vext01:
Well, we could do. Or if we write that script to automate this we don't need to... thoughts?

vext01:
```
  File "../../bin/warmup", line 274, in <module>
    main(options)
  File "../../bin/warmup", line 265, in main
    raise NotImplementedError()
NotImplementedError
```

Did you say HTML is not supported?

snim2:
Yes, I'll do that in another PR

vext01:
ok

vext01:
Seems to work. Just tried one dacapo dataset and got sensible output. Trying octane.

snim2:
A minor miracle!

README.md
```diff
@@ -60,9 +62,11 @@ Special notes:
 The following command will install all dependencies on Debian-based systems:
 
 ```sh
-$ sudo apt-get install build-essential python2.7 pypy bzip2 r-base libssl-dev \
-       pkg-config libcurl4-openssl-dev python-numpy python-rpy2 \
-       python-matplotlib python-seaborn texlive texlive-latex-extra
+$ sudo apt-get install build-essential python2.7 pypy bzip2 libssl-dev \
+       pkg-config libcurl4-openssl-dev python-numpy python-matplotlib\
+       python-seaborn python-pip texlive texlive-latex-extra wget xorg-dev
```
vext01: No need for `texlive`, `texlive-latex-extra` is sufficient.

build_stats.sh
```diff
@@ -0,0 +1,23 @@
...
+tar -xvzf R-3.3.2.tar.gz
+cd R-3.3.2
+
+# Build R.
+./configure --prefix=${INST_DIR} --enable-R-shlib
+make
+make install
+
+cd ..
+
```
vext01: Put R in path and add LD_LIBRARY_PATH.

snim2: How do we put R in the path in a way that isn't BASH specific?

ltratt: The `PATH` variable is, I believe, common to every shell, so "the normal" should be fine.

snim2: My normal would be `echo "export PATH=$PATH:..." >> ~/.bashrc` otherwise the new `PATH` will be lost when the shell is closed, that won't work for everyone, I think?

vext01:
Should we use a directory for R libs and rpy2?

build_stats.sh
ltratt: Ah, I see! Mutating people's `rc` files is a definite no-no. We have two choices.

First, we could add an `echo` to the end saying something like:

```sh
echo "You will need to add $INST_PATH to your $$PATH in order to run bin/warmup_stats"
```

Second -- and perhaps better in this case -- is to simply make `warmup_stats` invoke R from the directory it's been installed into.

vext01: Note you still need to set the lib path to use the latter.

build_stats.sh
```diff
@@ -12,9 +12,15 @@ cd R-3.3.2
 ./configure --prefix=${INST_DIR} --enable-R-shlib
 make
 make install
-
 cd ..
 
+# Correct PATH and LD_LIBRARY_PATH variables.
+export PATH=$PATH:${INST_DIR}
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${INST_DIR}/lib/R/lib
```
vext01: Other way around, or we could pick up a system installed R.

```
export PATH=${INST_DIR}/bin:${PATH}
```

Same for LD_LIBRARY_PATH.

Also, don't forget braces on LD_LIBRARY_PATH for consistency.

vext01: Note my edit to add `/bin`

build_stats.sh
```diff
@@ -12,9 +12,15 @@ cd R-3.3.2
...
+
+echo "Please add ${INST_DIR} to your PATH environment variable and \
```
vext01: again here: `${INST_DIR}/bin`

snim2:
@vext01 I think that answers your question on R packages, I'm not sure about Python packages, suggestions welcome.

vext01:
OK, so i we are using dirs for R and python libs, we would need to use virtualenv for python but share libraries with the system tree (or it won't see matplotlib etc. and we don't want to build our own):

```
$ virtualenv --help
...
  --system-site-packages
                        Give the virtual environment access to the global
                        site-packages.
```

Then the user would need to invoke that python interpreter, not the system one.

vext01:
Alternatively, we dont use virtualenv and say "pip wil install rpy2 in `~/.local`".

`~/.local` is shared with the system tree by default, so the user is ok to invoke the system interpreter in that case.

snim2:
@vext01 virtualenv sounds like overkill to me, maybe let's leave this as-is. Does the script work for you?

vext01:
I leave this decision up to you, as I'm not sure.

snim2:
OK, well unless @ltratt intervenes, I'm happy with `~/.local`

vext01:
fwiw, `pip uninstall --user ryp2` would remove rpy2 from `~/.local`

ltratt:
I *really* dislike messing with users home directories, full stop, so let's avoid `~/.local`. But I don't think we need to. Why doesn't our build script download and build rpy2 and then just add its directory to `warmup_stat`'s `sys.path`?

build_stats.sh
```diff
@@ -0,0 +1,34 @@
...
+./configure --prefix=${INST_DIR} --enable-R-shlib
+make
+make install
+cd ..
+
+# Correct PATH and LD_LIBRARY_PATH variables.
+export PATH=${INST_DIR}:${PATH}
+export LD_LIBRARY_PATH=${INST_DIR}/lib/R/lib:${LD_LIBRARY_PATH}
+
+echo "Please add ${INST_DIR}/bin to your PATH environment variable and \
```
vext01: This works, but you need to move all of the messages to the end, as pip scrolls a load of stuff up the screen.

vext01: Probably add `set -e` to the top of the script too (will try that now).

ltratt:
Can we not get rid of the `echo` hack, and just automatically find the version of R we've installed in the Python script? In an ideal world a user would just do:

```sh
$ ./build_stats.sh
$ warmup_stats -o plots.pdf data.csv
```

without messing with the `PATH` variable or pip or virtualenv or anything like that.

vext01:
rpy2 has dependencies, which we would need to also install (six and singledispatch). We might get away with installing those from OS package manager, and manually pulling in rpy2 and build it, but I'm not sure I like this. Folks are likely to be versed with pip, so we would probably be doing them a favor by doing things in the standard way.

I don't understand the second comment. rpy2 needs to be built against R using a C compiler. You can't pick up R at runtime.

ltratt:
Only hardcore Python people are well versed with Pip; it's a barrier to use. I have gone my entire life without using Pip (apart from once, and you know how that ended).

What I mean is: we don't have to tell people to alter their `PATH`. We can have code which looks something like:

```python
r_path = os.path.join(os.path.basename(__FILE__), "..", "r_inst")
os.system("%s %s" % (os.path_join(r_path, "bin", "r")), ...))
```

and find the R binary ourselves. If we don't want to do that, we can even alter the `PATH` variable inside the Python binary, so that the child shell that's invoked picks up R correctly.

vext01:
I'm going to try installing rpy2's deps and then fetching and compiling rpy2 manually.

> Only hardcore Python people are well versed with Pip

Na, pip is easy. But I know you don't like it.

ltratt:
I didn't say pip isn't easy, but it's not like everyone will have used it: it's something else people might need to look up and understand. In that sense, it's a barrier to use. I'm all for removing barriers :)

snim2:
They don't need to look it up and understand it, they just need to run the build_stats script which runs it for them.

vext01:
I'm going to try, but if we end up having to recurse far down rpy2's dependency trail, I'm going to suggest we just use a virtualenv.

vext01:
(this all assumes rpy2 can even run out of its source directory)

ltratt:
A thought. Does `pip` take a `--prefix` option? If so, we could install things into `work` in the warmup_experiment repo without having to manually do all the installation ourselves.

vext01:
That's pretty much a virtualenv.

vext01:
Let me expand on this.

A virtualenv is a Python environment in a directory with it's own `bin` and `lib` dirs. It also comes with a pip, so if you use that pip, packages are installed into the virtualenv. It's a nice tool and I use it a lot for my own stuff.

ltratt:
I don't think it's equivalent to virtualenv. IIRC that requires you to adjust your shell's path and so on.

What I'm suggesting *might* be possible (it might not!) is to do something like `pip install --prefix $WORK_DIR package1 package2` then the `sys.path` hack from earlier to read from that directory. That way the user a) doesn't have us fiddling with anything in their home dir 2) doesn't have to do any setup when running `warmup_stats`.

vext01:
Pip itself has:

```
  --root <dir>                Install everything relative to this alternate
                              root directory.
```

Never used this so can't vouch for it.

snim2:
@vext01 are you absolutely sure that your system rpy2 will not work with the current code / R version?

ltratt:
Assuming it compiles against an R installation's headers and so on (I guess it does?), the only safe thing to do is to recompile it against our R installation. The system one *might* work, but it may well be just by chance on one person's system.

vext01:
I'm sure.  It links against libR.so, which will be R version specific:

```
$ ldd /usr/lib/python2.7/dist-packages/rpy2/rinterface/_rpy_device.so
...
        libR.so => /usr/lib/libR.so (0x00007f9ab1e53000)
```

vext01:
```
 $ pip2 install --root /tmp/pip rpy2
...building
$ find /tmp/pip
/tmp/pip
/tmp/pip/usr
/tmp/pip/usr/local
/tmp/pip/usr/local/lib
/tmp/pip/usr/local/lib/python2.7
/tmp/pip/usr/local/lib/python2.7/dist-packages
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/RECORD
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/METADATA
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/WHEEL
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/top_level.txt
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/DESCRIPTION.rst
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/zip-safe
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch-3.4.0.3.dist-info/metadata.json
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch_helpers.pyc
/tmp/pip/usr/local/lib/python2.7/dist-packages/singledispatch.pyc
/tmp/pip/usr/local/lib/python2.7/dist-packages/rpy2
/tmp/pip/usr/local/lib/python2.7/dist-packages/rpy2/rinterface
/tmp/pip/usr/local/lib/python2.7/dist-packages/rpy2/rinterface/__init__.py
/tmp/pip/usr/local/lib/python2.7/dist-packages/rpy2/rinterface/__init__.pyc
...
```

So tihs could work, but shame about the yucky `/usr/local/lib/python2.7/dist-packages/` prefix.

Just going to see if that can be avoided.

vext01:
Boom! `pip install -t <dir>` is what we want. Then we set PYTHONPATH.

Happy with that @ltratt?

snim2:
I get this with pip install -t:
```
Exception:
  File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/lib/python2.7/dist-packages/pip/commands/install.py", line 335, in run
    prefix=options.prefix_path,
  File "/usr/lib/python2.7/dist-packages/pip/req/req_set.py", line 742, in install
    **kwargs
  File "/usr/lib/python2.7/dist-packages/pip/req/req_install.py", line 831, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/usr/lib/python2.7/dist-packages/pip/req/req_install.py", line 1032, in move_wheel_files
    isolated=self.isolated,
  File "/usr/lib/python2.7/dist-packages/pip/wheel.py", line 247, in move_wheel_files
    prefix=prefix,
  File "/usr/lib/python2.7/dist-packages/pip/locations.py", line 153, in distutils_scheme
    i.finalize_options()
  File "/usr/share/python-wheels/setuptools-26.1.1-py2.py3-none-any.whl/setuptools/command/install.py", line 38, in finalize_options
    orig.install.finalize_options(self)
  File "/usr/lib/python2.7/distutils/command/install.py", line 289, in finalize_options
    raise DistutilsOptionError("can't combine user with prefix, "
DistutilsOptionError: can't combine user with prefix, exec_prefix/home, or install_(plat)base
```

snim2:
In fact, I can't see a `-t` option in `pip --help`?

ltratt:
Assuming by "PYTHONPATH" you mean the `sys.path` hack (and not the poor user having to manually adjust the `PYTHONPATH` env var), then yes.

vext01:
@snim `pip install --help` will show the `-t` docs.

snim2:
@vext01 ok, I now see the `-t` in help, but that doesn't resolve the error. Have you managed to get it to work?

vext01:
I just did `pip install -t /tmp/pip rpy2` and it worked first time.

snim2:
What version of pip are you using?

vext01:
`1.5.6`

snim2:
```
$ pip -V
pip 8.1.2 from /usr/lib/python2.7/dist-packages (python 2.7)
```

build_stats.sh
```diff
@@ -0,0 +1,40 @@
+#! /bin/sh
+
+set -e
+
+HERE_DIR="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
```
ltratt: Can't this just be:
```
HERE_DIR=`basename $0`?
```

snim2: We used to have that, I think, but it doesn't provide a full path on my machine and one of the later commands fails.

build_stats.sh
```diff
@@ -0,0 +1,40 @@
...
+INST_DIR=${HERE_DIR}/R-inst
```
ltratt: Why don't we dump this stuff into `work` so that it doesn't clutter up the top-level of the repository? So I guess it would be something like `INST_DIR=${HERE_DIR}/work/R-inst` and so on.

build_stats.sh
```diff
@@ -0,0 +1,40 @@
...
+PIP_DIR=${HERE_DIR}/pip
+mkdir -p ${INST_DIR}
+R_LIBS_USER=${HERE_DIR}/R
+mkdir -p ${R_LIBS_USER}
+
+# Download a recent version of R (must be 3.3.1 or later).
+wget https://cran.r-project.org/src/base/R-3/R-3.3.2.tar.gz
+tar -xvzf R-3.3.2.tar.gz
```
ltratt: There's no need for `v`: just `tar xfz R-3.3.2.tar.gz` will do.

build_stats.sh
ltratt: You're right, this is tricky. I knew a good trick for this at one point, but I've forgotten it :/ I know that we can't use `readlink -f` (which is the obvious way) because OS X doesn't have support.

snim2: FWIW I've been testing with `sh` rather than `bash` and this seems to be working

ltratt: Hmm, it doesn't work properly on my system if I'm outside the directory but the script is in my path. Let me look into it.

ltratt: OK, so this is one of those things where to be fully portable you have to dance with both feet up in the air whilst simultaneously having no other part of your body touching the ground.

But a pragmatic approach is simply:
```
HERE_DIR=`dirname $_`
```
For example if I have a file `/tmp/woo.sh` which prints out using the "cd" mechanism and then the "$_" mechanism I get:
```sh
$ cd /home/ltratt
$ PATH=/tmp/ woo.sh
/home/ltratt
/tmp/
```
i.e. because the "cd" mechanism uses `pwd` it picks up where I'm executing the command from, even though that has nothing to do with the script.

snim2: This breaks for me in two ways. Firstly if I run `sh build_stats.sh` then `HERE_DIR` turns out to be `/bin/` (i.e. where the `sh` binary is stored). 

If I run it as `./build_stats.sh` the script fails (as described above) with:

```
configure: error: expected an absolute directory name for --prefix: ./work/R-inst
```

as it needs a full path for `HERE_DIR` and not just `.`.

ltratt: You're right :( So none of our mechanisms work. Damned OS X; `readlink -f` is what we really need...

snim2:
@vext01 with respect to the pip problem, I upgraded to `pip 9.0.1` and no longer have the previous problem. I've modified the scripts to use the local version of rpy2, it seems to work for me, hopefully it will work for you too!

build_stats.sh
ltratt: OK, let's just bite the bullet and do:

```sh
HERE_DIR=$(dirname $(readlink -f "$0"))
```

and if they don't have `readlink` they should install it. My guess is most proper Unix people do (on Linux and OS X it's probably under the `core-utils` package).

snim2:
Too long since last coffee.

snim2:
Squashed!

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
  * Since we cannot distribute Java Richards, a patch is required to derive the
    tracing version (`patches/trace_java_richards.diff`)
 
+## Using the scripts here to analyse an existing benchmark
+
+### Mandatory requirements
+
+  * Python 2.7 - the code here is not Python 3.x ready
+  * bzip2 / bunzip2 and bzip2 headers
+  * curl and headers
```
ltratt: Can we put something like Curl (including its header files)? "and headers" is a little vague, and not obviously relevant on OSs that routinely package headers with programmes.

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
...
+  * gcc and make
+  * gfortran and headers
+  * liblzma library and headers
```
ltratt: Same "headers" comment as earlier. [There are a couple below too.]

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
...
+  * PyPy (will allow some code here to run faster)
+  * Python modules required for plotting: matplotlib, seaborn
+  * Required for generating LaTeX tables: a LaTeX distribution which provides
+    pdflatex, and the following packages: amsmath, amssymb, booktabs, calc,
+    geometry, mathtools, multicol, multirow, rotating, sparklines, xspace.
+    The TeX Live distribution should be fine.
+
+### Installing on Debian systems
+
+The following command will install all dependencies on Debian-based systems:
```
ltratt: All "mandatory" or all "mandatory and optional"?

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
...
+
+```sh
+$ sudo apt-get install build-essential python2.7 pypy bzip2 libssl-dev \
+       pkg-config libcurl4-openssl-dev python-numpy python-matplotlib \
+       python-seaborn python-pip texlive-latex-extra wget xorg-dev \
+       libreadline-dev libbz2-dev liblzma-dev libpcre3-dev gfortran \
```
ltratt: Incorrect trailing `\`?

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
...
+
+
```
ltratt: Double blank line?

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
...
+$ ./build_stats.sh
+```
+
+### Setting up R
```
ltratt: Do we need this section, given that `build_stats.sh` does the necessary stuff? If so, we should say clearly "if you've used `build_stats.sh` you can skip this section."

README.md
ltratt: We should either use underscores for both, or for neither. That's just confusing.

I tend to agree about "proc_exec" too.

README.md
```diff
@@ -34,6 +33,121 @@ Special notes:
...
+
+The scripts here take CSV files as input. The format is as follows. The first row
+must contain a header with a process execution id, benchmark name and sequence
+of iteration numbers. Subsequent rows are data rows, one per process execution.
+The in-process iteration index columns should contain the time in seconds for
+the corresponding in-process iteration. Each process execution must execute the
+same number of iterations as described in the header. For example:
+
+```
+    process_exec_num, bench_name, 0, 1, 2, ...
```
ltratt: Ah, these don't match the text above.

ltratt:
I think we might to rethink the README layout a bit. At the moment there's lots of installation stuff (including two sections on R installation?) before getting to the "how to run it" bits. Putting all the one-time information at the top of the file feels a bit like we're optimising for the wrong thing.

We could either: make an `INSTALL` file with just the install instructions; or restructure the `README`. I don't feel strongly about which we do.

If we choose the restructuring route, I'd suggest starting with a section called something "quick guide" which says: "Run `build_stats.sh` to install. Then output CSV files in this format. Then run `warmup_stats` like this." For most people, that will be all they need.

snim2:
If that is everything, this will now need squashing again.

bin/csv_to_krun_json
ltratt: Did we sort this out? The defaults still seem to be in there?

bin/csv_to_krun_json
```diff
@@ -63,35 +32,21 @@ def create_cli_parser():
...
+    parser.add_argument('--uname', '-u', dest='uname', action='store', default='',
+                        type=str, help='uname -a string from benchmarking machine.')
     return parser
 
 
 if __name__ == '__main__':
     parser = create_cli_parser()
     options = parser.parse_args()
-    main(options.csv_files[0], options.language, options.vm, options.uname)
+    _, _ = main(options.csv_files[0], options.language, options.vm, options.uname)
```
ltratt: Remove the `_, _ =`?

bin/mark_changepoints_in_json
```diff
@@ -7,16 +7,29 @@ import argparse
...
+
+# We use a custom install of rpy2, relative to the top-level of the repo.
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                'work', 'pylibs'))
+# R packages are stored relative to the top-level of the repo.
+os.environ['PATH'] = ':'.join([str(os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                 'work', 'R-inst', 'bin')), os.environ.get('PATH', '')])
+os.environ['LD_LIBRARY_PATH'] = ':'.join([str(os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                 'work', 'R-inst', 'lib', 'R', 'lib')), os.environ.get('LD_LIBRARY_PATH', '')])
+os.environ['R_LIBS_USER'] = str(os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
```
ltratt: `os.path.join` returns a string, so we can remove the `str(...)` calls I think.

warmup/summary_statistics.py
```diff
@@ -0,0 +1,299 @@
...
+                first_steady_segment = len(data_dictionaries[machine]['changepoint_means'][key][p_exec]) - 1
+                num_steady_segments = 1
+                last_segment_mean = data_dictionaries[machine]['changepoint_means'][key][p_exec][-1]
+                lower_bound = min(last_segment_mean * (1.0 - half_bound), last_segment_mean - delta)
+                upper_bound = max(last_segment_mean * (1.0 + half_bound), last_segment_mean + delta)
+                for index in xrange(len(data_dictionaries[machine]['changepoint_means'][key][p_exec]) - 2, -1, -1):
+                    current_segment_mean = data_dictionaries[machine]['changepoint_means'][key][p_exec][index]
+                    if current_segment_mean >= lower_bound and current_segment_mean <= upper_bound:
+                        first_steady_segment -= 1
+                        num_steady_segments += 1
```
ltratt, line 67: Aren't these two variables effectively the same quantity? Can we track just one in the for loop, then create the other afterwards (if it needs to be broken out separately)?

warmup/summary_statistics.py
```diff
@@ -0,0 +1,299 @@
...
+                if data_dictionaries[machine]['changepoints'][key][p_exec]:
+                    steady_iter = data_dictionaries[machine]['changepoints'][key][p_exec][first_steady_segment - 1]
+                    steady_iters.append(steady_iter)
+                    to_steady = 0.0
+                    for index in xrange(steady_iter):
+                        to_steady += data_dictionaries[machine]['wallclock_times'][key][p_exec][index]
+                    time_to_steadys.append(to_steady)
+                else:  # Flat execution, no changepoints.
+                    time_to_steadys.append(0.0)
+            # Get overall and detailed categories.
```
ltratt, line 84: Would it be better to say `categories_set = set(categories)` rather than sometimes calculating it twice?

warmup/summary_statistics.py
```diff
@@ -0,0 +1,299 @@
...
+
+def write_latex_table(machine, all_benchs, summary, tex_file, num_splits, with_preamble=False):
+    """Write a tex table to disk"""
+
+    num_benchmarks = len(all_benchs)
+    all_vms = sorted(summary.keys())
+    num_vms = len(summary)
+
+    # decide how to lay out the splits
+    num_vms_rounded = \
```
ltratt: Not sure this line needs to be split over two?

warmup/summary_statistics.py
```diff
@@ -0,0 +1,299 @@
...
+        int(math.ceil(num_vms / float(num_splits)) * num_splits)
+    vms_per_split = int(num_vms_rounded / float(num_splits))
+    splits = [[] for x in xrange(num_splits)]
+    vm_num = 0
+    split_idx = 0
+    for vm_idx in xrange(num_vms_rounded):
+        try:
```
ltratt: `all_vms` is, I think, a list, so this would be clearer and quicker as a `len` check?

ltratt:
OK, a few mostly minor comments, but we're almost there!

snim2:
I can't reply to your "Did we sort this out? The defaults still seem to be in there?" because GH won't give me a reply button, but the original comment of this nature was a comment on a different script (where the default have been removed).

ltratt:
OK.

warmup/summary_statistics.py
snim2: I don't really follow this comment. `num_vms_rounded` is half the length of `len(all_vms)` if `num_splits` is 2, for example.

warmup/summary_statistics.py
snim2: That may be so, but I think the two variables are slightly clearer, and this code is a little tricky already.

warmup/summary_statistics.py
ltratt: But `vm_idx` is a number and `all_vms` is a list (right? or maybe I got that bit wrong!), so the try/catch is just confusing AFAICT.

ltratt:
There are quite a few `str(os.path.join(...))`s. Can we remove all of them?

warmup/summary_statistics.py
snim2: I'm really not sure I follow, how are you intending the code to end up?

ltratt: Surely it's just an `if vm_idx < len(all_vms)` type job?

warmup/summary_statistics.py
ltratt: OK.

ltratt:
So that looks my comments are covered. Before we squash this down, have you run a couple of examples through `warmup_stats` so that we can check we haven't accidentally broken something? If it's all OK, then you're welcome to squash before... gulp... I merge :)

snim2:
I have run a couple of tests, but I haven't nuked everything, run the build script again or tried the VM

ltratt:
We haven't touched the build script, so I'm only suggesting we try running `warmup_stats` itself.

snim2:
Well, I've tried the three features in warmup_stats (plots, latex, JSON) and they are all still working for me. Ready to squash?

ltratt:
Yep, let's squash!

snim2:
Squashed.

PR #336, opened 2017-02-28
Don't do any automatic processing on octane or dacapo data.

### LINKED ISSUE #334: Minor octane glitch on OpenBSD
```
bench-octane: build-krun build-vms
	PYTHONPATH=krun/ LD_LIBRARY_PATH=${GCC_LIB_DIR}:${LIBKRUN_DIR} ${PYTHON} extbench/runoctane.py
	bin/csv_to_krun_json -u "`uname -a`" -v V8 -l JavaScript octane.v8.results
	bin/csv_to_krun_json -u "`uname -a`" -v SpiderMonkey -l JavaScript octane.spidermonkey.results
```

The last line crashes on OpenBSD as we skip spidermonkey there, My initial instinct was to kill these conversion lines altogether, since we can use our shiny new frontend to abstract these details (and the other script invocations this target fails to run) away. However is it useful to have the `-u` and `-v` filled in automatically.

Shall we make a new target which runs the new frontend (passing the right uname etc) to make the all of the possible outputs (tables, html, plots...)?

vext01:
Fixes #334. If the user wants to process the CSVs then they use the bin/warmup frontend (or othter scripts directly if they know what they are doing).

I do wonder if we should dump the uname out into a file in case the user wants to run the analysis on a different machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ endif
 LIBKRUN_DIR=${PWD}/krun/libkrun
 
 CC=${PWD}/work/gcc-inst/bin/zgcc
+R_LD_LIBRARY_PATH=${PWD}/work/R-inst/lib/R/lib
 
 all: build-benchmarks build-startup
 	@echo ""
@@ -70,13 +71,9 @@ bench-startup-with-reboots: build-startup build-benchmarks
 
 bench-dacapo: build-krun build-vms
 	PYTHONPATH=krun/ JAVA_HOME=${JAVA_HOME} LD_LIBRARY_PATH=${GCC_LIB_DIR} ${PYTHON} extbench/rundacapo.py
-	bin/csv_to_krun_json -u "`uname -a`" -v Graal -l Java dacapo.graal.results
-	bin/csv_to_krun_json -u "`uname -a`" -v HotSpot -l Java dacapo.hotspot.results
 
 bench-octane: build-krun build-vms
 	PYTHONPATH=krun/ LD_LIBRARY_PATH=${GCC_LIB_DIR}:${LIBKRUN_DIR} ${PYTHON} extbench/runoctane.py
-	bin/csv_to_krun_json -u "`uname -a`" -v V8 -l JavaScript octane.v8.results
-	bin/csv_to_krun_json -u "`uname -a`" -v SpiderMonkey -l JavaScript octane.spidermonkey.results
 
 plot-warmup-results:
 	bin/mark_outliers_in_json -w ${WINDOW_SIZE} -t ${OUTLIER_THRESHOLD} warmup_results.json.bz2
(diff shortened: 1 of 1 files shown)


vext01:
Why don't we kill those targets? I thought the idea was that the user uses bin/warmup?

snim2:
If you prefer then sure, just so long as we don't leave broken targets in the Makefile

vext01:
If Laurie is OK with this, works for me.

ltratt:
I don't mind what we do.

vext01:
Wait, there's a lot of different plots in those targets. Do we need them for the paper?

snim2:
For the paper we have a script called `make_figures.sh` in the paper repo

ltratt:
[Personally I think we should do the smallest thing that doesn't interfere with the wider infrastructure we've developed. But I'm happy to leave this to your discretion.]

vext01:
Something like this is probably what we are after. If it looks ok, I will test.

snim2:
Looks sensible to me

vext01:
I think the targets are broken.

```
$ make plot-dacapo-results 
...
bin/mark_changepoints_in_json -s 500 -p dacapo.hotspot_outliers_w200.json.bz2
usage: Write changepoints and classifications into Krun results
file(s). If you want outliers to be excluded from the changepoint calculations,
you should first run your data through the ./bin/mark_outliers_in_json script.

This script does not alter your original Krun results file. Instead it writes
out a new file, with _changepoints added to the filename. For example if the
input file is:
...
```

This script doesn't accept -p. It was removed in: 
https://github.com/softdevteam/warmup_experiment/commit/cfe4d016a77f88890e0e01bb83cd3efcdac38449#diff-7c270985de044cc0148849b75f4d82faL324

I think that is the changepoint merging option? @snim2 Can I kill that switch in the invocation? Do you see anything else fishy in these targets while we are here?

snim2:
I think the rest is probably OK, but yes, definitely kill `-p`

vext01:
Ugh, despite installing devtools and changepoint, I have:
```
vext01@bencher2:~/research/warmup_experiment$ make plot-dacapo-results 
bin/csv_to_krun_json -u "`uname -a`" -v HotSpot -l Java dacapo.hotspot.results
bin/mark_outliers_in_json -w 200 -t 8 dacapo.hotspot.json.bz2
Marking outliers with sliding window size: 200
Loading: dacapo.hotspot.json.bz2
Writing out: dacapo.hotspot_outliers_w200.json.bz2
bin/mark_changepoints_in_json -s 500 dacapo.hotspot_outliers_w200.json.bz2
Marking changepoints and classifications.
Expecting a steady state to be reached before the last 500 iterations.
Using a half bound of 0.001% and a fixed bound of 0.001s.
/home/vext01/.local/lib/python2.7/site-packages/rpy2/rinterface/__init__.py:186: RRuntimeWarning: Error in loadNamespace(name) : there is no package called ‘changepoint’

  warnings.warn(x, RRuntimeWarning)
  File "bin/mark_changepoints_in_json", line 253, in <module>
    main(options.json_files[0], options.half_bound, options.delta, options.steady_state)
  File "bin/mark_changepoints_in_json", line 123, in main
    cpt = rpy2.interactive.packages.importr('changepoint')
  File "/home/vext01/.local/lib/python2.7/site-packages/rpy2/interactive/packages.py", line 31, in importr
    packinstance = _importr(packname, on_conflict = 'warn')
  File "/home/vext01/.local/lib/python2.7/site-packages/rpy2/robjects/packages.py", line 453, in importr
    env = _get_namespace(rname)
rpy2.rinterface.RRuntimeError: Error in loadNamespace(name) : there is no package called ‘changepoint’

Makefile:90: recipe for target 'plot-dacapo-results' failed
make: *** [plot-dacapo-results] Error 1
```

any ideas?

snim2:
Yes, you need to fix your env vars:

```
export LD_LIBRARY_PATH=.../warmup_experiment/work/R-inst/lib/R/lib:$LD_LIBRARY_PATH
```

snim2:
Incidentally, you should be able to just run `build_stats.sh`, export the new env var and run the scripts directly.

vext01:
Will give it a go

vext01:
This seems to work. Tested with and without missing data. Please check carefully.

snim2:
This looks sensible to me.

ltratt:
I'm fine with this. Do you want to squash before I merge?

vext01:
splat

ltratt (approved):

PR #340, opened 2017-03-01
Truncated confidence intervals

snim2:
This is a first go at the truncated (at zero) confidence intervals described in the recent email. This first commit deals with Steady Iter (#) and Steady Perf (s) columns in the LaTeX tables, the time to reach a steady state data and formatting the HTML tables is still to do.

This is the current output:

  * [bencher3_full.pdf](https://github.com/softdevteam/warmup_experiment/files/810583/bencher3_full.pdf)
  * [octane_full.pdf](https://github.com/softdevteam/warmup_experiment/files/810584/octane_full.pdf)


vext01:
Looks good to me.

vext01:
What may help is having the top-level warmup script invoke sub-scripts with `sys.exectutable` thus using the same interpreter as the top-level script did. Then it's up to the user to invoke the right interpreter. Just a thought.

snim2:
That's an interesting thought, I always forget about `sys.executable`.

snim2:
Latest tables:

  * [bencher3_full.pdf](https://github.com/softdevteam/warmup_experiment/files/811629/bencher3_full.pdf)
  * [bencher5_full.pdf](https://github.com/softdevteam/warmup_experiment/files/811630/bencher5_full.pdf)
  * [bencher6_full.pdf](https://github.com/softdevteam/warmup_experiment/files/811636/bencher6_full.pdf)
  * [octane_full.pdf](https://github.com/softdevteam/warmup_experiment/files/811637/octane_full.pdf)
  * [dacapo_full.pdf](https://github.com/softdevteam/warmup_experiment/files/811638/dacapo_full.pdf)

ltratt:
I don't get it. Are the CIs tuples (low, high)? Why is low always 0?

snim2:
So the advice was to use "truncated" CIs, so if the lower bound of the CI dips below 0, then we truncate that bound to 0, and leave the upper bound as it originally was. Apparently it is standard to notate this as a tuple. Where the CI does not dip below 0 I have left it as it was.

ltratt:
I am thoroughly confused. If I look at the bencher3 table with steady iter (s), nearly all the CIs are of the format (0, X) (with only 3 exceptions). What information is this giving me? Surely it can't be the case that *all* of these went below 0?

snim2:
This is a bit fiddly, so I'll try to be pedantically clear. 

If I print out the original (mean, error) values for that particular column, I get:

```
171.6783214 16.1687517
1.1202129 1.3054334
23.45364465 46.0020654
0.0020606 1.525e-05
41.4981153 81.7563615
39.6375719 0.0029553
948.5680029 6.20826725
14.3541823 21.14795915
0.971173 0.76069605
14.7604954 2.78043655
24.0494171 23.5690903
533.0211796 0.10718895
244.7590998 226.2743906
0.74947215 0.0354493
912.8859707 1.73691715
0.1595189 2.28e-05
19.93026835 0.1616295
92.2092862 3.18384485
26.81720505 26.0178468
46.2511747 1.058316
827.3925381 186.34715735
2.87993085 3.71188865
166.7571653 163.3372428
109.2599009 162.397294
49.21019845 15.75857315
121.74922225 2.3234641
166.5621226 248.1477018
952.0490524 4.2094498
701.4641915 207.24550525
0.843543 3.9e-05
```

So you are right in that the `mean - error` is often not less than zero. 

However, this is the time to reach a steady state which is the sum of the times of all the iterations before the (mean) steady state iteration. 

I remembered from our previous conversations that it was important that the lower bound of the time to reach a steady state did not imply that any particular iteration ran in less than 0 seconds. So the relevant code is this:

```python
            # Check whether the error is wider than the median, OR
            # whether the lower CI of the time to reach a steady state
            # implies that one or more of the early iterations took
            # < 0 seconds to complete.
            if (error_time_to_steady > median_time_to_steady or
                (median_time_to_steady - error_time_to_steady) / median_iter < 1.0):
```  

If there is a bug here, it is probably here.

ltratt:
Can the check not be simply `if error - mean < 0: CI = 0`?

snim2:
Possibly, this is the result of making that change on the bencher3 data: [bencher3_simpler.pdf](https://github.com/softdevteam/warmup_experiment/files/811749/bencher3_simpler.pdf)

I think I need to sleep on this before we go any further.

ltratt:
I think I vaguely see what the more complex check is trying to do, but I think it's very hard to defend -- most notably, what relationship does the median iteration have with the first few iterations (in the general case, no relation whatsoever, I think).

The simple check I suggested does have (at least) one annoying property: if you have lots of warmups and few flats, you can get slightly unintuitive results. But even that, I think, is defensible. So maybe being simple is the right thing to do?

snim2:
Perhaps I wasn't clear, the "median iter" is the mean of the first iteration to reach a steady state. So, what I was *trying* to say here was if the low CI on the average time to reach a steady state is `T`, and the table says it took `I` iterations to reach `T`, make sure that none of those iterations had to run in less than zero seconds to make the numbers work.

I think I probably mangled it though.

ltratt:
There's some part of that logic that I can't follow. If `T` is > 0 (which, unless all pexecs are flat, it will be), then `T / I` will always be > 0.

I think the calculation you initially provided does something quite a bit different. I think I understand your rough intent, but I don't think the code necessarily lives up to that intent. Indeed, I think that it would be quite difficult to meet that intent.

[I also have other problems with the initial suggestion. Since different pexecs reach a steady state at a different time, it's not clear to me that dividing their time in seconds by the mean number of iterations gives you an accurate idea of the iteration time. If nothing else, I think flats will mess this up significantly.]

snim2:
OK, these are the tables with the simpler truncation:

  * [octane_full.pdf](https://github.com/softdevteam/warmup_experiment/files/813533/octane_full.pdf)
  * [dacapo_full.pdf](https://github.com/softdevteam/warmup_experiment/files/813536/dacapo_full.pdf)
  * [bencher3_full.pdf](https://github.com/softdevteam/warmup_experiment/files/813539/bencher3_full.pdf)
  * [bencher5_full.pdf](https://github.com/softdevteam/warmup_experiment/files/813540/bencher5_full.pdf)
  * [bencher6_full.pdf](https://github.com/softdevteam/warmup_experiment/files/813541/bencher6_full.pdf)

warmup/summary_statistics.py
```diff
@@ -124,12 +124,7 @@ def collect_summary_statistics(data_dictionaries, half_bound, delta, steady_stat
                     else:
                         error_iter = (error_iter,)
                     median_time_to_steady, error_time_to_steady = bootstrap_confidence_interval(time_to_steadys)
-                    # Check whether the error is wider than the median, OR
-                    # whether the lower CI of the time to reach a steady state
-                    # implies that one or more of the early iterations took
-                    # < 0 seconds to complete.
-                    if (error_time_to_steady > median_time_to_steady or
-                        (median_time_to_steady - error_time_to_steady) / median_iter < 1.0):
+                    if error_time_to_steady > median_time_to_steady:
```
ltratt: Should this be `>` or `>=`? I suspect it should be `>` but I'm not entirely sure.

snim2: Good question. If the steady iter mean is 2, and the error is 2, then the lower bound of the ci is 0. That probably makes sense for steady iter and the time to reach a steady state. I'm not sure that it makes sense to say that the lower bound of the steady state perf is 0, as that would suggest that an iteration can complete in 0s.

ltratt: I tend to agree with that: can we have `>` for steady iter (#) and `>=` for steady iter (s) easily enough?

snim2:
Progress on this PR is waiting for a meeting to take place.
**Note to self** HTML output needs to match LaTeX output.

snim2:
Following the meeting, it looks like we won't need this PR. However, I'll keep the branch open until we have merged a new PR.

PR #344, opened 2017-03-02
Report summary macros for good/bad inconsistency.

### LINKED ISSUE #339: Add "good" and "bad" inconsistent macros to macro summary script

snim2:
Fixes #339 

Output of v0.7 data: [summary_macros.txt](https://github.com/softdevteam/warmup_experiment/files/813730/summary_macros.txt)

### DIFF
--- a/bin/create_summary_macros
+++ b/bin/create_summary_macros
@@ -16,9 +16,10 @@ from warmup.krun_results import read_krun_results_file
 NUMBERS = {0:'zero', 1:'one', 2:'two', 3:'three', 4:'four', 5:'five',
            6:'six', 7:'seven', 8:'eight', 9:'nine'}
 BLANK_CLASSES = {'flat': 0, 'warmup': 0, 'slowdown': 0, 'no steady state': 0}
-BLANK_CONSISTENT = {'consistent': 0, 'inconsistent': 0}
+BLANK_CONSISTENT = {'consistent': 0, 'good inconsistent': 0, 'bad inconsistent': 0}
 BLANK_CLASSES_PLUS_INCONSISTENT = {'flat': 0, 'warmup': 0, 'slowdown': 0,
-                                   'no steady state': 0, 'inconsistent': 0}
+                                   'no steady state': 0, 'good inconsistent': 0,
+                                   'bad inconsistent': 0}
 
 
 def main(data, latex_file):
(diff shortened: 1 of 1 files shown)


ltratt:
Can you stick in a quick PDF of table 1? My ability to read this sort of LaTeX is low ;)

snim2:
I wasn't quite sure what you wanted in terms of the paper and tables. This is a first effort: [table1_from_paper.pdf](https://github.com/softdevteam/warmup_experiment/files/814591/table1_from_paper.pdf)

I have given you macros here for "inconsistent (both types)", "good inconsistent" and "bad inconsistent". The sparklines make that a bit confusing in the table, so this probably isn't *exactly* what you want for the final paper.

ltratt:
I think we want just "good inconsitent" and "band inconsistent" in the paper (the two combined isn't really helpful, because the two types are very different IMHO), which should also make the (VM, benchmark) pair columns add up to 100%.

snim2:
Possibly a silly question, but the paper we have things like this:

```latex
On \bencherthree
\bencherthreevmbenchpairsslowdownpercentage\% of \vmbpair pairs slowdown
and \bencherthreevmbenchpairsinconsistentpercentage\% are inconsistent
(with over two thirds of those being obviously `bad'); \bencherfive and
\benchersix are broadly similar.
```

are you going to want to re-write that text, or should I leave the `\inconsistent` macros in?

ltratt:
Factoring out good and bad inconsistency will mean we need to rewrite quite a few of those bits. The basic problem is that we assume that bad inconsistency is just that: definitely bad. However, good inconsistency is a little fuzzier: it *could* be bad, but we're giving JITs the benefit of the doubt (because it could be the result of limitations in our classification).

snim2:
Next table: [table1_from_paper_no_inconsistent.pdf](https://github.com/softdevteam/warmup_experiment/files/814652/table1_from_paper_no_inconsistent.pdf)

You might notice that the numbers here are not the same as the numbers in the paper. I've had a look at this, and I'm pretty sure that the version of the macros in the paper included the empty benchmark, and that's why they are different.

ltratt:
So I think we can merge this one? Does it need squashing?

snim2:
Yes...

snim2:
Ok, all squashed

ltratt (approved):

PR #345, opened 2017-03-06
Specify where R libraries should be stored.

snim2:
Users who already have a local R installation will have libraries stored in a subdirectory of ~/R. This fix ensures that the changepoint library is installed relative to the warmup_experiment repo, and not in ~/R.

### DIFF
--- a/build_stats.sh
+++ b/build_stats.sh
@@ -27,8 +27,8 @@ export PATH=${R_INST_DIR}/bin:${PATH}
 export LD_LIBRARY_PATH=${R_INST_DIR}/lib/R/lib:${LD_LIBRARY_PATH}
 
 # Install R changepoint package.
-echo "install.packages('devtools', repos='http://cran.us.r-project.org')" | ${R_INST_DIR}/bin/R --no-save
-echo "devtools::install_github('rkillick/changepoint')" | ${R_INST_DIR}/bin/R --no-save
+echo "install.packages('devtools', lib='${R_INST_DIR}/lib/R/library', repos='http://cran.us.r-project.org')" | R_LIBS_USER=${R_INST_DIR}/lib/R/library/ ${R_INST_DIR}/bin/R --no-save
+echo "devtools::install_github('rkillick/changepoint')" | R_LIBS_USER=${R_INST_DIR}/lib/R/library/ ${R_INST_DIR}/bin/R --no-save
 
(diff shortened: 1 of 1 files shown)


vext01:
Does this fix the interpreter path bug too?

snim2:
Interpreter path bug?

vext01:
Should we check if the R library path already exists?

> Interpreter path bug?

You said there was some kind of caching of which R interpreter is used, right?

snim2:
Ah, this is the fix for that bug. The `export R_LIBS_USER` ensures that `devtools::install_github` uses the correct directory to install the changepoint library. There is a new switch to `install.packages` which ensures that the devtools library is installed in the right place.

ltratt:
I'll leave this one to you two.

build_stats.sh
```diff
@@ -25,9 +25,10 @@ cd ..
 # Correct PATH and LD_LIBRARY_PATH variables.
 export PATH=${R_INST_DIR}/bin:${PATH}
 export LD_LIBRARY_PATH=${R_INST_DIR}/lib/R/lib:${LD_LIBRARY_PATH}
+export R_LIBS_USER=${R_INST_DIR}/lib/R/library/  # Needed by devtools::install
```
ltratt: If this variable is purely local to this script (which I think is the case), we don't need (and shouldn't use) `export`.

snim2:
@vext01 ok, I think this is ready for testing!

vext01:
This seems to work. Tested on bencher2 with some octane data, built R afresh.

P.S. The JSON output has a load of trailing spaces. Is that the Python JSON library?

snim2:
I think it must be Python, I'm just doing:

```python
    if options.output_json:
        info('Generating JSON.')
        with open(options.output_json, 'w') as fd:
            json.dump(summary, fd, sort_keys=True, ensure_ascii=True, indent=4)
        debug('Written out: %s' % options.output_json)
```

so, nothing terribly clever happens at our end.

I have tested this on bencher 2, with R already installed / an existing `~/R/` directory, and you have tested on bencher 2, presumably without an `~/R/` directory. Do you think that is "enough" testing?

vext01:
I think this is fine. Would you like to squash?

snim2:
Squished.

vext01 (approved):

PR #346, opened 2017-03-06
Update startup table script

snim2:
This PR brings the startup script up to date with new APIs. 

It also simplifies the CLI options so that user does not have to provide machine names with input files (the user used to have to say `bin/table_startup_results ... bencher3:filename.json.bz2`). 

A `--with-preamble` option has been added, to match other scripts (with the switch a full LaTeX file is written out, and this can be compiled, without it only the table is written out, to be copied and pasted into a paper or similar).

### v0.7 output

  * Script run from master branch: [startup_master_branch.pdf](https://github.com/softdevteam/warmup_experiment/files/821705/startup_master_branch.pdf)
  * Script run from feature branch: [startup_new.pdf](https://github.com/softdevteam/warmup_experiment/files/821707/startup_new.pdf)

### DIFF
--- a/bin/table_startup_results
+++ b/bin/table_startup_results
@@ -6,7 +6,6 @@ import argparse
 import os
 import os.path
 import sys
-from collections import OrderedDict
 
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
 from warmup.krun_results import read_krun_results_file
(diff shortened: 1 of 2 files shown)


ltratt:
Externally speaking, the only visible difference is this is split over two pages? So it's just the internal changes that I should worry about.

snim2:
Yes, the PDF output is only for testing, so being split over two pages doesn't matter. Any difference in the actual tables would be a bug.

vext01:
This LGTM.

vext01 (approved):

snim2:
Last commit changed machine names to the macros we use in the paper: [startup_machine_names.pdf](https://github.com/softdevteam/warmup_experiment/files/822490/startup_machine_names.pdf)

Needs squashing.

ltratt:
OK, makes sense.

PR #347, opened 2017-03-14
Remove 'could not classify' from legend.

snim2:
mark_changepoints script now throws an exception if it cannot classify, so 'could not classify' cannot show up in any tables. This changes the legend in the table documents, not the table itself.

### Test case

  * Was: [old_b7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/842155/old_b7_table.pdf)
  * Now: [bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/842161/bencher7_table.pdf)

### DIFF
--- a/warmup/latex.py
+++ b/warmup/latex.py
@@ -3,7 +3,6 @@ _NUMBERS = {0:'zero', 1:'one', 2:'two', 3:'three', 4:'four', 5:'five',
 
 
 STYLE_SYMBOLS = {  # Requires \usepackage{amssymb} and \usepackage{sparklines}
-    'could not classify': '$\\bot$',
     'flat': '\\flatc',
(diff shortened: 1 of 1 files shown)


ltratt:
Can you can confirm that nowhere else in the code has the string "could not classify" or relies on that concept in anyway? If so, I'm happy to merge this.

snim2:
The only string I could find is in the exception:

```python
                try:
                    classifications[bench].append(segments.get_classification())
                except ValueError:
                    print 'Could not classify %s execution %d' % (bench, index + 1)
                    sys.exit(1)
```

ltratt:
That sounds fine to me.

ltratt (approved):

PR #350, opened 2017-03-21
New table iqrs

### LINKED ISSUE #343: Simplify warmup/kalibera.py
This module has a large number of features that we are not likely to use, and is very hard to read. The functions and methods we use are:

- `bootstrap_confidence_interval(self, iterations=10000, confidence="0.95")`
- ` bootstrap_means(self, iterations=1000)`
- ` _bootstrap_sample(self)`
- ` mean(self, indicies=())`
-  `_mean(l)`
-  `_bootstrap_sample(self)`
-  `confidence_slice(means, confidence="0.95")`
-  `confidence_slice_indicies(length, confidence_level=Decimal('0.95'))`
-  `ConfRange(tuple)`

snim2:
This PR removes the bootstrapped confidence intervals from tables and replaces it with a (5%, 95%) IQR and (to show the spread of values) histogram in each table cell.

### Example v0.8 bencher7: 

  * Before: [bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/858664/bencher7_table.pdf)
  * After: [iqr_bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/858670/iqr_bencher7_table.pdf)

Fixes #343


ltratt:
Interesting stuff. One problem with this is that often one has no idea of the x axis start/end because the boxes are too small. I wonder if violin plots would solve that?

vext01:
This is very cool Sarah.

It shows that pretty much none of the results are normally distributed too.

@ltratt you know roughly the x axis from the IQR figures, but I do agree that the diagrams are a little small.

vext01:
Looking at the size of the header cells, seems we do have a little more horizontal real estate in those cells?

snim2:
The histograms should be left-aligned, but they certainly don't look it. The LaTEx LOKS LIKE THIS:

```
\begin{array}{rlrl}
\scriptstyle{1.22} \\[-6pt]
\scriptscriptstyle{(1.126,}\\[-6pt]
\scriptscriptstyle{1.358)}\\
{\begin{sparkline}{2}
\sparkspike 0.10 0.07
\sparkspike 0.20 0.87
\sparkspike 0.30 0.03
\sparkspike 0.40 0.00
\sparkspike 0.50 0.00
\sparkspike 0.60 0.00
\sparkspike 0.70 0.00
\sparkspike 0.80 0.00
\sparkspike 0.90 0.00
\sparkspike 1.00 0.03
\end{sparkline}}
\end{array}
$
```

and there are always 10 bins in the histograms.

vext01:
Put a delimeter either side of the sparkline so you can see how big it's bounding box is?

snim2:
OK, with delimiters: [delimiters.pdf](https://github.com/softdevteam/warmup_experiment/files/858721/delimiters.pdf)

vext01:
Ah.

```
\begin{array}{rlrl}
```

Says, make me an array with 4 *columns*, aligned `lrlr`. You only ever use column 1, which is left aligned.

snim2:
oooooh :(

snim2:
So this is each cell right-aligned: [right-aligned.pdf](https://github.com/softdevteam/warmup_experiment/files/858762/right-aligned.pdf)

ltratt:
BTW, I'm now pretty convinced we should rotate the whole table by 90 degrees and put histograms/violinplots/whatever to the RHS of the numbers they're showing the distribution of. This will cut down the huge amount of whitespace we're wasting by having things portrait.

snim2:
Like this: [sideways.pdf](https://github.com/softdevteam/warmup_experiment/files/858819/sideways.pdf) ?

vext01:
I think he means, not literally rotate, but swap the roles of the x and y headers

snim2:
You mean have VMs as columns and steady state etc. as rows?

vext01:
I'll let Laurie answer, as I'm not 100% sure.

ltratt:
That's just rotated portrait mode: I mean make the table landscape and put the histograms/violin things *next to* (not underneath) the numbers.

snim2:
Do you want the column / row headers swapped round?

snim2:
OK, I think this is closer to what you are looking for: [sparklines_on_right.pdf](https://github.com/softdevteam/warmup_experiment/files/861139/sparklines_on_right.pdf)

snim2:
This version has improved spacing between cells: [sparklines_on_right_better_spacing.pdf](https://github.com/softdevteam/warmup_experiment/files/861153/sparklines_on_right_better_spacing.pdf)

vext01:
Looks good. I think you can afford to make your sparklines wider (see blank page margins), and a little taller (to match the cell height).

ltratt:
Aha, this is much better, and also solves the "I don't know wide the x axis is" problem (which I didn't expect).

I agree with Edd that we can now make use of the extra width, although I think the first thing to do is to flatten the arrays: there's no point them being multiline when we have horizontal space to spare (and when we're spilling over the vertical space boundary a bit). Then, when that's done, we can make the sparklines wider if there's space remaining.

snim2:
OK, this is where we are currently "at": [flatten_iqrs.pdf](https://github.com/softdevteam/warmup_experiment/files/861348/flatten_iqrs.pdf)

There is now a red dot at the top of the histogram bin that the median fell into. You might think it is a bit cluttered, but it shows something quite useful, I think.

vext01:
I like the median dot. How many decimal places should we use for the IQR figures?

snim2:
I kept the number of decimal places the same as we had it when we used CIs (always 1 more dp than the median above). One issue with the current tables here is that where there is no variance at all (e.g. V8 spectralnorm, steady iter) the table entry looks a bit silly. That said, it looks a lot better in this format than the vertical.

vext01:
I see. I thought it odd that some cells had more DPs than others, but I see the logic now.

Does this table fit within the page margin? It looks quite wide. If it isn't, I think we should make it. Once I tried to sneak a wide table into a paper, and at the editing stage I got burned by the publisher's editing process (where they scaled it down, rendering the text barely readable).

snim2:
Is there a clever way to find out (like drawing the margin)?

vext01:
Delete a few rows for a moment, then put a block of text above the table. You'll see the margin where the block wraps.

vext01:
Even better: https://texblog.org/2011/04/04/make-page-margins-visible/

snim2:
We have two slightly different needs here; in the stand-alone PDFs we are already changing the margins with: `\usepackage[margin=1.0cm]{geometry}`. For those documents the tables are fine. 

For the tables we export for the paper (where we just export the `tabular` and nothing else) we are way over the margin. I'll have a go at fixing that...

ltratt:
I like the PDF, although I don't know what the red blobs are. Also, should the y axis of the sparklines be scaled to the top?

snim2:
The red blob marks the histogram bin that the median fell into. I *think* that the height of the sparklines are roughly the same size as the table cell, If you look at V8 spectralnorm, where there is only 1 histogram bin it is almost the same size as the numbers next to it. I don't think we can do better without setting the table row height explicitly.

ltratt:
Rather than a red blob can we make that bar red? Also for many of the sparklines the highest bar isn't very high, so I don't think they're scaled?

ltratt:
Ah, is the y axis scaled across *all* sparklines? I was thinking (but maybe I'm stupid) that it would be scaled per-sparkline.

vext01:
Which to use depends if you want to compare histograms I suppose.

snim2:
The histograms are all the same size, so the taller bars are closer to 30 and the lower bars closer to zero.

I'm just working on making the bin with the median a red bar. Re-sizing the table for the paper is more of a struggle though.

ltratt:
How wide is each bin? Is it something like `(end - start) / num_bins`?

snim2:
You have to specify a number between 0 and 1 for the x-position of each bin, so I've made all the bins the same size, e.g.:

```latex
\begin{sparkline}{3}
\sparkspike 0.10 0.90 
\sparkspike 0.20 0.03 
\sparkspike 0.30 0.00 
\sparkspike 0.40 0.00 
\sparkspike 0.50 0.03 
\sparkspike 0.60 0.00 
\sparkspike 0.70 0.00 
\sparkspike 0.80 0.00 
\sparkspike 0.90 0.00 
\sparkspike 1.00 0.03 
\end{sparkline}
```

ltratt:
Right, but how do we scale those against our data?

snim2:
I'm not sure what you mean, so let me explain the whole process. 

We ask numpy for a histogram, and we tell it (arbitrarily) to use 10 bins. We create the sparkline with width `3ex` for then whole histogram. Each bar in the histogram is sized between 0 and 1, so we normalise the histogram that came from numpy get the height of each bar. The height of a `1.0` size bar is set by a macro in the sparklines package, by default it is 1.75ex.

ltratt:
OK, that makes sense. Thanks!

snim2:
Next version: [thinner_bars.pdf](https://github.com/softdevteam/warmup_experiment/files/862470/thinner_bars.pdf)

This fixes the "bars are so thick they overlap" bug. It also fixes the "can't-find-the-median" bin, which means that every histogram should have a red bar. Instead of taking the median and trying to find which bin it was placed in, we now use the histogram to estimate where the median will be, similarly to the algorithm described in [this SO answer](http://math.stackexchange.com/a/894986/177858).

ltratt:
I notice these are still using the old good/bad macros (which we thought were a bit... not good).

ltratt:
Otherwise I think this looks good.

snim2:
I haven't tested this extensively enough yet to start squashing, but one issue is that because we need to think about is that this PR currently changes the JSON output of `warmup_stats`, because we needed to add some extra detail to the summary statistics, which is the list of original data from which we generated the medians, etc.

Here is an example: https://gist.github.com/snim2/afd696ccf80e68950c1405ea71877b9b

If we don't want to expose this to users, I will need to update the APIs of a few functions.

Da Capo example: [dacapo_sparklines.pdf](https://github.com/softdevteam/warmup_experiment/files/862712/dacapo_sparklines.pdf)

ltratt:
I'm fine with the JSON stats changing, provided it's backwards compatible (new keys shouldn't affect anyone, but changing keys means thinking more carefully).

snim2:
Do we want to draw the x-axis, to indicate the extent of the histogram? It would look like this: [draw_x_axis.pdf](https://github.com/softdevteam/warmup_experiment/files/864270/draw_x_axis.pdf)

ltratt:
Good question. I *think* I like the a axis, but it's now quite difficult to distinguish "one entry in a bucket" from "no entries in a bucket". Two things that immediately spring to mind: a light grey x axis; or increasing the height of the sparklines. In fact, we could do that second thing anyway: currently sparklines at their maximum possible height are quite a bit shorter than the cell height.

vext01:
Are the ticks on the x-axis only drawn if there is no sample there?

snim2:
@vext01 yes

vext01:
That's fine.

snim2:
OK, how about this: [grey_stretched.pdf](https://github.com/softdevteam/warmup_experiment/files/864417/grey_stretched.pdf)

snim2:
@ltratt strictly speaking we have changed the JSON more than that, for example keys like `steady_state_iteration_confidence_interval` used to be a single float and are now a 2-tuple containing the IQR.

ltratt:
Well, since it isn't a CI, we should probably not call the key "steady_state_iteration_confidence_interval" any more. Not that that helps backward compatibility, but at least it'll be more obvious to people that something has changed.

Maybe we can add a key to the top-level of the JSON called something like "warmup conf format='0.1" or similar?

snim2:
OK, I will change the keys to say `_iqr` rather than `confidence_interval`. For `conf format` do you mean this should represent which percentages we chose for the IQR?

ltratt:
No, the format thing would say "this file is in format version X", so that if we change the format people can at least automatically say "wah, I don't understand this format".

snim2:
Oh, a version number. OK, will do that next and re-test.

snim2:
New JSON format: https://gist.github.com/snim2/95c3b07fc3364b0031c21c6598f93901

I've tested this with the v0.8 main experiment, v0.7 da capo, and all output formats with `warmup_stats`.

So, I think this might be finally ready for review.

warmup/kalibera.py
```diff
@@ -1,367 +0,0 @@
-#
```
vext01, line 1: I highly approve of this diff hunk.

snim2: :)

vext01:
The PR is quite large, so my review is perhaps superficial. I mainly wanted to see that you had killed the old kalibera code and checked the JSON version number. You've nailed both.

The stats code is a bit of a black box for me though.

warmup/summary_statistics.py
```diff
@@ -104,14 +105,14 @@ def collect_summary_statistics(data_dictionaries, half_bound, delta, steady_stat
             elif categories_set == set(['flat']):
                 median_iter, error_iter = None, None
                 median_time_to_steady, error_time_to_steady = None, None
-                median_time, error_time = bootstrap_confidence_interval(steady_state_means)
+                median_time, error_time =  median_iqr(steady_state_means)
```
ltratt: Two spaces after `=`? (also a couple of instances below).

warmup/summary_statistics.py
```diff
@@ -136,6 +141,9 @@ def collect_summary_statistics(data_dictionaries, half_bound, delta, steady_stat
 
 
 def convert_to_latex(summary_data, half_bound, delta, steady_state):
+    assert 'json_format_version' in summary_data and summary_data['json_format_version'] == JSON_VERSION_NUMBER, \
+        'Cannot process data from old JSON formats.'
+    del summary_data['json_format_version']
```
ltratt: Deleting the key seems pretty ugly (and likely to lead to confusion to one day). I assume this is because we want to keep the line `machine = summary_data.keys()[0]` as-is? Maybe there's a way to phrase that line that doesn't involve deleting the format key?

warmup/summary_statistics.py
```diff
@@ -304,6 +307,9 @@ def write_latex_table(machine, all_benchs, summary, tex_file, num_splits, with_p
 
 
 def write_html_table(summary_data, html_filename):
+    assert 'json_format_version' in summary_data and summary_data['json_format_version'] == JSON_VERSION_NUMBER, \
+        'Cannot process data from old JSON formats.'
+    del summary_data['json_format_version']
```
ltratt: Same comment as earlier. [Which also means that one can't produce HTML and LaTeX output in one run!]

ltratt:
OK, I have a few minor comments.

snim2:
Hopefully that addresses everything.

snim2:
The key is the machine name, not the string "machine", so if I use the uname from my laptop, the key is `snim2-X1`

ltratt:
Ah, I see. That's a slightly unfortunate design choice, in retrospect, but it's probably too late to change it now.

OK, so from my perspective this can be squashed. Edd?

vext01:
Nothing more from my side. Go ahead.

snim2:
Splat.

vext01 (approved):

PR #351, opened 2017-03-21
Clamp off by one and index fix

vext01:
The crash we were discussing on MM is tiggered by the code that `clamps the extent of the inset`. I think this code makes a data-set one sample too big.

I don't  fully understand that logic, so please check carefully. My brain hurts.

Good news is, once this is applied, it does see the crashed pexec:
```
WARNING: requested pexec crashed: Linux$_\mathrm{E3-1240v5}$, N-Body, C, 3
```

But the index should be 0, not 3. Fixed that too.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -682,9 +682,9 @@ class ProcessExecChart(object):
                                         + int(one_pc * 2.0)))
             # Clamp the extent of the inset.
             if inset_xlimit[1] < INSET_MIN_ITERS:
-                inset_xlimit = (0, min(INSET_MIN_ITERS, self.x_bounds[1] - self.x_bounds[0]))
+                inset_xlimit = (0, min(INSET_MIN_ITERS, self.x_bounds[1] - self.x_bounds[0] - 1))
             elif inset_xlimit[1] > INSET_MAX_ITERS:
-                inset_xlimit = (0, min(INSET_MAX_ITERS, self.x_bounds[1] - self.x_bounds[0]))
+                inset_xlimit = (0, min(INSET_MAX_ITERS, self.x_bounds[1] - self.x_bounds[0] - 1))
             inset.set_xlim(inset_xlimit[0], inset_xlimit[1], auto=False)
             # Plot a subset of the data on the inset.
             start, end = self.x_bounds[0] + inset_xlimit[0], self.x_bounds[0] + inset_xlimit[1] + 1
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -282,7 +282,7 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
                     else:
                         if data == page:  # Stops repeated printing of warning.
                             print("WARNING: requested pexec crashed: "
-                                  "%s, %s, %s, %s" % (mc, bmark, vm, index + 1))
+                                  "%s, %s, %s, %s" % (mc, bmark, vm, i))
                 return ret
 
             wct_page = only_uncrashed(page)


vext01:
Well, actually, the index that is printed is the index of the plot on the page, not the pexec number, so maybe we can do  better there.

bin/plot_krun_results
```diff
@@ -682,9 +682,9 @@ class ProcessExecChart(object):
                                         + int(one_pc * 2.0)))
             # Clamp the extent of the inset.
             if inset_xlimit[1] < INSET_MIN_ITERS:
-                inset_xlimit = (0, min(INSET_MIN_ITERS, self.x_bounds[1] - self.x_bounds[0]))
+                inset_xlimit = (0, min(INSET_MIN_ITERS, self.x_bounds[1] - self.x_bounds[0] - 1))
```
snim2, line 685: In the common case for our data, `self.x_bounds[0]` will be 0 and `self.x_bounds[1]` will be 2000, so your change should be fine because the minimum will always be `INSET_MIN_ITERS`. 

In the case where the user has passed in `--xlimits` and `self.xlimits[1] - self.xlimits[0] < 50` then the minimum won't be `INSET_MIN_ITERS`, so this is the case where we need to check the off-by-one. 

I don't fully understand what the test data looks like, but if there is no data for this pexec, then `self.xbounds == (0, 0)` and so your `-1` will make `inset_xlimit == (0, -1)`, which seems wrong to me. But maybe that isn't the test case?

vext01:
The data causing the issue is here:
http://theunixzoo.co.uk/random/example_results.json.bz2

The problem is just triggered by small pexecs that will invoke the clamping code. It doesn't appear to be related to crashing pexecs.

```
 683                 # Clamp the extent of the inset.                                                                        
 684                 if inset_xlimit[1] < INSET_MIN_ITERS:                                                                   
 685                     inset_xlimit = (0, min(INSET_MIN_ITERS, self.x_bounds[1] - self.x_bounds[0]))                       
 686                 elif inset_xlimit[1] > INSET_MAX_ITERS:                                                                 
 687                     inset_xlimit = (0, min(INSET_MAX_ITERS, self.x_bounds[1] - self.x_bounds[0]))                       
 688                 inset.set_xlim(inset_xlimit[0], inset_xlimit[1], auto=False)                                            
 689                 # Plot a subset of the data on the inset.                                                               
 690                 start, end = self.x_bounds[0] + inset_xlimit[0], self.x_bounds[0] + inset_xlimit[1] + 1                 
 691                 import pdb; pdb.set_trace()                                                                             
 692  ->             inset.plot(range(inset_xlimit[0], inset_xlimit[1] + 1),                                                 
 693                            self.wallclock_data[start:end], color=LINE_COLOUR,                                           
 694                            linewidth=PLOT_LINE_WIDTH, zorder=ZORDER_DATA)                                         
```

```
(Pdb++) self.wallclock_data[start:end]
array([ 1.000038,  1.000041,  1.000038,  1.000041,  1.000039])
(Pdb++) c
<boom>
``` 

So there's definitely data, but 5 samples for a 6 tick x-axis. Hence the crash.
```
(Pdb++) range(inset_xlimit[0], inset_xlimit[1] + 1)
[0, 1, 2, 3, 4, 5]
```
If you see what I mean?

snim2:
I think what you are saying is that we are correctly skipping the crashed pexec, but we have some tiny pexecs where the number of iterations is less than `INSET_MIN_ITERS`. In which case, I think we need to change the `INSET_MIN/MAX_ITERS` from strict numbers to percentages of the available iterations (so, `INSET_MIN_ITERS` would be `0.25`.

Does that sound sensible?

vext01:
Um, I'm not sure. We need to make sure that if we "clamp", the number of elements in the x-axis match the number of samples for the y-axis.

snim2:
So, in the failing case, how many iterations are there in the pexec?

vext01:
5, but we pass an x-axis with 6 elements, see above.

snim2:
OK, I've spent some time testing this and I think you are right, this is the easiest fix. Do you want to squash?

vext01:
Na, they are two separate fixes. Go ahead as is.

snim2 (approved):

PR #354, opened 2017-03-22
Don't obtain the length of the pexecs from a potentially crashed [] pexec

### LINKED ISSUE #352: In the event that the first process execution crashed, the plot script cannot run
```
    first_key = data['data'].keys()[0]
    first_mc = data['data'][data['data'].keys()[0]].keys()[0]
    first_pexec = data['data'][first_key][first_mc][0]
    if options. window_size >= len(first_pexec):  # Assume all execs are the same length
        fatal_error('Window size (%d) must be strictly less than the number '
                    'of iterations in each process execution (%d). Use '
                    '-w or --window on the command line.' %
                    (options.window_size, len(first_pexec)))
```

Is broken since a crash would put in the results file, an empty list of length 0. This means there is no `-w` that you could use to make the script run.

vext01:
Fixes #352, but blocked on #351 for proper testing. DONT MERGE YET.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -1357,14 +1357,29 @@ if __name__ == '__main__':
                             options.outliers, options.unique_outliers,
                             options.changepoints or options.changepoint_means)
 
-    first_key = data['data'].keys()[0]
-    first_mc = data['data'][data['data'].keys()[0]].keys()[0]
-    first_pexec = data['data'][first_key][first_mc][0]
-    if options. window_size >= len(first_pexec):  # Assume all execs are the same length
+    # Find the number of in-proc iterations in a non-crashed pexec
+    # Assumes we use the same number of in-proc iterations for all pexecs.
+    try:
+        for key_data in data['data'].itervalues():
+            for bench_data in key_data.itervalues():
+                for pexec in bench_data:
+                    if pexec == []:
+                        # indicates a crash
+                        continue
+                    else:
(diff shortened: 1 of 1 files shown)


vext01:
#351 is merged, so ready for review.

Tested on small examples. Should probably test on some real data. Will do that now.

bin/plot_krun_results
```diff
@@ -1357,14 +1357,29 @@ if __name__ == '__main__':
...
+            for bench_data in key_data.itervalues():
+                for pexec in bench_data:
+                    if pexec == []:
+                        # indicates a crash
+                        continue
+                    else:
+                        iter_lens = len(pexec)
+                        raise StopIteration()  # to break out of all loops at once
+        else:
+            print("couldn't find a non-crashing pexec")
```
snim2: We've used single quotes in the rest of the script.

snim2:
OK, this looks sensible, I'll give it a test now.

bin/plot_krun_results
vext01: Ah, I can never get used to that. fixed.

snim2:
OK, this seems good to me, squash away!

vext01:
splat.

vext01:
I plotted the data from bencher6 fine using this patch too.

snim2:
OK, cool. I was using b3 v0.7 and your crashed test case.

snim2 (approved):

PR #370, opened 2017-03-27
Autocorrelation

vext01:
Raising for discussion only, you wouldn't want to merge this code.

Here's a screenshot showing some likely correlations:
j

I think I'd like the auto-correlation plots in the main plot stack and probably with an indication of the steady iter.

### DIFF
--- a/bin/plot_acorrs.py
+++ b/bin/plot_acorrs.py
@@ -14,6 +14,7 @@ sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(
     os.path.abspath(__file__)))))
 from warmup.krun_results import parse_krun_file_with_changepoints
 from warmup.summary_statistics import collect_summary_statistics
+from scipy.stats.mstats import normaltest
 
 CORR_THRESHOLD = 0.5  # flag correlation coefficients above this value
 N_LAGS = 40           # No. of lags to analyse for correlations
@@ -28,7 +29,6 @@ def main(data_dct, classifier):
 
     summary_stats = collect_summary_statistics(data_dct, classifier['delta'],
                                                classifier['steady'])
-
     for machine, machine_data in data_dct.iteritems():
         for key in machine_data['wallclock_times'].keys():
             bench, vm, _ = key.split(":")
@@ -137,11 +137,12 @@ def analyse(data, key, steady_idxs, machine):
             # Normality Analysis
             # XXX prints all for now
 
-            from scipy.stats.mstats import normaltest
-            import math
+            # cut off bottom and top 25% to "zoom in" on histogram
             n_bins = 30
+            chop = int(float(n_bins/4))
+            n_bins -= chop * 2
             _, pval = normaltest(steady_iters_np)
-            hist = numpy.histogram(steady_iters_np, bins=n_bins)
+            hist = numpy.histogram(sorted(steady_iters_np)[chop:-chop], bins=n_bins)
 
             print("machine=%s, key=%s, pexec=%s, steady_iter=%s" % (machine, key, pnum, steady_iter))
             print(n_bins * "-")
--- a/bin/plot_acorrs.py
+++ b/bin/plot_acorrs.py
@@ -1,8 +1,4 @@
 #!/usr/bin/env python2.7
-"""
-usage:
-    XXX
-"""
 
 import sys
 import os
@@ -20,15 +16,51 @@ from warmup.krun_results import parse_krun_file_with_changepoints
 from warmup.summary_statistics import collect_summary_statistics
 from warmup.plotting import zoom_y_min, zoom_y_max
 
-CORR_THRESHOLD = 1.1, 2.9
-CORR_PLOT_DIR = "correlated_plots"
-NOCORR_PLOT_DIR = "uncorrelated_plots"
+CORR_THRESHOLD = 1.0, 3.0       # threshold for DW analysis
+MIN_CORR_SEG_LEN = 50           # shortest seg to apply DW to
 
 # debug bits
 DEBUG = False
 if os.environ.get("CORR_DEBUG"):
     DEBUG = True
 N_LAGS = 40
+CORR_PLOT_DIR = "correlated_plots"
+NOCORR_PLOT_DIR = "uncorrelated_plots"
+
+
+def get_steady_segs(cpts, steady_idx, pexec_len):
+    """Get a list of steady segments.
+
+    Each steady segment is a tuple (start, end), inclusive of start and
+    exlusive of end. Start and end are list indicies.
+    """
+
+    segs = []
+    start = 0
+    num_cpts_in_steady_state = 0  # used in sanity checks only
+    for end in cpts:
+        if start < steady_idx:
+            pass  # not yet steady
+        else:
+            num_cpts_in_steady_state += 1
+            segs.append((start, end))
+        start = end
+
+    # And add the final segment
+    segs.append((start, pexec_len))
+
+    # Sanity checks
+    assert len(segs) >= 1, 'should be at least one steady segment'
+    assert len(segs) == num_cpts_in_steady_state + 1, \
+        'should be 1 more seg than changepoint'
+    assert segs[-1][1] == pexec_len, 'last seg should end with pexec'
+    assert segs[0][0] == steady_idx, 'first seg should start at steady_idx'
+    num_segs = len(segs)
+    for seg_idx in xrange(len(segs)):
+        if seg_idx < num_segs - 1:
+            assert segs[seg_idx][1] == segs[seg_idx + 1][0], \
+                'end seg idx should be start of next'
+    return segs
 
 
 def main(data_dct, classifier):
@@ -38,8 +70,6 @@ def main(data_dct, classifier):
 
     summary_stats = collect_summary_statistics(data_dct, classifier['delta'],
                                                classifier['steady'])
-    # XXX print per-machine summaries
-    # XXX check all segs in steady state
     for machine, machine_data in data_dct.iteritems():
         for key in machine_data['wallclock_times'].keys():
             bench, vm, _ = key.split(":")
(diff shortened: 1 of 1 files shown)


vext01:
These are plots for the steady state only btw.

bin/plot_acorrs.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def draw_acrs(pdf, data, key, steady_idxs):
+    n_execs = len(data)
+    n_rows, n_cols = n_execs / 2, 2
+    fig = plt.figure(figsize=(20, 45))
+
+    for num, execu in enumerate(data):
+        mean = float(sum(execu)) / len(execu)
+        steady_iter = steady_idxs[num]
```
ltratt: If the plot in question is "no steady state", then steady_iter will be `None` (I think) and then the line below will crash?

vext01: Argh, damnit python:
```
In [1]: range(10)[None:]
Out[1]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

Wut is this madness!

ltratt: Now that I did not know.

vext01: I suppose you think of None as, no bound.

ltratt: Ah, the slice may get converted into function arguments, and `None` would then be indistinguishable from "no parameter passed".

ltratt: Either way, just so we agree, this means this bit of code is wrong, right? :)

vext01:
This now analyses correlations, giving for each pexec a lowdown like this:
```
------------------------------------------------------------------------------
spectralnorm:LuaJIT:default-lua, pexec 17, steady_iter 1157
------------------------------------------------------------------------------
*                                        
                     *                   
       *      *             *      *     
        *    * *      *           *      
 *   **  *          *     ** **     *    
  ***     ***   ****   ***     ***   ****

Min x: 0 Max x: 40 Min y: -0.424015022152 Max y: 1.0
------------------------------------------------------------------------------
Details:
  pexec=17, lag=7, corr=0.62389001965
  pexec=17, lag=14, corr=0.612652673481
  pexec=17, lag=21, corr=0.744359965697
  pexec=17, lag=28, corr=0.650224172228
  pexec=17, lag=35, corr=0.659764439801
```

And then for all pexecs summarises like this:

```
Summary:
  Correlation threshold: 0.5
  Number of lags: 40
  Total num pexecs: 1380
  Total num steady pexecs: 1029
  Total num steady pexecs showing correlation: 318
  Percent of steady correlated pexecs: 30.9037900875
```

So this says, of the 1029 pexecs reaching a steady state, 318 (~30%) showed an absolute correlation coefficient of > 0.5 in at least 1 of the 40 lags we tested.

The ascii plots are kind of a debugging aid. We'd probably want to remove them or find a better licensed library for those.

vext01:
As Sarah says in mattermost, should we instead/also directly check if the samples are normally distributed using (e.g.):
https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.mstats.normaltest.html

bin/plot_acorrs.py
```diff
@@ -0,0 +1,123 @@
...
+    number of stable pexecs for which one or more lag is correlated above the
+    threshold.
+    """
+
+    n_execs = len(data)
+    num_corr_pexecs = 0
+    num_steady_pexecs = 0
+
+    for pnum, execu in enumerate(data):
+        steady_iter = steady_idxs[pnum]
```
ltratt: It might be clearer at this point to say `if not steady_iter: continue` rather than having to indent the big block starting at line 82?

vext01:
Here are some manually inspected examples where the script flagged correlation. I used interactive mode, which is why the plots look different.

### 1



1a

### 2



2a


###3



3a


The first two clearly show correlation. The last one less so, although I can sympathise with the algorithm. The steady state wavers above and below the mean, whereas real noise would jump above and below the mean more readily.

I think we should try doing the normality tests.

vext01:
To be clear, the above plots come in pairs, one zoomed out, one zoomed to a section of the steady state.

snim2:
Do the zoom plots help much here?

ltratt:
Agreed, the first two look reasonable; the third, at least to the naked eye, not really. [I mean, it *could* be ACR on top of a sin pattern of sorts, but at some point one can't help but feel it's really just noise.]

bin/plot_acorrs.py
```diff
@@ -63,7 +68,17 @@ def main(data_dct):
...
 
-def analyse_correlations(data, key, steady_idxs):
+    print("  Skew threshold: +/-%s" % SKEW_THRESHOLD)
+    print("  Kurtosis threshold: +/-%s" % KURT_THRESHOLD)
+    print("  Total num steady pexecs approximately normal: %s" % total_num_normal)
+    if total_num_normal:
+        percent = float(total_num_normal) / total_num_steady * 100
+    else:
+        percent = "N/A"
+    print("  Percent of steady pexecs also normal: %s" % percent)
```
snim2: I think the "also" here is a bit confusing, maybe stick with "approximately", as above?

vext01: yup

vext01:
The code in this branch now attempts to find how many stable pexecs are normal, except, I'm only computing the last segment. Will fix this next, then post some output. Probably tomorrow.

vext01:
The following is the output from the normality checks on 0.8 bencher5 with a p-value of 0.01:
http://theunixzoo.co.uk/random/norm.txt

summary:
```

Summary:
  Correlation threshold: 0.5
  Number of lags: 40
  Total num pexecs: 1380
  Total num steady pexecs: 1028
  Total num steady pexecs showing correlation: 320
  Percent of steady correlated pexecs: 31.1284046693
  Normality p-value threshold: 0.01
  Total num steady pexecs approximately normal: 18
  Percent of steady pexecs approximately normal: 1.75097276265
```

As the P-value goes up, the number of pexecs with normal steady iterations goes down (does that sound right sarah?), and I believe 0.01 is already a quite conservative p-value (I think 0.05 is standard). So if I got this right, very few of our pexecs which to reach a steady state are usable for (e.g.) confidence intervals.  

I'm now going to check some of the steady indices by hand to check i got that code right.

I'm surprised distributions with double humps come out normal, but I think @snim2 said that is OK. Can you confirm Sarah?

vext01:
I guess what seems weird is that there is no obvious difference between the histograms considered normal, and for those not. I was expecting to be super obvious...

snim2:
The histogram comes from samples of real data, so it isn't expected to look *exactly* like a normal distribution from artificial data. Maybe there are better tests for normality than the one we are using, but that's something to ask a statistician.

ltratt:
1.75% doesn't seem plausible, if I'm honest. I suspect we've made an incorrect assumption somewhere, but I admit that I have no idea where.

vext01:
fwiw, here's a sample normal distribution in the same histogram format:
```
------------------------------------------------------------------------
               *              
             ** **            
           **     **          
          *         *         
        **           **       
********               *******

Min x: 0 Max x: 29 Min y: 2 Max y: 978
------------------------------------------------------------------------
```

vext01:
I wouldn't be surprised if we got < 5% normal, but I also wouldn't be surprised if there's an error in the code. More investigation after lunch.

ltratt:
I suppose one obvious question is whether the bucket size is right, and whether 500 samples and 30 buckets is enough to get a good answer from the test?

snim2:
The test doesn't depend on the buckets BTW, but clearly the visualisation may be misleading with some funny bin sizes.

ltratt:
Ah, I see.

Still, looking at the ASCII art, there are a lot of things that are pretty clearly fairly close to normal -- which, given the amount of data we've got, are surely close enough to normal -- that are classified as non-normal. That said, the overall trend does seem to be that at least 50% look non-normal, which is surprising and interesting.

ltratt:
I'm not quite sure how I should read entries like this:

```
machine=bencher5, key=richards:V8:default-javascript, pexec=16, steady_iter=4
------------------------------
*                             
                              
                              
                              
                              
 *****************************

Min x: 0 Max x: 29 Min y: 0 Max y: 1995
------------------------------
pval=0.0
^^^ IS NOT NORMAL
```

which looks like the x axis isn't scaled properly or something? There are quite a lot like this.

snim2:
There are 30 asterisks there, so each one is a bin. Does that help?

ltratt:
All but one bins appears to be empty, so I can't really read it properly.

ltratt:
I wonder if this skew is being caused by outliers? Should we take those out of the normality check since they'll obviously mess it up?

vext01:
Would we recommend users remove outliers from their steady samples to make stats? Seems a bit questionable...

vext01:
I don't think outliers would mess up the normality check. Outliers should be infrequent and have little influence.

It's occurred to me that these histograms are not totally useful. In the one laurie pointed out, the x-axis is like that because there were a handful of data points to the right hand side of the main bulk of data. This renders the main bulk of data as a spike, which if you had zoomed in on could be normal, bimodal, ...

Note that the normality check function does not use these histograms, and has no notion of bins (at least not visible to the user). The histograms are something I cobbled together thinking they might be useful, but now i'm not sure.

vext01:
Same again with 25% cut off the head and tail in the histogram:
http://theunixzoo.co.uk/random/norm2.txt

ltratt:
The current outlier check seems to require that the bell curve is both perfectly central and perfectly symmetric. So a single outlier seems likely to kill it.

That said, I do wonder if we really have enough samples for this. Consider this one:
```
machine=bencher5, key=fannkuch_redux:Graal:default-java, pexec=5, steady_iter=1
----------------
         *      
        * **    
       *        
      *     *   
     *       *  
*****         **

Min x: 0 Max x: 15 Min y: 6 Max y: 282
----------------
pval=0.0
^^^ IS NOT NORMAL
```
I mean, it's pretty clear that's normal from our perspective, with just a teeny tiny little bit of noise and a slight skew to right of centre.

vext01:
OK, i've been doing a lot of reading.

1) First, the x-axes on the histograms are not long enough to see what's going on. I've made them bigger. When you do, you see that indeed, many of the cases are not really anything like normal. I also chopped 1% off the head an tail for the histogram, so as to exclude boring outliers.

2) I've switched to a (hopefully) more robust method:
https://en.wikipedia.org/wiki/Anderson%E2%80%93Darling_test

I'm using a significance of 1% (meaning there is a 1% chance we say a pexec is normal when it is not (this is the lowest we can go with scipy). I'll be asking Rebecca if this all checks out.

3) Comments on lauries comments

> The current outlier check seems to require that the bell curve is both perfectly central
> and perfectly symmetric. So a single outlier seems likely to kill it.

Hrm, I don't think it's that simple. Have a look at the latest output and see if you still feel the same. I tend to agree with most of the outcomes at this point (but I'm not a statistician (tm)).

> That said, I do wonder if we really have enough samples for this

Ironically, it seems these tests tend to work best for fewer samples:

http://stackoverflow.com/questions/23251759/how-to-determine-what-is-the-probability-distribution-function-from-a-numpy-arra
 
```because your sample size is so large, even small deviations will lead to a rejection of the null hypothesis```

(i had another page making this claim, but seem to have lost it. )

I think the anderson method is more robust in this regard, but i need to ask Rebecca.

FWIW, the new output prints the sample size.

4) Equivalent segs make the distribution N-modal.

E.g. if your steady segments are:

```
--------------------------
                          |-------------------------
```

And the two segments are considered equivalent, then you get a bi-modal distribution. If you use only the last segment, you get a few more percent more classed as normal.





Here's the new output:

http://theunixzoo.co.uk/random/norm4.txt

vext01:
I think this is dead? Shall we kill it?

ltratt:
Yes, this PR is very dead. Closing (unless someone really, really wants to reopen it!).

PR #371, opened 2017-03-28
Highlight segments similar to steady state.

### LINKED ISSUE #327: Show visually which changepoints are within a small bound
The classification algorithm now differentiates between segments that are very similar to the last segment, and those that are different. It would be nice if the plotting script also distinguished these two sorts of changepoint / segment.

snim2:
Fixes #327 

The suggestion here was to highlight segments that are equivalent to the steady state in red. This is not nice to look at and it's impossible to see the segment means and changepoints through the data. Instead, this PR draws all data that is *not* equivalent to the steady state in grey. 

This has been especially tricky, so please check carefully.

### Test cases

Master branch:

  * [master_bencher6_plots.pdf](https://github.com/softdevteam/warmup_experiment/files/874114/master_bencher6_plots.pdf)
  * [master_warmup_no_migrate.pdf](https://github.com/softdevteam/warmup_experiment/files/874115/master_warmup_no_migrate.pdf)

Feature branch:

  * [bencher6_plots.pdf](https://github.com/softdevteam/warmup_experiment/files/874116/bencher6_plots.pdf)
  * [new_warmup_no_migrate.pdf](https://github.com/softdevteam/warmup_experiment/files/874119/new_warmup_no_migrate.pdf)

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -59,7 +59,8 @@ GRID_MAJOR_Y_DIVS = 6
 GRID_MINOR_Y_DIVS_SMALLER_PLOTS = 4
 GRID_MAJOR_Y_DIVS_SMALLER_PLOTS = 2
 
-LINE_COLOUR = 'k'
+LINE_COLOUR = 'grey'
+STEADY_COLOUR = 'k'
 MEDIAN_COLOUR = '#0072B2'
 FILL_ALPHA = 0.2
 
@@ -171,8 +172,9 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
     cycles_pages = list()
     instr_pages = list()
     all_outliers, all_common, all_unique = list(), list(), list()
-    all_changepoints, all_changepoint_means = list(), list()
+    all_changepoints, all_changepoint_means, all_changepoint_vars = list(), list(), list()
     all_classifications = list()
+    classifier = data_dcts['classifier']
 
     # By default, each benchmark from each machine is placed on a separate
     # page. If the --one-page switch has been passed in, then we place
@@ -183,7 +185,7 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
         cycles_page = list()
         instr_page = list()
         page_common, page_unique, page_outlier = list(), list(), list()
-        page_changepoints, page_changepoint_means = list(), list()
+        page_changepoints, page_changepoint_means, page_changepoint_vars = list(), list(), list()
         page_classifications = list()
         for key in sorted(data_dcts['data']):
             for machine in sorted(data_dcts['data'][key]):
@@ -211,14 +213,17 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
                     if changepoints:
                         page_changepoints.append(data_dcts['changepoints'][key][machine][index])
                         page_changepoint_means.append(None)
+                        page_changepoint_vars.append(None)
                         page_classifications.append(data_dcts['classifications'][key][machine][index])
                     if changepoint_means:
                         page_changepoints.append(data_dcts['changepoints'][key][machine][index])
                         page_changepoint_means.append(data_dcts['changepoint_means'][key][machine][index])
+                        page_changepoint_vars.append(data_dcts['changepoint_vars'][key][machine][index])
                         page_classifications.append(data_dcts['classifications'][key][machine][index])
                     else:
                         page_changepoints.append(None)
                         page_changepoint_means.append(None)
+                        page_changepoint_vars.append(None)
                         page_classifications.append(None)
         pages.append(page)
         cycles_pages.append(cycles_page)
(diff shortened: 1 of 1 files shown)


vext01:
Presentation-wise, very nice!

bin/plot_krun_results
```diff
@@ -171,8 +172,9 @@ def main(is_interactive, data_dcts, plot_titles, window_size, outfile,
     cycles_pages = list()
     instr_pages = list()
     all_outliers, all_common, all_unique = list(), list(), list()
-    all_changepoints, all_changepoint_means = list(), list()
+    all_changepoints, all_changepoint_means, all_changepoint_vars = list(), list(), list()
```
vext01, line 175: What do we mean by `vars` here?

ltratt:
This is quite a lot different than I was expecting -- not that that's necessarily a bad thing! I was expecting only the segment lines to be different colours, not the time-series data itself.

One challenge with this visualisation is that no steady states are now all grey. I think that's probably OK, and maybe even a good idea, but I need to get used to the idea.

bin/plot_krun_results
```diff
@@ -565,6 +587,48 @@ class ProcessExecChart(object):
                          label=('Common outliers (${%.2f}\\%%$)' % pc_common),
                          zorder=ZORDER_MARKERS)
 
+    def _plot_steady_equivalent(self, axis):
+        """Plot steady state segment and 'equivalents' in a different colour."""
```
vext01, line 591: newline after docstring

snim2:
Well, potentially we could change the segment means rather than the data. Sometimes the segment means are difficult to see though.

vext01 (commented):
couple of comments

bin/plot_krun_results
snim2: Variances

vext01:
> One challenge with this visualisation is that no steady states are now all grey. I think that's > probably OK, and maybe even a good idea, but I need to get used to the idea.

I also picked up on this. Initially I thought it was iffy, but then I realised it quite effectively communicates the fact that the whole 2000 samples are unusable from a benchmarking standpoint!

ltratt:
Yes, I'm growing to like this.

bin/plot_krun_results
```diff
@@ -565,6 +587,49 @@ class ProcessExecChart(object):
...
+
+        if not self.changepoint_means or self.classification == 'no steady state':
+            return
+        if self.classification == 'flat' and len(self.changepoints) == 0:
+            axis.plot(self.iterations,
```
ltratt: The first two params may as well be on one line.

bin/plot_krun_results
```diff
@@ -565,6 +587,49 @@ class ProcessExecChart(object):
...
+                      self.wallclock_data,
+                      color=STEADY_COLOUR, zorder=ZORDER_DATA + 1, linewidth=LINE_WIDTH)
+            return
+        lower_bound = min(self.last_segment_mean - self.last_segment_var,
```
ltratt: Is the second line of parameters supposed to be indented two past the bracket?

bin/plot_krun_results
```diff
@@ -565,6 +587,49 @@ class ProcessExecChart(object):
...
+                            self.last_segment_mean - self.delta)
+        upper_bound = max(self.last_segment_mean + self.last_segment_var,
+                            self.last_segment_mean + self.delta)
```
ltratt: Same indentation comment.

ltratt (approved):
A few cosmetic suggestions.

bin/plot_krun_results
snim2: It was intentional, but I see in this case it doesn't really help readability

snim2:
I realise the b6 test case was short of pages, here is the full document (sorry about the empties): [bencher6_plots.pdf](https://github.com/softdevteam/warmup_experiment/files/875205/bencher6_plots.pdf)

ltratt:
Once the comments are addressed, I think we should merge this.

snim2:
I'm pretty sure all the "cosmetic" comments have been addressed. @vext01 may have more comments to add? I must admit, I am a little nervous about merging changes of this size.

vext01:
I read some of the logic and didn't see anything that looks wrong.

snim2:
OK, so we are ready to squash?

vext01:
I think so

ltratt:
Hang on, we have some unsafe code we need to look at.

bin/plot_krun_results
```diff
@@ -344,17 +356,26 @@ class ProcessExecChart(object):
...
+                self.last_changepoint = None
+            self.last_segment_mean = changepoint_means[-1]
+            self.last_segment_var = changepoint_vars[-1]
             self.segment_means = list()  # ((x0, y0), (x1, y1)) pairs.
-            if self.x_bounds[0] > 0:
-                for index, changepoint in enumerate(self.changepoints):
-                    if changepoint < self.x_bounds[0]:
-                        del self.changepoint_means[index]
-                        del self.changepoints[index]
+            for index, changepoint in enumerate(changepoints):
```
ltratt, line 372: This is potentially unsafe, as we're also adding items to `changepoints` in the loop. Are those new things supposed to be considered by this loop as well? If not, it would be clearer to the reader if a new list was made inside the `for` loop, and then `changepoints.extend(newlist)` after the loop.

snim2: We add to `self.changepoints`, not `changepoints` (the argument to the constructor).

vext01: ah, yes. agreed.

ltratt:
OK, I think we're ready for squashing.

snim2:
Squished.

PR #378, opened 2017-03-30
Fix csv parser

vext01:
@snim2 can you guve this a go?

This fixes the parser for octane by sorting the rows ahead of parsing. It looks like a big change because  block of code got de-dented.

Also make the JSON easier to inspect by hand.

OK?

### DIFF
--- a/warmup/krun_results.py
+++ b/warmup/krun_results.py
@@ -166,7 +166,7 @@ def read_krun_results_file(results_file):
 
 
 def write_krun_results_file(results, filename):
-    """Write a Krun results file to disk.
-    """
+    """Write a Krun results file to disk."""
+
     with bz2.BZ2File(filename, 'wb') as file_:
-        file_.write(json.dumps(results))
+        file_.write(json.dumps(results, indent=4))
--- a/warmup/krun_results.py
+++ b/warmup/krun_results.py
@@ -35,31 +35,42 @@ _SKIP_OUTER_KEYS = ['audit', 'reboots', 'mperf_counts', 'aperf_counts',
 def csv_to_krun_json(in_files, language, vm, uname):
     for filename in in_files:
         data_dictionary = _BLANK_BENCHMARK
+
+        # First sort the lines by benchmark, then pexec number.
+        # We do this so we can easily check for gaps (missing pexecs).
         with open(filename, 'r') as fd:
-            data_dictionary['audit']['uname'] = uname
             reader = csv.reader(fd)
-            header = reader.next()  # Skip first row, which contains column names.
-            expect_idx = [0]  # check we get in-order indices, first always 0
-            for lineno, row in enumerate(reader):
-                # First cell contains process execution number.
-                assert int(row[0]) in expect_idx, \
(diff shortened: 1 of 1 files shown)


snim2:
OK, this works for me, will look at the code.

warmup/krun_results.py
```diff
@@ -35,31 +35,38 @@
...
+            header = reader.next()  # Skip header row.
+            rows = iter(reader)
+            sorted_rows = sorted(rows, key=lambda l: (l[1], int(l[0])))
+
+        data_dictionary['audit']['uname'] = uname
+        expect_idx = [0]  # check we get in-order indices, first always 0
+        for row in sorted_rows:
+            # First cell contains process execution number.
+            assert int(row[0]) in expect_idx, \
+                'found gaps in process executions!'
```
snim2: Can we give a little more detail here? Maybe which benchmark has gaps?

warmup/krun_results.py
```diff
@@ -35,31 +35,38 @@
...
+            bench = row[1]
+            if row[2] == 'crash':
+                data = []
+            else:
+                data = [float(datum) for datum in row[2:]]
+            key = '%s:%s:default-%s' % (bench, vm, language)
```
snim2, line 60: The user may have passed in a language with a `default-` prefix, so this code should check before adding it.

warmup/krun_results.py
```diff
@@ -35,31 +35,38 @@
...
+            if key not in data_dictionary['wallclock_times']:
+                data_dictionary['wallclock_times'][key] = list()
+                data_dictionary['core_cycle_counts'][key] = list()
+                data_dictionary['aperf_counts'][key] = list()
+                data_dictionary['mperf_counts'][key] = list()
+            data_dictionary['wallclock_times'][key].append(data)
+            data_dictionary['core_cycle_counts'][key].append(None)
+            data_dictionary['aperf_counts'][key].append(None)
+            data_dictionary['mperf_counts'][key].append(None)
+            expect_idx = [0, int(row[0]) + 1]  # expect the next index, or a 0
```
snim2: "expect the next index, or a 0" -> "expect the next process execution index, or a 0"

snim2:
OK, just some minor comments.

warmup/krun_results.py
vext01: Hrm. Not 100% sure about this. Arguably this library routine should accept `variant`, not `language`. But let's not worry ourselves too much with this right now. FWIW, I did not change this logic, only indented it.

vext01: Or, de-dented rather.

vext01:
Fixed 2 comments, left a comment on the remaining comment.

warmup/krun_results.py
snim2: I've looked again at the help for this script and `warmup_stats` and we don't advise users to prepend the `default-`, so I think you are right, this should be left as-is.

snim2:
OK, happy to squash?

vext01:
splat.

snim2 (approved):

PR #380, opened 2017-03-30
Allow changepoint-related plot annotations to happen with --wallclo-only and -b.

vext01:
Hi @snim2 I think this is the fix. Just de-dent these two blocks.

Testing a bit now...

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -1233,15 +1233,15 @@ def get_data_dictionaries(json_files, benchmarks=[], wallclock_only=False,
                                     get_instr_data(key, machine, instr_dir, [p_exec])[0])
                             else:
                                 data_dictionary['instr_data'][key][machine] = None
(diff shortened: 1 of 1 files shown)


snim2:
Worth testing with as many combinations of CLI options as you can :(

vext01:
I'm hoping it's harmless enough, as it only ever adds to the data dictionaries... Still... Plot script will find a way to make me sad, I'm sure.

vext01:
Confirmed this only affects runs with `-b` btw.

snim2:
Great, I was just about to ask.

vext01:
I think this is good, but I found an unrelated bug. Will file.

snim2:
Ready to merge?

snim2 (approved):

vext01:
Yeah, go ahead.

PR #384, opened 2017-04-05
Classification used to determine which pexecs are flat.

snim2:
Now that we keep changepoints in the data, even where there are "equivalent" segments, we can no longer use the number of changepoints to tell whether a pexec is flat or not.

### Test cases

  * Before this PR: [master_bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/900629/master_bencher7_table.pdf)
  * After this PR: [bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/900631/bencher7_table.pdf)


ltratt:
Aha, yes, this seems to solve several of the odd things I was seeing. Good catch!

ltratt (approved):

PR #385, opened 2017-04-07
Script to count pexecs given a list of results files.

vext01:
In response to a comment in the paper, here is a script to count how many pexecs we have, and thus how many plots we would have to inspect by hand if we didn't have automation.

The answer is:
```
========================================================================                
GRAND TOTAL: 3660                                                    
========================================================================
```

May as well put this in the repo.

### DIFF
--- a/bin/count_pexecs.py
+++ b/bin/count_pexecs.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python2.7
+
+import sys
+import os
+
+sys.path.insert(0, os.path.join(os.path.dirname(
+    os.path.dirname(os.path.abspath(__file__)))))
+from warmup.krun_results import read_krun_results_file
+
+
+def main(filenames):
+    all_files_total = 0
+    for filename in filenames:
+        print(filename)
+        file_total = 0
+        js = read_krun_results_file(filename)
+
+        for key, val in js['wallclock_times'].iteritems():
+            assert isinstance(val, list)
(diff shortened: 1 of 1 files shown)


ltratt:
Are there any other things we need to count for the paper? If not, this looks good!

ltratt:
One additional comment: this script looks like it's printing *lots* of stuff to stdout. Is that useful? I guess not, in which case I'd just as soon it didn't print anything out other than the total.

vext01:
We can add additional counts later if we need any.

I deliberately printed lots of counts as it lets me verify that there are either 30 or 0 (for skipped) pexecs per key.

ltratt:
Verifying-by-eye sucks: can we not add `asserts`?

vext01:
We could do yeah, but then we'd need a way to specify what to check and this raises questions like "do other users want to vary the number fo pexecs on a per-machine basis?", "do other users want to vary the number of pexecs on a per vm basis?" etc.

It doesn't take long to skim down the list tbh. I made the numbers all line up.

vext01:
I also don't think we should over-engineer this, with so many more important tasks on the stack ;)

ltratt:
Point taken. I'm just a bit allergic to incontinent scripts in general!

Will merge.

ltratt (approved):

vext01:
Thanks

PR #389, opened 2017-04-10
Use 99% confidence intervals for steady perf

snim2:
### Test cases 

  * Bencher 7 table from master branch: [master_bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/910664/master_bencher7_table.pdf)
  * Bencher 7 table from this branch: [bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/910678/bencher7_table.pdf)

  * Startup table using IQRs: [iqr_startup.pdf](https://github.com/softdevteam/warmup_experiment/files/910660/iqr_startup.pdf)
  * New startup table: [means_startup.pdf](https://github.com/softdevteam/warmup_experiment/files/910662/means_startup.pdf)

### DIFF
--- a/warmup/bootstrapper.py
+++ b/warmup/bootstrapper.py
@@ -1,0 +1,85 @@
+#!/bin/env pypy
+
+"""
+This script is designed to be run with PyPy via a pipe.
+
+It has been factored out because the code here is too slow to run on CPython,
+and we need numpy as a dependency in related parts of the code base. Although
+there is a version of numpy for PyPy, we wanted to reduce the number of
+complications that arise for end-users.
+
+It will read JSON format data from STDIN, and will write a comma-separated pair
+of floats (mean, CI) on STDOUT.
+
+Input data should be as per. the needs of the tables for the "main" warmup
+experiment -- i.e. a list of pexecs, each containing a list of (steady state)
+segments, each containing a list of floats.
+
+Much of the code here comes from libkalibera.
+"""
+
+import math
+import random
+
+from decimal import Decimal, ROUND_UP, ROUND_DOWN
+
+
+BOOTSTRAP_ITERATIONS = 100000
+CONFIDENCE_LEVEL = '0.99'  # Must be a string to pass to Decimal.
+
+
+def _mean(data):
+    assert len(data), '_mean() received no data to average.'
+    return math.fsum(data) / float(len(data))
+
+
+def bootstrap_steady_perf(steady_segments_all_pexecs, confidence_level=CONFIDENCE_LEVEL):
+    """This is not a general bootstrapping function.
+    Input is a list containing a list for each pexec, containing a list of
+    segments with iteration times.
+    """
+
+    # How many bootstrap samples do we need from each pexec? We want at least
+    # BOOTSTRAP_ITERATIONS samples over all. If we want 100,000 samples in total
+    # and we have 30 pexecs, we need 3333 samples from each pexec. In total we
+    # will have 3333 * 30 bootstrapped samples, and 3333 * 30 == 99990. So, we
+    # add a 1 here to ensure that we end up with >= BOOTSTRAP_ITERATIONS samples.
+    n_resamples = int(math.floor(BOOTSTRAP_ITERATIONS / len(steady_segments_all_pexecs))) + 1
+    bootstrapped_samples = list()  # Final list of BOOTSTRAP_ITERATIONS resamples.
+
+    for segments in steady_segments_all_pexecs:  # Iterate over pexecs.
+        for _ in xrange(n_resamples):
+            sample = list()
+            for seg in segments:
+                sample.extend([random.choice(seg) for _ in xrange(len(seg))])
+            bootstrapped_samples.append(sample)
+    assert len(bootstrapped_samples) >= BOOTSTRAP_ITERATIONS
+
+    means = sorted([_mean(sample) for sample in bootstrapped_samples])
+
+    # Compute reported mean and confidence interval. Code below is from libkalibera.
+    assert not isinstance(confidence_level, float)
+    confidence_level = Decimal(confidence_level)
+    assert isinstance(confidence_level, Decimal)
+    exclude = (1 - confidence_level) / 2
+    length = len(means)
+    # There may be >1 median index if data is even-sized.
+    if length % 2 == 0:
+        median_indices = (length // 2 - 1, length // 2)
+    else:
+        median_indices = (length // 2, )
+    lower_index = int((exclude * length).quantize(Decimal('1.0'), rounding=ROUND_DOWN))
+    upper_index = int(((1 - exclude) * length).quantize(Decimal('1.0'), rounding=ROUND_UP))
+    lower, upper = means[lower_index], means[upper_index - 1]  # upper is exclusive.
+    median = _mean([means[i] for i in median_indices])  # Reported mean.
+    ci = _mean([upper - median, median - lower])  # Confidence interval.
+    return median, ci
+
+
+if __name__ == '__main__':
+    import json
+    import sys
+    data = json.loads(sys.stdin.readline())
+    results = bootstrap_steady_perf(data)
+    sys.stdout.write(','.join([str(result) for result in results]))
+    sys.stdout.flush()
--- a/warmup/latex.py
+++ b/warmup/latex.py
@@ -241,6 +241,19 @@ $
 """  % (median_s, error_s, _SPARKLINE_WIDTH + 'ex', _histogram(data))
 
 
+def format_median_ci(median, error, data):
+    median_s = '%.5f' % median
+    error_s = '%.6f' % error
+    return """$
+\\begin{array}{r}
+\\scriptstyle{%s} \\\\[-6pt]
+\\scriptscriptstyle{\\pm%s}
+\\end{array}
+$
+\\noindent\\parbox[p]{%s}{%s}
+"""  % (median_s, error_s, _SPARKLINE_WIDTH + 'ex', _histogram(data))
+
+
 def preamble(title, doc_opts=DEFAULT_DOCOPTS):
     return __LATEX_PREAMBLE(title, doc_opts)
 
--- a/warmup/latex.py
+++ b/warmup/latex.py
@@ -127,6 +127,7 @@ DEFAULT_DOCOPTS = '10pt, a4paper'
 
 __LATEX_PREAMBLE = lambda title, doc_opts=DEFAULT_DOCOPTS: """
 \\documentclass[%s]{article}
+\\usepackage{adjustbox}
 \\usepackage{amsmath}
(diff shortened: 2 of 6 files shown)


snim2:
I just noticed that this PR does not use the new bootstrapper for the "other" tables (i.e. Octane and Da Capo) so will push another commit.

ltratt:
I think this could do with @vext01's eyes on it too.

warmup/bootstrapper.py
```diff
@@ -0,0 +1,76 @@
...
+    n_resamples = int(math.floor(BOOTSTRAP_ITERATIONS / len(steady_segments_all_pexecs))) + 1
+    bootstrapped_samples = list()  # Final list of BOOTSTRAP_ITERATIONS resamples.
+
+    for segments in steady_segments_all_pexecs:  # Iterate over pexecs.
+        for _ in xrange(n_resamples):
+            sample = list()
+            for seg in segments:
+                sample.extend([random.choice(seg) for _ in xrange(len(seg))])
+            bootstrapped_samples.append(sample)
+    assert len(bootstrapped_samples) >= BOOTSTRAP_ITERATIONS
```
ltratt, line 56: Why is this `>=`? Shouldn't we know precisely how many it contains?

warmup/bootstrapper.py
```diff
@@ -0,0 +1,76 @@
...
+    if length % 2 == 0:
+        median_indices = (length // 2 - 1, length // 2)
+    else:
+        median_indices = (length // 2, )
+    lower_index = int((exclude * length).quantize(Decimal('1.0'), rounding=ROUND_DOWN))
+    upper_index = int(((1 - exclude) * length).quantize(Decimal('1.0'), rounding=ROUND_UP))
+    lower, upper = means[lower_index], means[upper_index - 1]  # upper is exclusive.
+    median = _mean([means[i] for i in median_indices])  # Reported mean.
+    ci = _mean([upper - median, median - lower])  # Confidence interval.
+    return median, ci
```
ltratt, line 76: The function says "returns mean and CI" but this says "return median and CI"?

ltratt:
This looks mostly OK to me, modulo the two comments above and the fact that we have 3 instances of the "shell out to PyPy" code, which seems like it might be better off in a library somewhere?

warmup/bootstrapper.py
snim2: Say we want 100,000 bootstrap samples, and we have 30 pexecs. From each pexec we need 3333 samples. In the end we will have 3333 * 30 bootstrapped samples, and 3333 * 30 == 99990. 

So, we add a 1 somewhere in the calculations to find out how many samples we need per pexec, and get a few more samples than we asked for.

warmup/bootstrapper.py
snim2: This is from the libkalibera code. We have a sorted list of means, which contains the mean of each bootstrapped sample pexec, we take the median mean for the reported mean.

warmup/bootstrapper.py
ltratt: OK, let's maybe briefly document that (as it is a little confusing)?

warmup/bootstrapper.py
ltratt: I assume we should report the mean of the means? I'm not sure though.

snim2: I'm not sure how that would work. The "middle" mean is related by distance to the means that make up the CI, how would the "mean of the means" be related?

ltratt: Hmm, that's an interesting point. OK, let's go with that, as I certainly can't argue against it at the moment!

warmup/bootstrapper.py
```diff
@@ -0,0 +1,76 @@
+#!/bin/env pypy
+
+"""
+This script is designed to be run with PyPy via a pipe.
```
vext01, line 4: Shall we say why we separated this out? That we need numpy elsewhere and that's annoying under PyPy.

warmup/bootstrapper.py
snim2: Commented on lines 36-37

warmup/bootstrapper.py
```diff
@@ -0,0 +1,76 @@
...
+CONFIDENCE_LEVEL = '0.99'  # Must be a string to pass to Decimal.
+
+
+def _mean(data):
+    return math.fsum(data) / float(len(data))
+
+
+def bootstrap_steady_perf(steady_segments_all_pexecs, confidence_level=CONFIDENCE_LEVEL):
+    """This is not a general bootstrapping functions.
+    Input is a list containing a list for each pexec, containing a list of
```
vext01: Perhaps re-word this sentence. Complete nitpick though.

warmup/bootstrapper.py
ltratt: The current comment doesn't explain this sufficiently -- at least not as well as your GH comment does!

warmup/bootstrapper.py
```diff
@@ -0,0 +1,76 @@
...
+    segments with iteration times.
+    """
+
+    # How many bootstrap samples do we need from each pexec? We want at least
+    # BOOTSTRAP_ITERATIONS samples over all.
+    n_resamples = int(math.floor(BOOTSTRAP_ITERATIONS / len(steady_segments_all_pexecs))) + 1
```
vext01: I don't understand this line. Won't the number of samples we need differ on a per-pexec basis?

warmup/bootstrapper.py
vext01: yes please :)

warmup/bootstrapper.py
```diff
@@ -0,0 +1,76 @@
...
+
+from decimal import Decimal, ROUND_UP, ROUND_DOWN
+
+
+BOOTSTRAP_ITERATIONS = 100000
+CONFIDENCE_LEVEL = '0.99'  # Must be a string to pass to Decimal.
+
+
+def _mean(data):
+    return math.fsum(data) / float(len(data))
```
vext01: assert len(data)?

warmup/latex.py
```diff
@@ -240,6 +241,26 @@ def format_median_error(median, error, data, as_integer=False, brief=False):
 """  % (median_s, error_s, _SPARKLINE_WIDTH + 'ex', _histogram(data))
 
 
+def format_median_ci(median, error, data, as_integer=False, brief=False):
+    if as_integer:
+        median_s = '%d' % int(median)
```
vext01: Is rounding down (as `int()` will) ok here?

warmup/summary_statistics.py
```diff
@@ -57,21 +65,54 @@ def collect_summary_statistics(data_dictionaries, delta, steady_state):
                 # steady state. However, the last segment may be equivalent to
                 # its adjacent segments, so we first need to know which segments
                 # are steady-state segments.
+                if data_dictionaries[machine]['classifications'][key][p_exec] == 'no steady state':
+                    continue
+                # Capture the last steady state segment for bootstrapping.
```
vext01, line 65: I think we could modularise this kind of thing at a later date. I think I wrote similar code in my branch. Shall we leave an XXX?

warmup/summary_statistics.py
```diff
@@ -57,21 +65,54 @@ def collect_summary_statistics(data_dictionaries, delta, steady_state):
...
                     current_segment_var = data_dictionaries[machine]['changepoint_vars'][key][p_exec][index]
                     if (current_segment_mean + current_segment_var >= lower_bound and
                             current_segment_mean - current_segment_var<= upper_bound):
+                        # Extract this segment from the wallclock data for bootstrapping.
+                        segment_data = list()
+                        if index == 0:
+                            start = 0
+                            end = data_dictionaries[machine]['changepoints'][key][p_exec][index] + 1
+                        else:
+                            start = data_dictionaries[machine]['changepoints'][key][p_exec][index - 1] + 1
```
vext01, line 97: Does this deal wit the final segment correctly? There's an implicit changepoint end at the end of the pexec, right?

warmup/summary_statistics.py
```diff
@@ -103,15 +144,41 @@ def collect_summary_statistics(data_dictionaries, delta, steady_state):
...
+                    output = pipe.stdout.readline().strip()
+                    mean_str, ci_str =  output.split(',')
+                    mean_time, error_time = float(mean_str), float(ci_str)
+                except:
+                    print 'Bootstrapper script failed:'
+                    traceback.print_exc()
             else:
-                median_time, error_time = median_iqr(steady_state_means)
+                mean_time, error_time = None, None
+                try:
```
vext01: This code is mostly duplicated above? Would it be practical to use a function?

vext01 (commented):
Some comments.

warmup/bootstrapper.py
snim2: This one should now be resolved....

warmup/summary_statistics.py
snim2: Similar (but not identical) code appears in a number of places, for example, where we choose whether to plot data in grey or black. If we put an XXX in here I would worry that we might miss cases elsewhere. We probably should try to factor it out somehow, but since all the "similar" cases are subtly different, this will need some careful thought.

warmup/summary_statistics.py
snim2: Yes, this does deal with the final segment correctly.

warmup/latex.py
snim2: Hrm, not sure...

warmup/summary_statistics.py
vext01: Yeah, I don't think now is the time, hence suggesting an XXX as opposed to immediate modularisation.

warmup/summary_statistics.py
vext01: ok

snim2:
Results following last changes:

  * New startup table: [new_means_startup.pdf](https://github.com/softdevteam/warmup_experiment/files/913351/new_means_startup.pdf)
  * New bencher7 table: [new_new_bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/913369/new_new_bencher7_table.pdf)

  * New b5 Octane table: [new_bencher5_octane_table.pdf](https://github.com/softdevteam/warmup_experiment/files/913362/new_bencher5_octane_table.pdf)

warmup/bootstrapper.py
vext01: OK, i think I get this part now.

snim2:
OK, are we ready to squash?

vext01:
I think so, if Laurie is OK...

ltratt:
Yep, I'm fine. Edd, I'll leave you to merge this one.

snim2:
Splat.

vext01 (approved):

PR #392, opened 2017-04-11
Report steady state iter as float.

### LINKED ISSUE #391: Integral statistics could be misleading.
There is a section of code in tabling routines where we attempt to give an integral median and IQR for the starting steady state iteration.

Suppose we get a symetric IQR (out of luck) 1.5 +/- 0.1. It would look like this:
```
......|_____________|......
             ^
      1.4   1.5    1.6
```

How would we round this? Well we need to be safe, so we should round the lower bound of the IQR down and the upper bound up, thus over-approximating error. This would give us:
```
......|_____________|......
             ^
      1     1.5      2
```

But what would we do with the median?

If we round down we get:

```
......|_____________|......
      ^
      1              2
```

If we round up we get:

```
......|_____________|......
                    ^
      1              2
```

In neither case is the median in the middle of the error bound, which I find confusing. More confusing that showing a floating point value for the steady state iter.

Hrm...

snim2:
Fixes #391 

**NEEDS SQUASHING**

New bencher 7 table: [bencher7_table.pdf](https://github.com/softdevteam/warmup_experiment/files/913860/bencher7_table.pdf)

### DIFF
--- a/warmup/latex.py
+++ b/warmup/latex.py
@@ -221,16 +221,15 @@ def _histogram(data):
     return '\n'.join(sparkline)
 
 
-def format_median_error(median, error, data, as_integer=False, brief=False):
-    if as_integer:
-        median_s = '%d' % int(median)
-        error_s = '(%d, %d)' % (int(error[0]), int(error[1]))
-    elif brief:
+def format_median_error(median, error, data, one_dp=False, two_dp=False):
+    if one_dp:
+        median_s = '%.1f' % median
+        error_s = '(%.1f, %.1f)' % (error[0], error[1])
+    elif two_dp:
         median_s = '%.2f' % median
         error_s = '(%.3f, %.3f)' % (error[0], error[1])
     else:
-        median_s = '%.5f' % median
-        error_s = '(%.6f, %.6f)' % (error[0], error[1])
+        assert False
     return """$
 \\begin{array}{r}
 \\scriptstyle{%s} \\\\[-6pt]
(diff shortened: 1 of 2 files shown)


ltratt:
I don't understand why we have the "one" and "two" dp options (especially as the latter isn't used anywhere?). Shouldn't we just fix that this function *always* returns a string of the type we want (i.e. one DP; at least, until we change our minds ;) )?

snim2:
The `format_median_error` function is currently used in two places: 

  * to format the steady state iter (#) column data with 1 dp
  * to format the time to reach steady state column with 2/3 dp (this is the `twodp` option in the code)

the default option (which the last commit removes) used to be for the steady state perf columns, which now use a confidence interval.

warmup/latex.py
```diff
@@ -221,16 +221,15 @@ def _histogram(data):
...
+    if one_dp:
+        median_s = '%.1f' % median
+        error_s = '(%.1f, %.1f)' % (error[0], error[1])
+    elif two_dp:
         median_s = '%.2f' % median
         error_s = '(%.3f, %.3f)' % (error[0], error[1])
     else:
-        median_s = '%.5f' % median
-        error_s = '(%.6f, %.6f)' % (error[0], error[1])
+      assert False
```
ltratt: Wrong indentation?

ltratt:
OK, squash then I can merge?

snim2:
Squashed!

ltratt (approved):

PR #393, opened 2017-04-11
Compare our steady state detection to Georges et. al. (2007)

snim2:
Example output: [georges.pdf](https://github.com/softdevteam/warmup_experiment/files/914792/georges.pdf)

### DIFF
--- a/bin/table_georges_comparison
+++ b/bin/table_georges_comparison
@@ -1,0 +1,140 @@
+#!/usr/bin/env python2.7
+
+"""
+Create a LaTeX summary comparing annotated Krun results to those reported by
+the method described in Georges et. al. (2007).
+"""
+
+import argparse
+import numpy
+import os
+import os.path
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+from warmup.krun_results import parse_krun_file_with_changepoints
+from warmup.latex import preamble, end_document, end_table
+from warmup.latex import machine_name_to_macro, STYLE_SYMBOLS
+
+
+_TITLE = 'Comparison with Georges et. al. (2007)'
+_START_TABLE = lambda format_, headings: """
+{
+\\begin{tabular}{%s}
+\\toprule
+%s \\\\
+\\midrule
+""" % (format_, headings)
+
+
+def _start_table(format_, headings):
+    """Unlike the start_table() function in warmup.latex, this one does not
+    contain any code to configure sparklines.
+    """
+
+    return _START_TABLE(format_, headings)
+
+
+def _cov(data):
+    """Computes the coefficient of variation of data."""
+
+    return numpy.std(data) / numpy.mean(data)
+
+
+def compute_comparison(in_files, threshold):
+    """For each classification type, find the number of process executions that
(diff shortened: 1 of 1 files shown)


ltratt (approved):

bin/table_georges_comparison
```diff
@@ -0,0 +1,140 @@
...
+def _start_table(format_, headings):
+    """Unlike the start_table() function in warmup.latex, this one does not
+    contain any code to configure sparklines.
+    """
+
+    return _START_TABLE(format_, headings)
+
+
+def _cov(data):
+    """Computes the Coefficient of Variation of data."""
```
ltratt: Small 'c' in "coefficient"?

bin/table_georges_comparison
```diff
@@ -0,0 +1,140 @@
...
+                print('WARNING: Skipping: %s from %s (benchmark crashed)' %
+                      (key, machine))
+            else:
+                bench, vm, variant = key.split(':')
+                n_pexecs = len(data_dictionaries[machine]['wallclock_times'][key])
+                for pexec in xrange(n_pexecs):
+                    stop = len(data_dictionaries[machine]['wallclock_times'][key][pexec]) - steady
+                    classification = data_dictionaries[machine]['classifications'][key][pexec]
+                    counts[machine]['total ' + classification] += 1
+                    for iteration in xrange(len(data_dictionaries[machine]['wallclock_times'][key][pexec])):
```
ltratt, line 73: We need to think a bit about this, as we're imposing our idea of when the steady state might be onto Georges. For example, if there are 2000 iterations, we say a steady state must have happened by 1500. It *could* be argued that Georges might still find a steady state somewhere between 1500 and 2000. My initial inclination is to give them the benefit of the doubt. Any thoughts?

snim2: Georges also tries to find a steady state before a particular iteration number, so in this script I have retrieved the number we used (500) from the JSON and used the same number for the Georges method -- that is defined on line 70. In their paper they call our "steady" "k": "Suppose that we want to retain k measurements per invocation. ..."

snim2:
This is the relevant part of the paper:

screenshot from 2017-04-12 10-25-45

ltratt:
Yes, but as I read the code above, we stop at 1500 iterations and say "there is no steady state in our methodology, so Georges didn't find one either". Warning: I might be wrong.

snim2:
I'm not sure I follow. In the code we are (hopefully) implementing the method from Georges. In Step 1 we say "we want to retain k=500 iterations". In Step 2 we say "loop over iterations 1 to 1501 and test whether `_cov(wallclock_times[iteration:]) < threshold`, if it is then Georges says that is a steady state".

ltratt:
But what if Georges says it's a steady state after iteration 1500?

snim2:
Like us, Georges wants to retain a certain number of iterations as the minimum length of the steady state (their `k`), so we need to use the same value for `k` for both methods. If we say "Might Georges find a steady state on iteration 1750", then I think we need to ask the same question of our method, otherwise the comparison is skewed.

ltratt:
Good point.

ltratt:
So I think we can merge this after squashing?

snim2:
Squashed!

PR #394, opened 2017-04-12
Look for instrumentation data in the right place.

snim2:


ltratt:
I think @vext01 should do this one (I'm clueless about this part of things). Switching assignees.

vext01:
So where was it erroneously looking before? Just wondering why it worked when we make the plots manually...

snim2:
I suspect because everything was in the same directory. When the data was in the `annotated_data` directory, it would look for `annotated_dataannotated_data/FILENAME`

vext01:
OK

vext01 (approved):

PR #409, opened 2017-06-28
depot_tools is not fetching fixed versions of V8 deps. Fix them.

vext01:
This makes sure that V8 dependencies are fixed before the build. The versions you see are from bencher6 0.9.

The Python script scrapes the git versions out of an existing v8 source directory and generates a shell script to run as part of build.sh.

Let's not merge this just yet, because I need to test the resultant d8 works, but reviewing can happen. Looks OK?


vext01:
I've run a tiny-sized V8 experiment with no issues. Now a medium-sized one is running. Should be done after lunch.

ltratt:
This looks good to me, although I think a brief doc somewhere would help the future you and I remember what this does.

ltratt (approved):

vext01:
How is this? OK to squash?

bin/find_v8_git_versions.py
```diff
@@ -1,4 +1,13 @@
 #!/usr/bin/env python2.7
+"""
+This script exists because depot_tools (the git wrapper used by V8) does not
+fetch fixed versions of the dependencies it downloads via Git. There are some
+20+ git repos fetched. This script extracts the git version hashes from an
+existing V8 source directory (with dependencies pre-fetched) and generates a
```
ltratt, line 6: I found the bit after this a little confusing. How about something like "generates a shell script which checks out those specific versions. This can then be called in the future to ensure that later builds always use precisely the same versions as when the shell script was generated."?

vext01: sure

vext01:
Tweaked your text slightly, looks good?

ltratt:
OK -- squash away.

vext01:
splat

PR #412, opened 2017-07-13
Slightly wider histograms in tables.

snim2:
What it says on the tin.


ltratt (approved):

ltratt:
Is it just me or have the bars not got wider but the gaps between bars is wider? [This may be a reasonable thing to do, but I'm not sure if it's what we intended or not.]

snim2:
Good question, let me have a look.

snim2:
OK, let me know if this is good to squash. I've checked through and I don't *think* any of the bars are overlapping (which was the reason behind the rather cautious value for the bar width from before).

ltratt:
I like it -- please squash away!

snim2:
Squashed!

PR #414, opened 2017-07-17
Add scripts related to truncated iterations.

snim2:
This PR adds two scripts dealing with truncated iterations. Of the changes we discussed in person, these changes have been made:

 - Sort columns in reverse
 - Add new row for original data
 - ASSERT that says all same + different are always the same
 - New columns to report DELTA between adjacent rows

These changes will happen in later PRs:

 - Aggregate data from all three machine data
 - Re-run outliers script
 - Scale down window size (10%) and steady state (25%) with last_iter

### DIFF
--- a/bin/generate_truncated_json
+++ b/bin/generate_truncated_json
@@ -1,0 +1,143 @@
+#!/usr/bin/python2.7
+
+# Copyright (c) 2017 King's College London
+# created by the Software Development Team <http://soft-dev.org/>
+#
+# The Universal Permissive License (UPL), Version 1.0
+#
+# Subject to the condition set forth below, permission is hereby granted to any
+# person obtaining a copy of this software, associated documentation and/or
+# data (collectively the "Software"), free of charge and under any and all
+# copyright rights in the Software, and any and all patent rights owned or
+# freely licensable by each licensor hereunder covering either (i) the
+# unmodified Software as contributed to or provided by such licensor, or (ii)
+# the Larger Works (as defined below), to deal in both
+#
+# (a) the Software, and
+# (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+# one is included with the Software (each a "Larger Work" to which the Software
+# is contributed by such licensors),
+#
+# without restriction, including without limitation the rights to copy, create
+# derivative works of, display, perform, and distribute the Software and make,
+# use, sell, offer for sale, import, export, have made, and have sold the
+# Software and the Larger Work(s), and to sublicense the foregoing rights on
+# either these or other terms.
+#
+# This license is subject to the following condition: The above copyright
+# notice and either this complete permission notice or at a minimum a reference
+# to the UPL must be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import argparse
+import os.path
+import sys
+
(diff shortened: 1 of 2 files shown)


bin/generate_truncated_json
```diff
@@ -0,0 +1,160 @@
...
+
+import argparse
+import os.path
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+from warmup.krun_results import read_krun_results_file, write_krun_results_file
+
+MAX_ITERS = 1950
+
```
ltratt: Let's put the constants on consecutive lines (save newline bytes! save trees!).

bin/table_truncated_json
```diff
@@ -0,0 +1,148 @@
...
+        if 'truncated_' not in filename:  # Original results file.
+            print 'Loading original results.'
+            _, original_results = parse_krun_file_with_changepoints([filename])
+            truncated_results[2000] = copy.deepcopy(original_results)
+        else:
+            last_iter = filename.split('.')[0].split('_')[1]
+            print 'Loading last iter', last_iter, 'results file.'
+            _, truncated_results[int(last_iter)] = parse_krun_file_with_changepoints([filename])
+    assert original_results is not None, 'No original results file.'
+    machine = original_results.keys()[0]  # ASSUME one machine per results file.
```
ltratt: Can I suggest we do `XXX assumes one machine per results file`?

ltratt: Or change into an assert?

snim2:
New PDFs: [new_summary.pdf](https://github.com/softdevteam/warmup_experiment/files/1153241/new_summary.pdf)

bin/generate_truncated_json
```diff
@@ -0,0 +1,155 @@
...
+MIN_ITERS = 100
+ITER_STEP = -50
+_BLANK_BENCHMARK = { 'wallclock_times': dict(), # Measurement data.
+                    'core_cycle_counts': dict(), 'aperf_counts': dict(),
+                    'mperf_counts': dict(), 'audit': dict(), 'config': '',
+                    'reboots': 0, 'starting_temperatures': list(),
+                    'eta_estimates': list(), 'error_flag': list(),
+                    'all_outliers': dict(), 'unique_outliers': dict(),
+                    'common_outliers': dict(), }
+_SKIP_OUTER_KEYS = ['audit', 'reboots', 'mperf_counts', 'aperf_counts',
```
vext01: Would it be more concise to have a list of keys that you *do* wish to use? Instead of the opposite?

bin/generate_truncated_json
```diff
@@ -0,0 +1,155 @@
...
+                    'eta_estimates', 'starting_temperatures', 'core_cycle_counts',
+                    'config', 'error_flag', 'window_size']
+
+
+def parse_krun_file_with_changepoints(json_files):
+    """Simplified version of the same function from krun_results."""
```
vext01: Simplified in that it does fewer things? or?

I guess I'm wondering if we can avoid more duplication... Maybe not?

bin/generate_truncated_json
```diff
@@ -0,0 +1,155 @@
...
+    window_size = None
+    for filename in json_files:
+        assert os.path.exists(filename), 'File %s does not exist.' % filename
+        data = read_krun_results_file(filename)
+        machine_name = data['audit']['uname'].split(' ')[1]
+        if '.' in machine_name:  # Remove domain, if there is one.
+            machine_name = machine_name.split('.')[0]
+        if machine_name not in data_dictionary:
+            data_dictionary[machine_name] = data
+        else:  # We may have two datasets from the same machine.
```
vext01: Can this happen? I cant think of any situation where a machine's data is split across many files. I may be wrong.

vext01 (commented):
LGTM. A handful of comments.

bin/generate_truncated_json
snim2: I don't think so, but this will have to change in the next PR anyway.

bin/generate_truncated_json
snim2: It doesn't happen in our datasets, but I don't think there is a reason why it can't happen.

bin/generate_truncated_json
snim2: Not sure, note this branch is only taken if we have two datasets from the same machine.

bin/generate_truncated_json
ltratt: If we don't want it to happen, let's add an `assert` or similar.

snim2:
I think this is up to date now, w.r.t. the comments. Are we good to squash?

ltratt:
Looks good to me. If Edd's OK, please squash and then I'll merge.

ltratt (approved):

vext01:
The new column makes it very clear where our cliff is :)

This looks fine to me. Go ahead.

snim2:
Splat!

PR #415, opened 2017-07-18
Add python-dev to prerequisites.

snim2:
What it says on the tin.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -33,7 +33,7 @@ Debian-based systems:
 ```sh
 $ sudo apt-get install build-essential python2.7 pypy bzip2 libssl-dev \
        pkg-config libcurl4-openssl-dev python-numpy python-matplotlib \
-       python-pip texlive-latex-extra wget xorg-dev \
+       python-pip texlive-latex-extra wget xorg-dev python2.7-dev \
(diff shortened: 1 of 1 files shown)


ltratt:
Are we definitely very sure we need this? I assume this is for `Python.h` or something? What actually uses that?

vext01 (approved):

vext01:
It's an R extension.

ltratt:
Ah is this for the "link Python to R" stuff?

snim2:
I had to install this on bencher3 before the build_stats script could work.

ltratt:
Yes, but I just want to briefly understand what it fixed.

snim2:
Ah, as Edd said, the R packages could not be compiled without it.

vext01:
I suppose R embeds Python? Is that the case Sarah?

ltratt:
"R needs Python" just seems a bit weird on the face of it right?

snim2:
We use `r2py` which calls R from Python, but R probably does embed it also.

ltratt:
Right, so `r2py` sounds very plausible. Makes sense.

vext01:
Sorry, that's what I meant. r2py causes R to embed Python.

PR #416, opened 2017-07-18
Rerun outlier script

snim2:
This PR adds the missing functionality from the truncation scripts:

  * The outlier script is run on truncated files before the changepoint script is run
  * The window size and "steady" values that are passed to the two external scripts are scaled with the number of iterations

This only leaves aggregating the values from different machines.

Old PDF: [old_summary.pdf](https://github.com/softdevteam/warmup_experiment/files/1156977/old_summary.pdf)

New PDF: [new_summary.pdf](https://github.com/softdevteam/warmup_experiment/files/1156978/new_summary.pdf)

### DIFF
--- a/bin/generate_truncated_json
+++ b/bin/generate_truncated_json
@@ -47,6 +47,8 @@ from warmup.krun_results import read_krun_results_file, write_krun_results_file
 MAX_ITERS = 1950
 MIN_ITERS = 100
 ITER_STEP = -50
+WINDOW_PROPORTION = 0.1  # Needed for outlier script options.
+STEADY_PROPORTION = 0.25  # Needed for changepoint script options.
 _BLANK_BENCHMARK = { 'wallclock_times': dict(), # Measurement data.
                     'core_cycle_counts': dict(), 'aperf_counts': dict(),
                     'mperf_counts': dict(), 'audit': dict(), 'config': '',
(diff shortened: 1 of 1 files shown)


ltratt:
Looks reasonable to me!

ltratt (approved):

ltratt:
BTW, which machine(s) are included in the example PDF?

ltratt:
There is also a really interesting / surprising / we-should-check it jump at 1050 -> 1000. At 1050 the %different is 19.13 but at 1000 it's 12.83. That's worth looking into to make sure nothing's gone funny.

snim2:
Data is from Bencher 7

ltratt:
OK. Can we check that blip carefully? It's quite possibly correct, but it's so unexpected that we need to be very sure.

ltratt:
[Might be worth comparing it to b5. I imagine it'll be the same, but you never know...]

snim2:
Updated (note the numbers are slightly different, possibly due to bootstrapping): 
[new_summary.pdf](https://github.com/softdevteam/warmup_experiment/files/1159039/new_summary.pdf)

snim2:
Needs squashing

ltratt:
Squash away!

snim2:
Splat.

PR #421, opened 2017-07-20
Plot 'same' truncated data values.

snim2:
Example:

image


snim2:
Updated to match font sizes in other charts in the paper: 
image

ltratt:
Excellent stuff! That graph really brings the data to life. I hadn't appreciated that there's a definite curve, but that the 1024 thing really throws a spanner in the works.

Some small suggestions (nothing major):
 * I think it might be easier to see the kink in the data if we join the dots.
 * Can we add the discontinuous sign to the bottom left of the axes?
 * The x axis should stop at either 1950 or 2000 (both are defensible), but not 2050. I wonder if the axis should actually be 100 to 2000? [Having the intervals at 50 makes sense in one way, but it's hard to read!]
 * x axis label should be "#in-process iterations" (to fit with our tables)?
 * y axis label could be something like "% of process executions with the same classification as n=2000"?

snim2:
Needs squashing

ltratt:
I think we need to address the comments before squashing?

snim2:
Ah, somehow I missed them...

snim2:
Next table: [new_summary.pdf](https://github.com/softdevteam/warmup_experiment/files/1174985/new_summary.pdf)

Next plot: [same_plot_tens.pdf](https://github.com/softdevteam/warmup_experiment/files/1174987/same_plot_tens.pdf)

This is only missing the "truncated axis" markers, so this seems like a good time to re-review and merge this PR, and the truncated axis can be pushed separately.

ltratt:
Agreed. If you squash, I'll merge.

snim2:
Squashed.

ltratt (approved):

PR #423, opened 2017-07-26
First go at plotting truncated pexecs trials.

snim2:
Current result:

[trunc_pexecs.pdf](https://github.com/softdevteam/warmup_experiment/files/1176864/trunc_pexecs.pdf)


bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,221 @@
...
+                    counts = [0, 0, 0, 0]
+                    for _ in xrange(n_pexecs):
+                        next_class = rangen.choice(classifications[key])
+                        counts[CATEGORIES.index(next_class)] += 1
+                    trials[n_pexecs][key].append(counts)
+    return classifications, trials
+
+
+def do_cis_differ((x1, y1), (x2, y2)):
+    """Given to CIs return True if they do NOT overlap."""
```
ltratt: s/to/two/

bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,221 @@
...
+    trials = dict()
+    for n_pexecs in xrange(MIN_PEXECS, MAX_PEXECS + 1):
+        trials[n_pexecs] = dict()
+        for key in classifications:
+            if not all_same_category(classifications[key]):
+                trials[n_pexecs][key] = list()
+                for _ in xrange(NUMBER_TRIALS):
+                    counts = [0, 0, 0, 0]
+                    for _ in xrange(n_pexecs):
+                        next_class = rangen.choice(classifications[key])
```
ltratt, line 116: This is sampling with replacement: is that what we wanted? I have no idea, but I wondered if this method needs sampling without replacement.

bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,221 @@
...
+                        counts[CATEGORIES.index(next_class)] += 1
+                    trials[n_pexecs][key].append(counts)
+    return classifications, trials
+
+
+def do_cis_differ((x1, y1), (x2, y2)):
+    """Given to CIs return True if they do NOT overlap."""
+
+    return y1 < x2 or y2 < x1
```
ltratt: It's probably worth adding `assert y1 > x1 and y2 > x2` before the `return`. Without that guarantee I think the `return` is incorrect.

bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,221 @@
...
+
+
+def count_changed_trials(classifications, trials):
+    """What proportion of trials changed classifications?"""
+
+    num_similar = dict()  # n_pexecs -> int
+    num_different = dict()  # n_pexecs -> int
```
ltratt, line 134: These two sets store identical (albeit inversed) stats, right? So do we only need one of them?

bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,221 @@
...
+            if all_same_category(classifications[key]):
+                # If the original pexecs for this key all had same classification,
+                # no pexecs will have changed in the trials.
+                num_similar[n_pexecs] += NUMBER_TRIALS
+            else:
+                # Of the NUMBER_TRIALS we generated for this n_pexecs, how many
+                # trials resulted in a significantly different set of classifications
+                # to the original? Here be dragons.
+                for trial in xrange(NUMBER_TRIALS):
+                    similar = True  # Assume trial is similar to original.
```
ltratt: Lines 155-165 can be simplified to something like:

```python
for category in CATEGORIES:
    count = trials[n_pexecs][key][trial].count(category)
    this_ci = proportion_confint(count, n_pexecs, alpha=ALPHA, method=METHOD)
    if do_cis_differ(original_cis[key][category], this_ci):
        num_different[n_pexecs] += 1
        break
else:
    num_similar[n_pexecs] += 1
```

bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,221 @@
...
+
+def draw_plot(same, changed, outfile):
+    """Plot 'same' data and write to PDF file."""
+
+    pdf = PdfPages(outfile)
+    fig, axis = pyplot.subplots()
+    # Prepare data.
+    assert same.keys() == changed.keys()
+    x_vals = sorted(same.keys())
+    y_vals = [float(same[x]) / (changed[x] + same[x]) * 100.0 for x in sorted(same.keys())]
```
ltratt: This is the core calculation. What happens if `same` and `changed` (should that be `different`?) have been inverted somewhere. If this was:

```python
y_vals = [float(changed[x]) / (changed[x] + same[x]) * 100.0 for x in sorted(same.keys())]
```

I think we'd get the shape of graph we expect. Of course, that suggests there's a bug elsewhere...

bin/plot_truncated_pexecs
snim2: It's a good question. We need to know what the total of `num_similar[p_exec] + num_different[p_exec]` is later on (for plotting percentages). That number is always the same, but because the original experiment can skip benchmarks / VMs I think we need to compute is (although to be fair, this may not be the simplest way...).

ltratt: Ah, good point. Then I think this mechanism is fine.

bin/plot_truncated_pexecs
snim2: We do, and it does :(

bin/plot_truncated_pexecs
snim2: Welll... if we have original data like this: `warmup=10, slowdown=10, flat=10, nss=0` then every time we sample, the probability of getting a `warmup` is 1/3, which seems right to me. I can try it the other way but I'm not sure why that would help.

ltratt: I'm not saying it will help :) What I'm saying is that we have to know if the method we're using requires sampling with or without replacement, as they have different effects.

snim2: OK, I've done some reading, and it seems that sampling without replacement is OK, but only if the population size is very large compared to sample size. Which in our case I think it isn't, so we want sampling with replacement.

ltratt: OK.

bin/plot_truncated_pexecs
```diff
@@ -0,0 +1,219 @@
...
+        for key in classifications:
+            if all_same_category(classifications[key]):
+                # If the original pexecs for this key all had same classification,
+                # no pexecs will have changed in the trials.
+                num_similar[n_pexecs] += NUMBER_TRIALS
+            else:
+                # Of the NUMBER_TRIALS we generated for this n_pexecs, how many
+                # trials resulted in a significantly different set of classifications
+                # to the original? Here be dragons.
+                for trial in xrange(NUMBER_TRIALS):
```
ltratt: I'm not convinced by lines 155-163 in the sense that I can't work out how they function. Specifically, what does line 157 do?

snim2:
This has been superseded by #424

PR #424, opened 2017-07-28
Plot truncated pexecs (R version) 

snim2:
This is a version of #423 which uses an R library to compute confidence intervals, rather than the Python statsmodels library.

Latest result: [multinomial_R.pdf](https://github.com/softdevteam/warmup_experiment/files/1182775/multinomial_R.pdf)

The R library has been added to `build_stats.sh` (in a separate commit).

### DIFF
--- a/bin/plot_truncated_pexecs
+++ b/bin/plot_truncated_pexecs
@@ -1,0 +1,233 @@
+#!/usr/bin/python2.7
+
+# Copyright (c) 2017 King's College London
+# created by the Software Development Team <http://soft-dev.org/>
+#
+# The Universal Permissive License (UPL), Version 1.0
+#
+# Subject to the condition set forth below, permission is hereby granted to any
+# person obtaining a copy of this software, associated documentation and/or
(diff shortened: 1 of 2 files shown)


ltratt:
The `os.environ` stuff is a bit evil, but if it works it works. Do you want to squash?

snim2:
Yes, it's what we do in the other scripts. Do you want to leave commit e279887 in or out?

ltratt:
You mean as a separate non-squashed commit? I don't mind either way.

snim2:
No, I mean do we really need it? Without it users will need to do some R magic to make this script work, but I'm not sure we are really expecting other users to run this one?

ltratt:
Ah, I see. I think this script is useful: just because our *current* data shows this pattern doesn't mean others' data will. Other people might also want to work out if they can get away with less pexecs in their case. So I think we should keep it in.

snim2:
OK, splat.

ltratt (approved):

PR #426, opened 2017-08-04
Truncated iteration stats

snim2:
**NEEDS SQUASHING**

Latest PDF from bencher7 data: [b8_fixed.pdf](https://github.com/softdevteam/warmup_experiment/files/1199924/b8_fixed.pdf)

The logic for collecting stats for each truncated iterations JSON file are quite complicated. This is partly because the stats available for each <VM, benchmark> pair differ depending on the classification of the <VM, benchmark> (as per the tables). In particular:

  * If all pexecs are flat we DO NOT have data for steady iter (#, s)
  * If all pexecs are NSS we have no summary data
  * If the <VM, benchmark> is inconsistent and contains an NSS we have no summary data

Please check carefully!

### DIFF
--- a/bin/plot_truncated_iterations
+++ b/bin/plot_truncated_iterations
@@ -51,7 +51,6 @@ import sys
 
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
 from warmup.krun_results import parse_krun_file_with_changepoints
-from warmup.latex import end_document, preamble
 from warmup.plotting import add_margin_to_axes, compute_grid_offsets, style_axis, STYLE_DICT
 
 # Set matplotlib styles, similar to Seaborn 'whitegrid'.
@@ -105,60 +104,6 @@ def summarise_truncated(data_dir):
     return summary
 
 
-def start_table(format_, headings):
-    return """
-{
-\\begin{longtable}{%s}
-\\toprule
-%s \\\\
-\\midrule
-""" % (format_, headings)
-
-
-def end_table():
-    return """
-\\bottomrule
-\\end{longtable}
-}
-"""
-
-
-def write_table(summary, outfile, with_preamble=False):
-    """Write out a LaTeX table, with or without preamble."""
-
-    table_format = 'l' + ('r' * 6)
-    table_headings = '&'.join(['\\textbf{\\# Iterations}', '\\textbf{\\# Same}',
-                              '\\textbf{\\# Different}', '\\textbf{\\# Delta}',
-                              '\\textbf{\\% Same}', '\\textbf{\\% Different}',
-                              '\\textbf{\\% Delta}',])
-    total_original = float(summary[DEFAULT_ITERS]['same'] + summary[DEFAULT_ITERS]['different'])
-    with open(outfile, 'w') as fp:
-        if with_preamble:
-            fp.write(preamble('Truncated data results'))
-            fp.write('\\centering\n')
-        fp.write(start_table(table_format, table_headings))
-        for last_iter in sorted(summary, reverse=True):
-            total = float(summary[last_iter]['same'] + summary[last_iter]['different'])
-            assert total == total_original
-            if last_iter == DEFAULT_ITERS:
-                fp.write('%d & %d & %d & %s & %.2f & %.2f & %s \\\\' %
-                         (last_iter, summary[last_iter]['same'], summary[last_iter]['different'],
-                          'n/a', summary[last_iter]['same'] / total * 100.0,
-                          summary[last_iter]['different'] / total * 100.0, 'n/a'))
-            else:
-                pc_delta = (summary[last_iter + ITER_STEP]['same'] / total * 100.0) - \
-                             (summary[last_iter]['same'] / total * 100.0)
-                fp.write('%d & %d & %d & %d & %.2f & %.2f & %.2f\\\\' %
-                         (last_iter, summary[last_iter]['same'], summary[last_iter]['different'],
-                          summary[last_iter + ITER_STEP]['same'] - summary[last_iter]['same'],
-                          summary[last_iter]['same'] / total * 100.0,
-                          summary[last_iter]['different'] / total * 100.0, pc_delta))
-        fp.write(end_table())
-        if with_preamble:
-            fp.write(end_document())
-    print('Saved: %s' % outfile)
-
-
 def draw_plot(summary, outfile):
     """Plot 'same' data and write to PDF file."""
 
@@ -198,23 +143,14 @@ def create_cli_parser():
     parser = argparse.ArgumentParser(description=description)
     parser.add_argument('directory', action='store', default='.', type=str,
                         help='Directory containing truncated results files.')
-    parser.add_argument('--texfile', '-t', action='store', dest='latex_file',
-                        type=str, help='Name of the LaTeX file to write table to.',
-                        required=True)
     parser.add_argument('--pdffile', '-p', action='store', dest='plot_file',
                         type=str, help='Name of the PDF file to write plot to.',
                         required=True)
-    parser.add_argument('--with-preamble', action='store_true',
-                        dest='with_preamble', default=False,
-                        help='Write out a whole LaTeX article (not just the table).')
     return parser
 
 
 if __name__ == '__main__':
     parser = create_cli_parser()
     options = parser.parse_args()
-    if options.with_preamble:
-        print 'Writing out full document, with preamble.'
     summary = summarise_truncated(options.directory)
-    write_table(summary, options.latex_file, options.with_preamble)
     draw_plot(summary, options.plot_file)
--- a/bin/plot_truncated_iterations
+++ b/bin/plot_truncated_iterations
@@ -38,24 +38,53 @@
 # SOFTWARE.
 
 import argparse
-import copy
 import glob
+import json
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as pyplot
 from matplotlib.backends.backend_pdf import PdfPages
+import numpy
 import os
 import os.path
 import sys
 
+# We use a custom install of rpy2, relative to the top-level of the repo.
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                'work', 'pylibs'))
+# R packages are stored relative to the top-level of the repo.
+os.environ['PATH'] = ':'.join([os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                               'work', 'R-inst', 'bin'), os.environ.get('PATH', '')])
+os.environ['LD_LIBRARY_PATH'] = ':'.join([os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                         'work', 'R-inst', 'lib', 'R', 'lib'), os.environ.get('LD_LIBRARY_PATH', '')])
+os.environ['R_LIBS_USER'] = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                         'work', 'R-inst', 'lib', 'R', 'library')
+
+import rpy2
+import rpy2.interactive.packages
+import rpy2.robjects
 
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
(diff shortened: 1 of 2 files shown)


ltratt:
This looks like a good start! A couple of immediate comments.

There's something odd going on at about x=1930: the intersection is above the class line which, I think, should be impossible. Should we have an assertion that guarantees this can't happen?

The steady iter line and legend don't seem consistent? The legend has it as a dashed orange line, but the graph is orange circles with blue/orange dashed lines?

snim2:
Ah, the second one is deliberate. The steady iter (#, s) lines coincide, so you are seeing the blue solid line underneath the dashed line.

ltratt:
Oh they're always the same. Then we should probably just get rid of one of those lines, because they're both the same measure.

snim2:
When we discussed this in the thrash, you took the view that they might be the same in our data and different in data from other machines. (To be clear, I'm not convinced of that, but it's probably worth thinking about)

ltratt:
Not sure if we're talking about the same thing? Steady state (s) and steady iter (#) are the same measure by definition (just in different units).

snim2:
We are. I'll get rid of one of the lines.

snim2:
Test case (iteration 1970 was the problematic case):  [fix.pdf](https://github.com/softdevteam/warmup_experiment/files/1200405/fix.pdf)
 
debug output:

```
Truncated to 1900 iterations
summary[1900][steady_iter] = {u'different': 2, u'same': 25}, % same = 92.59
summary[1900][steady_state_time] = {u'different': 2, u'same': 25}, % same = 92.59
summary[1900][intersection] = {u'different': 10, u'same': 36}, % same = 78.26
summary[1900][classifications] = {u'different': 18, u'same': 1362}, % same = 98.70

Truncated to 1930 iterations
summary[1930][steady_iter] = {u'different': 1, u'same': 25}, % same = 96.15
summary[1930][steady_state_time] = {u'different': 1, u'same': 25}, % same = 96.15
summary[1930][intersection] = {u'different': 10, u'same': 36}, % same = 78.26
summary[1930][classifications] = {u'different': 18, u'same': 1362}, % same = 98.70

Truncated to 1970 iterations
summary[1970][steady_iter] = {u'different': 0, u'same': 26}, % same = 100.00
summary[1970][steady_state_time] = {u'different': 0, u'same': 26}, % same = 100.00
summary[1970][intersection] = {u'different': 6, u'same': 40}, % same = 86.96
summary[1970][classifications] = {u'different': 9, u'same': 1371}, % same = 99.35
```

ltratt:
Looks plausible! Will be interesting to see the full graph.

ltratt:
Although here's a question: I think that, at any point, the intersection line must be greater than or equal to the sum of the other 3 lines (in all cases taking 100-y as the measure, if that makes sense). But take x=1900. It looks to me like Class=3, steady iter=8, steady time=8, but the intersection is 22. That suggests that some pexecs are being classified as changed even though none of their classification, steady iter, or steady time doesn't match?

snim2:
We are looking at each individual classification for each pexec, whereas we look at the steady time and steady iter for each <VM, benchmark> pair. So it may be that (for example) a "good inconsistent" <VM, benchmark> pair has similar steady iter and times, but different classifications in various pexecs.

ltratt:
Erp, I've confused myself. Can you give me a concrete example of this to unstick my brain?

snim2:
These are the tables for 1900 and 2000 iterations:

  * 1900: [trunc1900.pdf](https://github.com/softdevteam/warmup_experiment/files/1200723/trunc1900.pdf)
  * 2000: [trunc2000.pdf](https://github.com/softdevteam/warmup_experiment/files/1200727/trunc2000.pdf)

The summary for 1900 is:

```
summary[1900][steady_iter] = {u'different': 2, u'same': 25}, % same = 92.59
summary[1900][steady_state_time] = {u'different': 2, u'same': 25}, % same = 92.59
summary[1900][intersection] = {u'different': 10, u'same': 36}, % same = 78.26
summary[1900][classifications] = {u'different': 18, u'same': 1362}, % same = 98.70
```

So, I think your question is "why does `intersection[same] + intersection[different]` not equal `steady_iter[same] + steady_iter[different]`? 

The answer is that we always have classification data for all pexecs, but we don't always have steady iter and time data (e.g. when ALL pexecs are NSS). For example:

```
richards:C:default-c
process exec 21 differ in classification: no steady state flat
ALL bad inconsistent
```

So, in this case we would have had 1 classification `different`, 29 classifications `same` but no data for steady iter or time.

I've left the full debug output here: https://gist.github.com/snim2/85540f2ad11deb90288580435da99ca2 (but it is pretty unreadable).

ltratt:
> The answer is that we always have classification data for all pexecs, but we don't always have steady iter and time data (e.g. when ALL pexecs are NSS).

Good point, I'd forgotten about that.

bin/plot_truncated_iterations
```diff
@@ -0,0 +1,336 @@
...
+def do_iqrs_differ((x1, y1), (x2, y2)):
+    """Given two IQRs return True if they do NOT overlap."""
+
+    assert (y1 > x1 or x1 == y1) and (y2 > x2 or x2 == y2)
+    return y1 < x2 or y2 < x1
+
+
+def do_cis_differ(mean1, ci1, mean2, ci2):
+    """Given two CIs return True if they do NOT overlap."""
+
```
ltratt: If ci1 and cit2 are always positive (maybe it isn't?), the code below can be simplified to:

```
x1 = mean1 - ci1
y1 = mean1 + ci1
x2 = mean2 - ci2
y2 = mean2 + ci2
```

bin/plot_truncated_iterations
```diff
@@ -0,0 +1,336 @@
...
+                summary['steady_state_time'][last_iter]['different'] += 1
+                summary['intersection'][last_iter]['different'] += 1
+                continue
+            # All three measures should be available in both the DEFAULT_ITER and last_iter cases.
+            if do_iqrs_differ(truncated_summaries[DEFAULT_ITERS]['machines'][machine][vm][bench]['steady_state_iteration_iqr'],
+                              truncated_summaries[last_iter]['machines'][machine][vm][bench]['steady_state_iteration_iqr']):
+                summary['steady_iter'][last_iter]['different'] += 1
+                intersection = False
+            else:
+                summary['steady_iter'][last_iter]['same'] += 1
```
ltratt: Should there be a `continue` here?

snim2: No, at this stage we have gone through all the special cases and we need to test the CIs for the steady iter and time values.

ltratt: OK.

bin/plot_truncated_iterations
snim2: I ran through the small test case and didn't get any negative CIs, but in general they can be negative, so I'd be inclined to leave this.

ltratt: How can they be negative? If that does happen, it's probably a bug, and we'd probably end up printing CIs like "+/- -x" which would be... weird. So I'd suggest adding an `assert` that the CIs are `>= 0` and then simplify the code.

snim2: Ah, I forgot we do this with bootstrapping, so you are right.

ltratt:
OK, so now I guess we wait for an updated graph then probably it'll be time for squashing.

snim2:
Next PDF: [b8.pdf](https://github.com/softdevteam/warmup_experiment/files/1202317/b8.pdf)

ltratt:
Interesting. I wonder why there's such a significant drop at x=1990?

snim2:
Minor tweak: [b8_zorder.pdf](https://github.com/softdevteam/warmup_experiment/files/1202476/b8_zorder.pdf)

ltratt:
Where are we in terms of tidying this up, reviewing it, and getting it merged?

snim2:
[For the record] this PR is about to be squashed and force-pushed before re-review.

ltratt:
Agreed.

snim2:
Squashed and read for the next review.

bin/plot_truncated_iterations
```diff
@@ -0,0 +1,374 @@
...
+import numpy
+import os
+import os.path
+import sys
+
+# We use a custom install of rpy2, relative to the top-level of the repo.
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+                                'work', 'pylibs'))
+# R packages are stored relative to the top-level of the repo.
+os.environ['PATH'] = ':'.join([os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
```
ltratt, line 56: We know this stuff doesn't work, right?

bin/plot_truncated_iterations
```diff
@@ -0,0 +1,374 @@
...
+    for key in original_results[machine]['classifications']:
+        if len(original_results[machine]['classifications'][key]) == 0:  # Skipped benchmark.
+            continue
+        class_counts = [original_results[machine]['classifications'][key].count(category) for category in CATEGORIES]
+        class_cis[key] = numpy.array(MCI.multinomialCI(rpy2.robjects.FloatVector(class_counts), ALPHA))
+    # Generate summary of results to be plotted.
+    summary = {'classifications': dict(), 'steady_iter': dict(), 'steady_state_time': dict(), 'intersection': dict()}
+    for last_iter in truncated_summaries:
+        if last_iter not in summary:
+            summary['classifications'][last_iter] = { 'same': 0, 'different': 0 }
```
ltratt: One thing I'd like to do in this function -- which is fairly easy, but will also help performance a fair bit -- is to change all the strings to constants e.g. `same` -> `SAME` (or whatever). In fact for the dictionaries which just have two keys, we'd be much better making them lists (e.g. `[<SAME>, <DIFFERENT>]`) which should also save a lot of memory (which we really need to do).

ltratt:
OK so two things only. I think we should fix the `os.environ` thing in this PR: I don't mind if we fix the int constants & lists stuff in this PR or one that shortly follows it (provided it gets done soon).

bin/plot_truncated_iterations
snim2: We do, but I would prefer to merge this and the other PR now and apply a fix for #430 to all the scripts that need it in one go.

ltratt: OK.

snim2:
OK, I think this is ready for re-review / squashing...

ltratt:
Can you give me a rough idea of the performance improvement from using lists instead of dicts?

bin/plot_truncated_iterations
```diff
@@ -0,0 +1,376 @@
...
+    machine = original_results.keys()[0]
+    # Generate CIs for DEFAULT_ITER classification data.
+    class_cis = dict()
+    for key in original_results[machine]['classifications']:
+        if len(original_results[machine]['classifications'][key]) == 0:  # Skipped benchmark.
+            continue
+        class_counts = [original_results[machine]['classifications'][key].count(category) for category in CATEGORIES]
+        class_cis[key] = numpy.array(MCI.multinomialCI(rpy2.robjects.FloatVector(class_counts), ALPHA))
+    # Generate summary of results to be plotted.
+    summary = {'classifications': dict(), 'steady_iter': dict(), 'steady_state_time': dict(), 'intersection': dict()}
```
ltratt: This would also be better as a list.

ltratt:
Ignore my performance comment for the time being -- let's get rid of more of these dictionaries-that-should-be-lists. We'll save a fair bit of memory and almost certainly make things quite a bit faster, both of which we need to improve!

snim2:
With lists: 16mins, with dicts: 18mins (using test data, and not a quiet machine)

ltratt:
OK, so that's a promising start, so let's rid ourselves of more dictionaries :)

snim2:
OK, ready again for review...

snim2:
Testing on a small dataset:

  * before changing dicts to lists (13 minues): [test_small_before_time.pdf](https://github.com/softdevteam/warmup_experiment/files/1314706/test_small_before_time.pdf)
  * after (12 minutes): [test_small_latest_time.pdf](https://github.com/softdevteam/warmup_experiment/files/1314709/test_small_latest_time.pdf)

ltratt:
Ready to squash?

snim2:
Splat.

ltratt (approved):

PR #431, opened 2017-08-26
Improve shuffled pexecs

snim2:
This PR adds steady iter / perf / 'overall' data to the shuffled pexecs chart. The commits need a **lot** of squashing!

This is a test chart, you will notice it is very jagged, unlike the current chart in the paper, this is because this test run was completed with `NUMBER_TRIALS = 1`, to get some quick output. A full run will bump that number up.

[test_pexecs.pdf](https://github.com/softdevteam/warmup_experiment/files/1253551/test_pexecs.pdf)

### DIFF
--- a/bin/plot_truncated_pexecs
+++ b/bin/plot_truncated_pexecs
@@ -38,11 +38,13 @@
 # SOFTWARE.
 
 import argparse
+import json
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as pyplot
 from matplotlib.backends.backend_pdf import PdfPages
 import os.path
+import multiprocessing
 import numpy
 import random
 import sys
@@ -65,6 +67,7 @@ import rpy2.robjects
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
 from warmup.krun_results import parse_krun_file_with_changepoints
 from warmup.plotting import add_margin_to_axes, compute_grid_offsets, style_axis, STYLE_DICT
+from warmup.summary_statistics import collect_summary_statistics
 
 # Set matplotlib styles, similar to Seaborn 'whitegrid'.
 for style in STYLE_DICT:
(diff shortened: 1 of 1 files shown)


snim2:
trials=10 example: [pexecs_10_trials.pdf](https://github.com/softdevteam/warmup_experiment/files/1253759/pexecs_10_trials.pdf)

ltratt:
Where are we in terms of tidying this up, reviewing it, and getting it merged?

snim2:
Now the last two commits are in, this PR can be squashed prior to re-review.

bin/plot_truncated_pexecs
```diff
@@ -190,19 +190,17 @@ def generate_trials(json_file):
...
-
-    print 'Start parallel'
-    # FIXME: NEEDS CHANGING BY HAND
-    start_ends = [(2, 5), (5, 10), (10, 15)]  # [inclusive, exclusive) intervals.
+
+    print 'Start parallel execution.'
+    # start_ends describes which work is done on which core, and must match the
+    # physical machine. Intervals are [inclusive, exclusive).
+    start_ends = [(2, 11), (10, 21), (20, 31)]  # 4-core machine.
+    # start_ends = [(2, 3), (3, 4), (4, 5), (5, 6),  # 32-core machine.
```
ltratt: I'm not sure I'm super comfortable with this in the PR, even though I know why it's there. It's such a hack that I feel we should either maintain it locally or make it some sort of feature (e.g. we could provide settings for 4, 8, 16, and 32 core machines).

ltratt:
So one comment that needs addressing before we can think about squashing.

We should also consider the same string -> int (and, if possible, dict -> list) hack(s) from the other PR.

snim2:
Somewhere between commits 40fcb2d and fc1f70a there is a bug here that triggers one of the assertions....

snim2:
OK, the bugs that were introduced by turning `dict`s into `list`s have now been resolved, so this is ready for re-review.

bin/plot_truncated_pexecs
```diff
@@ -202,13 +202,24 @@ def generate_trials(json_file):
...
     # physical machine. Intervals are [inclusive, exclusive).
-    start_ends = [(2, 11), (10, 21), (20, 31)]  # 4-core machine.
-    # start_ends = [(2, 3), (3, 4), (4, 5), (5, 6),  # 32-core machine.
-    #               (6, 7), (7, 8), (9, 10), (10, 11),
-    #               (11, 12), (12, 13), (13, 14), (14, 15), (15, 16),
-    #               (16, 17), (17, 18), (18, 19), (19, 20), (20, 21),
-    #               (21, 22), (22, 23), (23, 24), (24, 25), (25, 26),
-    #               (26, 27), (27, 28), (28, 29), (29, 30), (30, 31)]
+    # The second number in the last tuple in the list must by 1 + MAX_PEXECS.
+    assert cores in (4, 8, 16, 32), 'Invalid number of cores (%d)' % cores
```
ltratt: I think we need a more robust error message: `assert`s can be optimised away. So something like:

```python
if cores not in [4, 8, 16, 32]:
    raise ...
```

or `sys.stderr.write(...)` or similar.

ltratt:
Apart from that one comment, this looks good to me.

snim2:
OK to squash?

ltratt:
Yes, and you'll need to rebase against master at the same time.

snim2:
Squashed and rebased...

ltratt (approved):

PR #434, opened 2017-08-29
Sort in place rather than copying.

ltratt:
This gives about a 15% speedup -- not a huge amount, but better than nothing.


snim2:
This looks good to me, should I test it before merging?

ltratt:
That's up to you.

snim2:
OK, I ran the table script on this branch and the main table seems to come out with the same numbers, so it looks good!

snim2 (approved):

PR #436, opened 2017-09-01
Each benchmark should have at least 6 rows in the classifier table.

vext01:
Pad the table up if we have to.

A dirty, but effective, hack :gift_heart:

### DIFF
--- a/bin/table_classification_summaries_main
+++ b/bin/table_classification_summaries_main
@@ -54,6 +54,10 @@ from warmup.latex import start_table, STYLE_SYMBOLS
 from warmup.summary_statistics import collect_summary_statistics, convert_to_latex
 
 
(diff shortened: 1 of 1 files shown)


bin/table_classification_summaries_main
```diff
@@ -185,6 +189,10 @@ def write_latex_table(machine, all_benchs, summary, steady_state, tex_file,
                 else:
                     fp.write('\\\\ \n')
                 vm_idx += 1
+            while vm_idx < MIN_VM_ROWS:
+                # pad up table
```
ltratt: I'm not sure this comment is very helpful. I think I'd delete it.

vext01:
Fixed and squashed (since it was trivial).

ltratt (approved):

Issue #381, opened 2017-03-30
Crash in plot script: 'NoneType' object is not iterable


vext01:
```
wilfred:warmup_experiment> ./bin/plot_krun_results annotated_results/warmup_results_0_8_linux2_i7_4790_outliers_w200_changepoints.json.bz2 -b bencher5:richards:C:default-c:0 -o /tmp/a.pdf --cycles 1,2                        
Saving results to: /tmp/a.pdf
Using matplotlib version 1.5.3 with backend agg
Charting with sliding window size: 200, xlimits: None
Loading: annotated_results/warmup_results_0_8_linux2_i7_4790_outliers_w200_changepoints.json.bz2
No VM instrumentation data is available.
Adding run sequence to  richards:C:default-c bencher5
Plotting Linux$_\mathrm{4790}$: Richards (C) on page 01 of 01.
1 plots arranged in 1 rows and 1 columns.
Saved: /tmp/a.pdf
Traceback (most recent call last):
  File "./bin/plot_krun_results", line 1564, in <module>
    inset_xlimits=options.inset_xlimits)
  File "./bin/plot_krun_results", line 316, in main
    core_cycles, cycles_ylimits, inset_xlimits)
  File "./bin/plot_krun_results", line 930, in draw_page
    tukey, core_cycles, cycles_ylimits, inset_x_bounds))
  File "./bin/plot_krun_results", line 415, in __init__
    for core in self.core_cycles:  # For each core specified on command line.
TypeError: 'NoneType' object is not iterable
```

Note that `--cycles` is not a valid argument. This does not happen if you replace `--cycles` with `--core-cycles`. Is the argument being interpreted as `--cycles-ylimits` perhaps?

snim2:
Possibly, I'm not sure how `argparse` deals with things like that.

snim2:
OK, I have had a quick look at this, the problem is that you are taking in a string and splitting the string into ints (similarly with `--cycles-ylimits`). The argparse way of doing things is to allow the parser to split the list and check the types, in which case the script would have to be called like this:

```sh
./bin/plot_krun_results --core-cycles 1 2 -o silly.pdf  warmup_results_0_7_linux1_i7_4790k.json.bz2
```

Note, if you put `--core-cycles` after `-o` then argparse will be confused between the optional number of arguments to `--core-cycles` and the positional argument `warmup_results_0_7...`. 

Do you want me to push a fix for this or leave it for now?

vext01:
[...]

PR #441, opened 2017-09-19
Use argparse to check CLI options.

snim2:
**TODO** This branch should be merged before splitting the repo.

### DIFF
--- a/bin/plot_krun_results
+++ b/bin/plot_krun_results
@@ -1405,11 +1405,12 @@ def create_cli_parser():
                         action='store',
                         dest='xlimits',
                         default=None,
-                        type=str,
-                        help="Specify X-axis limits as a comma separated pair "
-                             "'start,end'. Samples start from 0. e.g. "
-                             "'-x 100,130' will show samples in the range "
-                             "100 to 130.")
+                        nargs=2,
+                        type=int,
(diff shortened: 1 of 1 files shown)


snim2:
The current commit in this PR is very old, and the branch needs bringing up to date. Is it OK to force-push this now?

ltratt:
Feel free to do a rebase.

snim2:
Rebased. This needs some more testing then it will be ready to merge.

snim2:
This PR was intended to "fix" #381, which reported a "crash" in the plotting script. That "crash" was caused by passing in invalid parameters to the script. The fix proposed in this PR changes the CLI options of the plotting script quite radically and uses the `argparse` library to check that valid options will be passed in. It has the complication of not comma-separating lists of arguments, e.g.:

```
... --core-cycles 1, 2, 3 -x 100, 150 ...
```

would become:

```
... --core-cycles 1 2 3  -x 100 150 ...
```

which makes the CLI options trickier to parse, and will require us to re-write parts of the `Makefile` in the other repo, the `warmup_stats` script, and so on. 

I'm minded to think that this PR should be closed without merging, because it needs a much more careful rethink if we are going to use `argparse` in this way in the future.

ltratt:
Agreed -- let's close it and rethink later.

PR #447, opened 2017-10-11
The great cleave

ltratt:
** DO NOT MERGE YET **

This removes the files that have been put into https://github.com/softdevteam/warmup_stats and adjusts things -- except I haven't been able to test this because `build.sh` doesn't yet work on Debian 9.


ltratt:
OK, we need to think what to do with this. Should we merge it even though I haven't tested it due to the lack of Debian 9 support?

vext01:
The other option is to test it on Debian 8.

ltratt:
When/how can we do that?

vext01:
We can either wait for a machine to free up or use a VM.

I got Debian 9 running on VMM the other day (8 shouldn't be hard). If you'd like to try that, I can explain how it's done :)

build.sh
```diff
@@ -114,6 +114,15 @@ if [ ! -d ${SYS_JDK7_HOME} ]; then
     exit 1
 fi
 
+build_warmup_stats() {
+    echo "\n===> Download and build stats\n"
+    if ! [ -d "${HERE}/warmup_stats" ]; then
+        cd ${HERE} && git clone https://github.com/softdevteam/warmup_stats || exit $?
+    fi
+
+    cd ${HERE}/warmup_stats && ./build.sh || exit $?
```
vext01: This will use the system C compiler. You want to set it to zgcc like elsewhere.

I can take care of this. OK for me to add commits?

ltratt: I'll fix this one, but then after that, please go ahead!

ltratt: Actually, do we really care about this? After all, we're not *benchmarking* the warmup_stats repo...

vext01: Hah, you are right.

vext01: I have the fix locally, but I will roll it back.

I have some other small fixes I'd like to push if that's ok with you?

ltratt: OK, then consider that I have committed the null commit, thus fixing the problem ;) Go ahead with your commits.

vext01:
There's what I have so far. Pretty boring stuff. Maybe more.

ltratt:
OK, let me know when you're done.

Makefile
```diff
@@ -91,32 +91,32 @@ plot-warmup-outliers-by-threshold:
...
-		bin/plot_krun_results --wallclock-only -w ${WINDOW_SIZE} -m -t --with-outliers --with-changepoints -o octane.spidermonkey_${PLOTS_WITH_CPTS} octane.spidermonkey_outliers_w${WINDOW_SIZE}_changepoints.json.bz2; \
+		warmup_stats/bin/csv_to_krun_json -u "`uname -a`" -v SpiderMonkey -l JavaScript octane.spidermonkey.results && \
+		warmup_stats/bin/mark_outliers_in_json -w ${WINDOW_SIZE} -t ${OUTLIER_THRESHOLD} octane.spidermonkey.json.bz2 && \
+		LD_LIBRARY_PATH=${R_LD_LIBRARY_PATH}:${LD_LIBRARY_PATH} warmup_stats/bin/mark_changepoints_in_json -s ${STEADY_STATE_EXPECTED} octane.spidermonkey_outliers_w${WINDOW_SIZE}.json.bz2 && \
+		warmup_stats/bin/plot_krun_results --wallclock-only -w ${WINDOW_SIZE} -m -t --with-outliers -o octane.spidermonkey_${PLOTS_NO_CPTS} octane.spidermonkey_outliers_w${WINDOW_SIZE}.json.bz2 && \
+		warmup_stats/bin/plot_krun_results --wallclock-only -w ${WINDOW_SIZE} -m -t --with-outliers --with-changepoints -o octane.spidermonkey_${PLOTS_WITH_CPTS} octane.spidermonkey_outliers_w${WINDOW_SIZE}_changepoints.json.bz2; \
```
vext01, line 119: In the next target:

```
bin/table_classification_summaries_main -o bencher3.table warmup_results_0_7_linux1_i7_4790k_outliers_w200_changepoints.json.bz2
  File "bin/table_classification_summaries_main", line 51, in <module>
    from warmup.krun_results import parse_krun_file_with_changepoints
ImportError: No module named warmup.krun_results
```

Even if the path were correct, I'm not sure about those results filenames. Running the benchmarks would give `warmup_results.json.bz2`, obviously only for the machine you are on. But this target seems to expect the 0.7 data from all the machines named as they are in archive.org.

Thoughts?

FWIW, the plotting target works and just plots using the data from `warmup_results.json.bz2`.

ltratt: I guess the machine-specific names *did* make sense when this was a sort-of internal repo, but I agree, they don't make sense now.

Should those 3 lines be collapsed to just a single line like:

```make
	bin/table_classification_summaries_main -o bencher3.table warmup_results.json.bz2
```

?

I'll fix `bin/table_classification_summaries_main` now.

ltratt:
In theory (but I haven't tried in practise!), the scripts in `bin/` should now be able to find the `warmup` package. Let me know.

ltratt:
@vext01 Have you had any further thoughts on this PR?

Makefile
vext01: Yes, I think we should collapse the lines, but the output file should be machine independent (i.e. not `bencher3.table`).

vext01: How about just `warmup.table`?

ltratt: Agreed. I've pushed a change to that effect.

vext01:
The crash i pasted a while back seems to be fixed btw.

I'll test a few more use-cases tomorrow to see if we can weed anything else out.

ltratt:
Thanks!

vext01:
This isn't your fault, but the dacapo and octane table targets assume the existence of both results files, but on OpenBSD VMs get skipped.

Should we add guards on the existence of the files?

ltratt:
I'm not totally sure what I'd need to guard. Any pointers?

vext01:
Similar to the way the plotting targets work:

https://github.com/softdevteam/warmup_experiment/blob/33d375785d27d29434f40cbdda157bcd35d5cda2/Makefile#L107-L120

vext01:
Note the: `if [ -e octane.spidermonkey.results ]; then \ `

vext01:
If you prefer, I can apply these fixes since i'm geared up to test.

ltratt:
What I'm unsure about is whether we need any more checks. The Dacapo and Octane targets already only do the Graal & SpiderMonkey bits if the files are present and those are the only two that I can see aren't present on OpenBSD. Maybe I'm missing out on something.

vext01:
Look at the table target, it will clearly fail on OpenBSD:

https://github.com/softdevteam/warmup_experiment/blob/33d375785d27d29434f40cbdda157bcd35d5cda2/Makefile#L125-L126

vext01:
similarly for octane:

https://github.com/softdevteam/warmup_experiment/blob/33d375785d27d29434f40cbdda157bcd35d5cda2/Makefile#L128-L129

ltratt:
But those are targets that the user has to explicitly invoke, right? i.e. they're not referenced elsewhere in the Makefile.

vext01:
correct, but why is that important?

I'm kinda wondering how useful all of these plotting/tabling targets really are... I don't use them since they make assumptions about filenames etc. Does Sarah use them I wonder?

ltratt:
The plot- targets are pretty involved, so it seems somewhat useful to have them there. The tables- targets aren't very involved, so I could live with them disappearing.

[As to your first question: since they're not on the default path, no-one can encounter the problem unless they go looking for it. Making tables-dacapo a no-op on OpenBSD might be even more confusing than having it error!]

vext01:
@snim2 what do you think?

vext01:
I also notice that the table targets neglect to run the outlier and annotation steps.

vext01:
Just discovered #449. Highly unlikely to be caused by your changes.

vext01:
I don't know the answer to this, but I notice that the table scripts got split between the repos:

* In `warmup_stats`, we have: `table_classification_summaries_others`
* In `warmup_experiment` we have: `table_classification_summaries_main`, `table_startup_results` and `table_georges_comparison`.

Is there some logic behind this?

snim2:
To answer the last question (before I go back through the thread) there was some logic; the `_others` script is needed by non-Krun users and needs to be kept with the `warmup_stats` script. The `_main` script is really only used by us, which is also true of the `startup` and `georges` scripts (which were just used for the paper). I'm not suggesting that this was the best possible thing, maybe it wasn't, but the purpose was to avoid confusing users outside the group.

vext01:
OK, I'll buy that :)

snim2:
So, I think the table targets may assume that the plotting scripts have already run, so the outliers and changepoints should have been marked, but that may not be written into the Makefile as a dependency. I added the table targets partly to make sure that we documented exactly what we did somewhere, but I think in practice I used a custom script that pulled down the daily machine results and ran all the scripts in a cron job. I don't mind if the table targets go, but if they stay we should check that they are up to date with the CLI changes that happened recently (may only have affected `_others`?).

vext01:
I see. Ideally we would have separate targets for each step of the conversion/annotation process and then we could take advantage of `make`'s dependency tracking too.

I don't think this is super-necessary for Laurie's changes here.

snim2:
OK.

ltratt:
OK, so where does this leave us? :)

vext01:
I've tested many of the make targets that do cross-repo invocations. Other than the bits above (which I have raised issues for where necessary), I think we are good.

There might be other subtle bugs, but I don't think we can expect to find them all at this stage.

What do you think? Shall we merge?

ltratt:
If you're OK with it, I think I should squash first.

vext01:
Yes, ofcourse :) go ahead.

ltratt:
OK, squashed. @vext01 since you're an author of a commit in here, @snim2 should be the one to do the final review and merge.

bin/generate_truncated_json
```diff
@@ -114,7 +114,7 @@ def create_outlier_filename(in_file_name, window):
 def generate_truncated_json(json_file):
     """Generate an output file for each truncated data set."""
 
-    bin_dir = os.path.dirname(os.path.abspath(__file__))
+    warmup_stats_bin_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../warmup_stats/bin")
```
snim2: In the other files we have avoided the use of `..` in filenames and instead used another call to `os.path.dirname`, we should probably be consistent here.

snim2: Similarly `/` should probably be removed in favour of a call (or another argument) to `os.path.join`

snim2:
Ok, that's the end of my review comments

ltratt:
OK, that latest commit hopefully addresses the comment -- but @vext01 will need to test it I think.

vext01:
@snim2 I've never used this part of the repo. What command should I use?

snim2:
Oh boy, I didn't realise it was this file. Firstly change `MAX_ITERS` to `20`, so you only write out one file, otherwise you'll be here all day. Then run the script with `-h` just to check, but you need to pass it a full results file as an argument. Shout if you have any problems, I may well have forgotten something!

vext01:
Oops!

```
vext01@debian8:~/research/warmup_experiment$ ./bin/generate_truncated_json warmup_results_outliers_w200_changepoints.json.bz2
  File "./bin/generate_truncated_json", line 151, in <module>
    generate_truncated_json(options.results_file)
  File "./bin/generate_truncated_json", line 118, in generate_truncated_json
    outlier_script = os.path.join(bin_dir, 'mark_outliers_in_json')
NameError: global name 'bin_dir' is not defined
```

ltratt:
OK, see if that new commit helps.

vext01:
It doesn't crash :)

Watch this space, it's taking a little while.

vext01:
There we go:

```
vext01@debian8:~/research/warmup_experiment$ ./bin/generate_truncated_json warmup_results_outliers_w200_changepoints.json.bz2
Truncated to iteration: 20
Marking outliers with sliding window size: 2
Loading: truncated_20.json.bz2
Writing out: truncated_20_outliers_w2.json.bz2
Marking changepoints and classifications.
Expecting a steady state to be reached before the last 5 iterations.
Using a fixed bound of 0.001s.
Using R version 3.3.2 and changepoint library 2.2.3
Loading: truncated_20_outliers_w2.json.bz2
Writing out: truncated_20_outliers_w2_changepoints.json.bz2
```

ltratt:
Thanks!

I think we're ready for another squash @snim2?

snim2:
Yes please!

ltratt:
OK, squashed!

snim2 (approved):

PR #448, opened 2017-10-12
Talk about Debian compatibility in the README.

vext01:
Small doc update. OK?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ consult the Krun documentation (fetched into `krun/` by `build.sh`), as there
 is a great deal of manual intervention needed to compile a tickless kernel,
 disable Intel P-states, set up `rc.local` etc.
 
+Note that the experiment is designed to run on amd64 machines running Debian 8
+or OpenBSD. Newer versions of Debian do not currently work due to a C++ ABI
+bump which would require a newer C++ compiler (a newer GCC or perhaps clang).
+
 ## Print-traced Benchmarks
 
 The paper mentions that to ensure benchmarks are "AST deterministic",  we


ltratt (approved):

README.md
```diff
@@ -16,6 +16,10 @@ consult the Krun documentation (fetched into `krun/` by `build.sh`), as there
 is a great deal of manual intervention needed to compile a tickless kernel,
 disable Intel P-states, set up `rc.local` etc.
 
+Note that the experiment is designed to run on amd64 machines running Debian 8
+or OpenBSD. Newer versions of Debian do not currently work due to a C++ ABI
+bump which would require us to use a newer C++ compiler.
```
ltratt: Can we say something like "which would require porting the experiment to use a newer C++ compiler" (I'm hoping if we don't say "us" that some kind soul might take it upon themselves to doing the porting!).

vext01: Yeah, but I expect that if someone were to update it, they would not care for OpenBSD.

Updating GCC is easy if you don't care about OpenBSD.

ltratt: OK, so maybe we should say "a newer C++ compiler (i.e. a newer version of GCC or clang)"?

vext01:
Try this.

ltratt:
You don't think it's worth being clear that "newer" could mean GCC or clang?

vext01:
I don't want to be too specific because I don't know the constraints on clang wrt libstdc++.

ltratt:
If we say "e.g." would that be OK? I just think that if I read that, I'd think "only GCC will do", whereas clang is, at the very least, a plausible substitute. [I'm partly thinking that since clang is now better supported on OpenBSD, it might be the lower effort route.]

vext01:
I've revised the wording to include clang as a possible next step.

My hesitancy comes from the fact clang has its own `libc++`, whereas g++ uses `libstdc++` and I'm not sure to what degree you can mix them. It's likely the system C++ libs link libstdc++.

OK to squash?

ltratt:
Works for me! Please squash.

vext01:
splat

PR #452, opened 2017-11-13
Give up on running lusearch. Keeps hanging (on both graal and jdk).

vext01:
I must have restarted dacapo on Linux 3 or 4 times now. It always gets stuck on lusearch. Initially I though this was local to graal, but no. Let's disable it so we can progress.

OK?

Note: merge base is the hotfix branch.


vext01:
Don't forget this Laurie ;)

Issue #451, opened 2017-11-13
More stuff to switch off in hooks


vext01:
As found by @fsfod:

```
+    "sudo systemctl stop apt-daily.timer",
+    "sudo systemctl --runtime disable apt-daily.timer", 
+    "sudo systemctl stop apt-daily-upgrade.timer",
+    "sudo systemctl --runtime disable apt-daily-upgrade.timer",
+    
+    "sudo systemctl stop systemd-timesyncd",
+    "sudo systemctl --runtime disable systemd-timesyncd",
```

vext01:
None of these services exist in Debian 8, so I will hold off implementing these until such a time (if it exists) as we port to Debian 9.

PR #454, opened 2017-11-20
Don't clean /tmp during benchmarks please.

vext01:
Prompted by #451 I've re-audited system services, and found one more to kill:

```
systemd-tmpfiles-clean.timer      loaded active waiting   Daily Cleanup of Temporary Directories
```

We don't want /tmp to be cleaned mid-benchmark.

Still testing, so hold off merge for now.

Looks good?

### DIFF
--- a/warmup.krun
+++ b/warmup.krun
@@ -172,6 +172,9 @@ if sys.platform.startswith("linux"):
         "sudo systemctl stop cron",
         "sudo systemctl stop atd",
         "sudo systemctl stop postfix",
+        "sudo systemctl stop systemd-tmpfiles-clean.timer",
+        "sudo systemctl stop dbus",
+        "sudo systemctl stop dbus.socket",
         "sudo systemctl stop ssh",
     ]
 
@@ -183,6 +186,12 @@ if sys.platform.startswith("linux"):
         "sudo systemctl start cron || true",
         "sudo systemctl start atd || true",
         "sudo systemctl start postfix || true",
+        "sudo systemctl start systemd-tmpfiles-clean.timer || true",
+        "sudo systemctl start dbus",
+        "sudo systemctl start dbus.socket",
+        # these will have been taken down by dbus going down
+        "sudo systemctl start console-kit-daemon",
+        "sudo systemctl start polkitd",
     ]
 elif sys.platform.startswith("openbsd"):
     PRE_EXECUTION_CMDS = [


ltratt:
Is there a way to list all services that are enabled? Maybe we should disable pretty much everything?

fsfod:
systemctl list-timers --no-legend --all

vext01:
It's not that simple. There are things which would likely cause the system to stop working. For example I know dbus and polkit are deeply embedded in a modern Linux system.

We need to be conservative.

The following services are running or waiting on bencher3. I think we already stop the main offenders (cron, ssh, atd, ...):
```
acpid.path                                                                                       loaded active running   ACPI Events Check
session-278.scope                                                                                loaded active running   Session 278 of user vext01
acpid.service                                                                                    loaded active running   ACPI event daemon
atd.service                                                                                      loaded active running   Deferred execution scheduler
console-kit-daemon.service                                                                       loaded active running   Console Manager
cron.service                                                                                     loaded active running   Regular background program processing daemon
dbus.service                                                                                     loaded active running   D-Bus System Message Bus 
getty@tty1.service                                                                               loaded active running   Getty on tty1
polkitd.service                                                                                  loaded active running   Authenticate and Authorize Users to Run Privileged Tasks
postfix.service                                                                                  loaded active running   LSB: Postfix Mail Transport Agent
rpcbind.service                                                                                  loaded active running   LSB: RPC portmapper replacement
rsyslog.service                                                                                  loaded active running   System Logging Service
serial-getty@ttyS0.service                                                                       loaded active running   Serial Getty on ttyS0
ssh.service                                                                                      loaded active running   OpenBSD Secure Shell server
systemd-journald.service                                                                         loaded active running   Journal Service
systemd-logind.service                                                                           loaded active running   Login Service
systemd-udevd.service                                                                            loaded active running   udev Kernel Device Manager
user@1000.service                                                                                loaded active running   User Manager for UID 1000
acpid.socket                                                                                     loaded active running   ACPID Listen Socket
dbus.socket                                                                                      loaded active running   D-Bus System Message Bus Socket
syslog.socket                                                                                    loaded active running   Syslog Socket
systemd-journald-dev-log.socket                                                                  loaded active running   Journal Socket (/dev/log)
systemd-journald.socket                                                                          loaded active running   Journal Socket
systemd-udevd-control.socket                                                                     loaded active running   udev Control Socket
systemd-udevd-kernel.socket                                                                      loaded active running   udev Kernel Socket
proc-sys-fs-binfmt_misc.automount                                                                loaded active waiting   Arbitrary Executable File Formats File System Automount Point
systemd-ask-password-console.path                                                                loaded active waiting   Dispatch Password Requests to Console Directory Watch
systemd-ask-password-wall.path                                                                   loaded active waiting   Forward Password Requests to Wall Directory Watch
systemd-tmpfiles-clean.timer                                                                     loaded active waiting   Daily Cleanup of Temporary Directories
```

vext01:
The output of @fsfod's command on bencher3 is:
```
Mon 2017-11-20 16:44:01 GMT  5h 36min left n/a  n/a    systemd-tmpfiles-clean.timer systemd-tmpfiles-clean.service
n/a                          n/a           n/a  n/a    systemd-readahead-done.timer systemd-readahead-done.service
```

It seems they only really started using timers post Debian 9.

ltratt:
Which of those do we stop? e.g. `postfix.service` isn't one I remember us stopping, but I would suggest that we turn it off.

vext01:
https://github.com/softdevteam/warmup_experiment/blob/f9f284e2971cbbf02c45025a21e1a46bad6452cd/warmup.krun#L166-L176

vext01:
Postfix is certainly off.

ltratt:
`dbus` also sounds like another candidate for stopping.

vext01:
Did you read above about dbus?

ltratt:
Yes, but it wasn't very convincing ;) Certainly for a desktop you need dbus. For a server? I'm not very sure. We could try turning it off, running benchmarks with 1 pexec and 1 in-process iteration; if anything fails, OK, we need dbus. If not... off with its head!

vext01:
The whole service management framework stops working if you disable dbus:

```
vext01@bencher7:~/research/warmup_experiment$ systemctl > /tmp/pre
vext01@bencher7:~/research/warmup_experiment$ sudo systemctl stop dbus
Warning: Stopping dbus.service, but it can still be activated by:
  dbus.socket
vext01@bencher7:~/research/warmup_experiment$ sudo systemctl stop dbus.socket
vext01@bencher7:~/research/warmup_experiment$ systemctl > /tmp/post
Failed to get D-Bus connection: Connection refused
vext01@bencher7:~/research/warmup_experiment$ cat /tmp/post
vext01@bencher7:~/research/warmup_experiment$ systemctl
Failed to get D-Bus connection: Connection refused
```

It's deeply baked in. I'd rather leave it on.

ltratt:
Ye gods. OK, I stand corrected. And horrified.

OK, so shall we merge this PR?

vext01:
yay systemd.

Let me test first.

vext01:
Hold on, i think the error was because I didn't sudo. Let me investigate (rebooting to get back to where we were).

vext01:
OK, so if you turn off dbus and its socket activation part:
```
vext01@bencher7:~$ systemctl --all | grep -E 'running|waiting' > /tmp/pre
vext01@bencher7:~$ sudo systemctl --all | grep -E 'running|waiting' > /tmp/pre
vext01@bencher7:~$ sudo systemctl stop dbus dbus.socket
vext01@bencher7:~$ sudo systemctl --all | grep -E 'running|waiting' > /tmp/post
```

```
--- /tmp/pre	2017-11-20 11:28:39.271916215 +0000
+++ /tmp/post	2017-11-20 11:28:58.559916598 +0000
@@ -5,11 +5,8 @@
   session-1.scope                                                                                  loaded    active   running   Session 1 of user vext01
   acpid.service                                                                                    loaded    active   running   ACPI event daemon
   atd.service                                                                                      loaded    active   running   Deferred execution scheduler
-  console-kit-daemon.service                                                                       loaded    active   running   Console Manager
   cron.service                                                                                     loaded    active   running   Regular background program processing daemon
-  dbus.service                                                                                     loaded    active   running   D-Bus System Message Bus
   getty@tty1.service                                                                               loaded    active   running   Getty on tty1
-  polkitd.service                                                                                  loaded    active   running   Authenticate and Authorize Users to Run Privileged Tasks
   postfix.service                                                                                  loaded    active   running   LSB: Postfix Mail Transport Agent
   rpcbind.service                                                                                  loaded    active   running   LSB: RPC portmapper replacement
   rsyslog.service                                                                                  loaded    active   running   System Logging Service
@@ -20,7 +17,6 @@
   systemd-udevd.service                                                                            loaded    active   running   udev Kernel Device Manager
   user@1000.service                                                                                loaded    active   running   User Manager for UID 1000
   acpid.socket                                                                                     loaded    active   running   ACPID Listen Socket
-  dbus.socket                                                                                      loaded    active   running   D-Bus System Message Bus Socket
   syslog.socket                                                                                    loaded    active   running   Syslog Socket
   systemd-journald-dev-log.socket                                                                  loaded    active   running   Journal Socket (/dev/log)
   systemd-journald.socket                                                                          loaded    active   running   Journal Socket
```

so it kills polkit and consolekit. Killing polkit (which deals with granting normal users certain system priveleges) seems like a bad idea. consolekit is to do with logging in.

But it's true that `systemctl` as a normal user works before, but not after killing dbus, which make me nervous too.

If it's all the same with you, I think we should leave dbus up.

ltratt:
Killing consolekit doesn't seem like the world's biggest problem. I have no idea about polkit. Why don't we just run a 1 pexec 1 iteration experiment and see what happens? After all, if we *can* turn it off, we've just reduced the noise that bit more...

vext01:
OK, but I think we are playing with fire here.

I've noted that bringing dbus back up again doesn't bring up the dependants, so I will need to explicitly kill and start those too.

ltratt:
What's the worst that can happen? It doesn't work. Then we reenable it and carry on our merry way.

vext01:
I just wonder if (for example) any VMs expect to talk to dbus. We shall find out :)

vext01:
proposed diff above. testing...

vext01:
So it looks like it might work with dbus off, but it does seem to somewhat break sudo.

After stopping dbus, invoking sudo will hang for 5-10 seconds before putting an obscure message in the authlog about a socket timeout.

(tests still running)

vext01:
OK a small run completed.

What do you want to do?

ltratt:
Does that mean that we've got lots and lots of instances of `sudo` hanging for 5-10s when running things?

vext01:
I think a new login shell (i.e. as the krun privsep user) will not cause sudo to hang. It's only when I use sudo from an existing shell (I think).

ltratt:
Why don't we try a run with dbus turned off... If it turns out to be a bad idea, we'll probably know within 2-3 days at which point we can change our mind. You never know, it might be responsible for some of the weird things we see. [All I know about dbus is that it occasionally likes to randomly print out huge quantities of debugging stuff to log files once in a while.]

vext01:
Can do. In theory it's just another form of IPC, but it's just amazing how fundamental it has become to Linux.

So ready for me to squash?

ltratt:
Yep, please squash.

vext01:
here we go!

ltratt (approved):

PR #455, opened 2017-11-20
Fix versions

vext01:
We should probably freeze the versions of components marked with `XXX` in build.sh.

I've used the latest versions. The krun tag is the current head.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -125,7 +125,7 @@ build_warmup_stats() {
     fi
 }
 
-# XXX when we stabilise, fix the krun revision.
+KRUN_VERSION=warmup_experiment_results_v1.3
 build_initial_krun() {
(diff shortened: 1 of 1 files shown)


ltratt:
Do we use libkalibera any more I wonder?

vext01:
I wondered the same myself, but I don't want to get into that just now.

I have a vague feeling @snim2 copied the bits we need into `warmup_stats`:
```
warmup_stats/warmup/bootstrapper.py: Much of the code here comes from libkalibera.
```

Here's an issue for this anyway: #456

ltratt:
OK. Merge as-is?

vext01:
Let me squash. I fudged the initial commit.

vext01:
splat.

ltratt (approved):

PR #457, opened 2017-11-20
Skip warmup_stats on OpenBSD-<6.2.

vext01:
As discussed on MM.

Shell script can't compare floats! So I've had to use a python one-liner!

Tested on b6.

OK?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -117,6 +117,14 @@ fi
 
 build_warmup_stats() {
     echo "\n===> Download and build stats\n"
+    # Older OpenBSDs don't have a new enough libzip
+    if [ "`uname`" = "OpenBSD" ]; then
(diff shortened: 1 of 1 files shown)


snim2:
Maybe this is a stupid question, but don't shell scripts usually compare version numbers as strings rather than floats?

vext01:
We need a less-than, though... Or do you mean compare them according to ASCII ordering?

vext01:
So you can compare lexicographically, but it's not available in all shells.

snim2:
Well, a less than can be converted to a geq by swapping the arguments. Anyhow, let's stick to your version, StackOverflow is showing me some heinous BASH code.

Are you happy to merge as-is?

vext01:
yup.

snim2 (approved):

PR #458, opened 2017-11-21
Don't disable dbus, as it's more baked into Linux than we expected.

vext01:
K?


ltratt:
Thanks for this and sorry for sending you on a wild goose chase!

ltratt (approved):

PR #459, opened 2017-11-21
Use a hand-rolled V8 tarball to *really* ensure The versions are fixed.

vext01:
depot_tools continues to cause us problems. 

To fix this, I've used a sledgehammer approach to ensure that there is absolutely no way that we can get different code between builds.

This change:
 * Adds a script to make a V8 tarball
 * Changes `build.sh` to use the tarball.

This means a regular build never uses depot_tools, only the script to make tarballs.

Still testing (compiling as we speak, then we should do a small run of V8 only).

If tests succeed, I will upload the tarball to archive.org and have the script fetch it.

Approach seems sound?

I also raised an issue upstream, since the inability to build a frozen version of V8 seems like a shortcoming:
https://bugs.chromium.org/p/v8/issues/detail?id=7107

### DIFF
--- a/bin/make_v8_source_tarball
+++ b/bin/make_v8_source_tarball
@@ -1,0 +1,69 @@
+#! /bin/sh
+#
+# Roll a "frozen" V8 source tarball.
+
+set -e
+
+DIRNAME=`dirname $0`
+TOPDIR=`readlink -f ${DIRNAME}/..`
+WRKDIR=${TOPDIR}/work
+
+mkdir -p ${WRKDIR}
+echo "===> Working in $WRKDIR"
+
+# Look here to know which V8 branch is currently stable:
+# https://omahaproxy.appspot.com/
+#
+# Note that there is often a newer tag in git than is shown on that page.
+# Make sure you check the tags for the stable branch. Don't use the github
+# mirror to look for tags, as it doesn't have them all.
+#
+# XXX If you are updating the V8 version, you probably want to change/remove
+# the part of this script that rewinds the git versions of the dependencies.
+# Perhaps checkout the deps with the same date as the V8 release instead (then
+# you can also kill fix_v8_versions and find_v8_git_versions too)
+V8_V=5.8.283.32
+V8_TARBALL=v8_fullsource_${V8_V}_`date +%F`.tar.gz
+
+cd ${WRKDIR}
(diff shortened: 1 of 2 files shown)


bin/make_v8_source_tarball
```diff
@@ -0,0 +1,69 @@
...
+cd ${WRKDIR}
+
+if [ -f ${WRKDIR}/${V8_TARBALL} ]; then
+    echo "tarball already exists"
+fi
+
+# The build requires that you clone using this git wrapper: depot_tools
+#
+# We've tried using a fixed version of depot tools in the past. It has a
+# tendency to update itself and swithcing of auto-update causes other
```
vext01: typo `swithcing`

vext01:
To be clear, I have not updated to a new V8. There's a few details:

 * depot_tools updates itself behind your back
 * the switch to stop this behaviour breaks something else
 * the newest depot_tools works for our V8 and deps versions, but on Linux only

So we make the tarball on Linux and use it on BSD. The tarball also shields us from further depot_tools breakage.

vext01:
Works on OpenBSD :)

ltratt:
This all sounds sensible to me. Before we upload a (probably quite big) tarball to archive, let's maybe give the v8 folk a day or two to a) see if there is a freezing mechanism we're unaware of b) if not, if they might be prepared to implement one in short order.

I suggest you fix the typo you found and squash in one go, then I'll merge this PR.

vext01:
I've already uploaded the tarball :P

Just waiting for it to appear and test the download code (coming soon).

We can still wait for the V8 folk of course.

vext01:
There's a few bits. The tarball fetch is as yet untested.

ltratt:
The new commits all look fine.

vext01:
Let's push on with this. I don't think Google are likley to get back to us.

I've tested the fetch, shall I squash?

ltratt:
Yes, please squash.

vext01:
Splat.

I think we've learnt something here. If we see a non-trivial "fetch stuff from all over the place" build system or VCS wrapper, we should probably try to make a static tarball from the start to be sure that no code can be a moving target. Let's bear this in mind for the future.

ltratt (approved):

PR #463, opened 2017-12-13
Simplify creation of V8 tarball.

### LINKED ISSUE #462: Simplify V8 tarball creation.
A post on v8-users reveals how you are supposed to fetch an old version of the V8 src:
https://groups.google.com/forum/#!topic/v8-users/Ng3yx6zshq8

I suggest we make a new tarball via this method (the way we did it is convoluted). Since we won't be running any experiments until the servers move, over the next few days is probably a good time.

Thoughts @ltratt ?

vext01:
Fixes #462.

New tarball uploaded to archive.org, and all tested on bencher6 and 7.

Looks good?


ltratt (approved):

vext01:
Thanks. Feels good to delete so much code :)

